PR #1, opened 2015-02-09
Move to Ginkgo and add Godep for dependencies management

kouno:
As I believe we feel more comfortable with RSpec style tests and stays more inline with the Ruby library.

I still have to export a few tests already written by @lmars before (Time and Stats), but I thought it was not necessary yet.

A bit of work went into moving the testApp code and update it to allow the app to be deleted.

cc/ @mattheworiordan


rest/channel_test.go
```diff
@@ -0,0 +1,44 @@
...
+			err := channel.Publish(event, message)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("reads messages from history", func() {
+			messages, err := channel.History()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(messages[0].Name).To(Equal(event))
+			Expect(messages[0].Data).To(Equal(message))
+		})
```
lmars: This test is identical to the publish one

rest/client.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/hmac"
 	"crypto/sha256"
+	_ "crypto/sha512"
```
lmars: Any reason for this import?

test/support/test_app.go
```diff
@@ -0,0 +1,163 @@
...
+func (t *testApp) populate(res *http.Response) error {
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(body, &t.Config)
+	if err != nil {
+		return err
+	}
```
lmars: The above 9 lines could be simplified to:

```go
if err := json.NewDecoder(res.Body).Decode(&t.Config); err != nil {
	return err
}

File: rest/channel_test.go (hunk 1 above)
Comment (kouno) [inline rest/channel_test.go]: Woops. Sounds like I can't really check Publish without History anyway. I'll merge these tests.

File: test/support/test_app.go (hunk 3 above)
Comment (kouno) [inline test/support/test_app.go]: Sounds good :+1:

File: rest/rest_suite_test.go [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+}
+
+var (
+	client  *rest.Client
+	channel *rest.Channel
+)
+
+var _ = BeforeSuite(func() {
+	_, err := TestAppInstance.Create()
+	Expect(err).NotTo(HaveOccurred())
```
lmars: i think it would be more idiomatic to assign a variable rather than using a global from the other package, for example:

```go
import (
	"github.com/ably/ably-go/test/support"
)

var testApp *support.TestApp

var _ = BeforeSuite(func() {
	var err error
	testApp, err = support.CreateTestApp()
	Expect(err).NotTo(HaveOccurred())
})
```

This also avoids cluttering the namespace with globals from the support package.

rest/channel_test.go
```diff
@@ -0,0 +1,44 @@
...
+var _ = Describe("Channel", func() {
+	var (
+		event   string
+		message string
+	)
+
+	BeforeEach(func() {
+		event = "sendMessage"
+		message = "A message in a bottle"
+	})
```
lmars: It looks like these variables don't change, so I would drop these assignments and make them constants:

```go
const (
	event   = "sendMessage"
	message = "A message in a bottle"
)

rest/client.go
mattheworiordan: I'll leave @kouno to respond, but I am pretty sure this relates to a very odd issue we have with our Comodo certificate, we should add a comment if this is the case.

kouno: I think @lmars already knows about it. I believe I saw some code he wrote with this line before.

In case he doesn't it's well explained here: http://bridge.grumpy-troll.org/2014/05/golang-tls-comodo/. TL;DR is that we have an SSL certificate that is newer with a different encryption and this is to support this new SSL certificate.

On the other hand I may have not needed this line in `rest/rest_suite_test.go` and `test/support/test_app.go`. I'll crosscheck and remove if necessary.

mattheworiordan:
Nice, glad to see you're making progress.  Will this be finished by the end of this week then? ;)

PR #2, opened 2015-02-26
Structure change

kouno:
@lmars Would you mind taking a look into this?

To keep things more or less consistent with other libraries, I had to find a way to have an equivalent of `Ably::Rest::Client.new()` or `new Ably.Rest.Client()`. So I created a struct in the `ably` namespace for `Rest` which doesn't do anything special apart from being a fake sub-package and allow me to call `ably.Rest.NewClient()`.

We could accept that people will call `rest.Client` like we did in most tests but I don't think that it conveys the right meaning (it's an ably client, not just an HTTP client). Unless you think we should advise to rename package on import (`import ably "github.com/ably/ably-go"`), but again I don't think it's conventional.

Thoughts?


kouno:
cc/ @mattheworiordan

lmars:
I'm not sure I like the `Rest` and `Realtime` structs, how about `ably.NewRestClient()` and `ably.NewRealtimeClient()`?

I also see you have moved the params into a `config` package, which suffers the same fate of not being called `ably` when imported.

I think the usage should be:

```go
import "github.com/ably/ably-go"

client := ably.NewRestClient(ably.Config{...})

File: README.md [hunk 1]
```diff
@@ -1,29 +1,120 @@
-# Ably Go Client Library
+# [Ably](https://ably.io)
 
-A Go client library for https://ably.io, the real-time messaging service.
+[![Build Status](https://travis-ci.org/ably/ably-go.png)](https://travis-ci.org/ably/ably-go)
+[![Gem Version](https://badge.fury.io/rb/ably.svg)](http://badge.fury.io/rb/ably)
```
lmars: This should be dropped

kouno: Oops I forgot that was a Ruby thing. :)

kouno:
> I also see you have moved the params into a config package, which suffers the same fate of not being called ably when imported.

Exactly. I'm not happy with this. I even find `config` to be too common as a package name... If I could I would have placed it under `ably` but Go would find that I have recursive dependencies between `rest` and `ably`... So not a great choice.

> I'm not sure I like the Rest and Realtime structs, how about ably.NewRestClient() and ably.NewRealtimeClient()?

Sounds like I'll have to accept that I can't have sub-package in golang... I'll update with `ably.NewRestClient()`

kouno:
@lmars updated and I made it so that `config.Params` is wrapped in `ably.ClientConfig`. I'll probably merge soon as I need to continue, but feel free to leave feedback that I'll address :).

mattheworiordan:
This is clearly my lack of Go understanding, but why can't we implement something more consistent with all other libraries such as `client := ably.Rest.Client(ably.Config{...})`?

mattheworiordan:
BTW. I can't see the old commits so can't comment on @kouno's first solution.

lmars:
@mattheworiordan that approach is possible with @kouno's original implementation, but it is not idiomatic, so I guess we need to make a decision what is more important; idiomatic Go code or consistency across libraries?

mattheworiordan:
@lmars, would it be more idiomatic to instead import the Rest or Realtime library directly, so perhaps something like:
```
import "github.com/ably/ably-go/realtime";
client = realtime.NewClient(ably.Config{...});
```
We could perhaps still offer the `ably.NewRestClient` for people who import `/ably/ably-go`, but I suspect 99% of the time, people will either be using Realtime or REST and not both in the same file they are working on.

WDYT?

kouno:
> This is clearly my lack of Go understanding, but why can't we implement something more consistent with all other libraries such as client := ably.Rest.Client(ably.Config{...})?

You have no constructor in Golang and all happens by creating a `NewXxxx` that handles instantiating the object. So I had to do something like:

```go
// In ably.go
import "github.com/ably/ably-go/rest"

type Rest struct {}

func (r *Rest) NewClient(config ably.ClientConfig) *rest.Client {
  return rest.NewClient(config.ToParams())
}
```

Note that it creates a useless variable `r`, and eventually that's just a trick to get the `rest` package to be available on the ably package.

> @mattheworiordan that approach is possible with @kouno's original implementation, but it is not idiomatic, so I guess we need to make a decision what is more important; idiomatic Go code or consistency across libraries?

I personally prefer to keep it Go-friendly as much as possible. We already have specifics in different languages... For example our Ruby library doesn't implement the listeners that Java needs, and it uses `EventMachine` to handle concurrency. So we may as well keep it consistent with the usual go guidelines.

kouno:
> We could perhaps still offer the ably.NewRestClient for people who import /ably/ably-go, but I suspect 99% of the time, people will either be using Realtime or REST and not both in the same file they are working on.

Well, I prefer having it in `ably` for clarity when using the library. `rest` and `realtime` are not telling that they are using a third party library in their code.

Just like I would expect NewRelic to have a `newrelic` package that I can get my objects from, and not just `stats` for example.

kouno:
Also, to be fair... Your scenario is by default covered because Go allows you to import package directly... But I suppose you were more concerned on what approach we are advising in our documentation, right?

mattheworiordan:
Yes, exactly, what is our preferred approach. In many ways I prefer the conciseness of the following:
```
import "github.com/ably/ably-go/rest";
client = rest.NewClient(ably.Config{...});
```

But then on the downside, the word `ably` is now not in the test so instead you refer to it as `rest`, so perhaps the below is indeed better, even if it's different from other libs.
```
import "github.com/ably/ably-go";
client = ably.NewRealtimeClient(ably.Config{...});
```

PR #4, opened 2015-03-08
Rest presence

kouno:
Get presence and presence history. I added the paginated result struct that will also be used in `channel.History()`.

I don't cover encryption yet, and I will certainly move on to adding the pagination to history first before hitting the encryption/auth subject.

As a side note, since overloading is not a thing in go, the normal call to `Presence.Get` and `Presence.History` have a `PaginatedPresenceMessages` parameter. Anyway the java library seems to do the same and even use `asArray()` to transform the paginated object in an array of messages. I use `paginatedPresenceMessages.Current` but I could just do the same as the java library on this point if necessary.

cc/ @mattheworiordan @paddybyers

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -16,6 +16,10 @@
 			"Rev": "736b4d597fe39e75c9777942245ee02a24bb0d26"
 		},
 		{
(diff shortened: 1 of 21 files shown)


mattheworiordan:
Nice, making progress...

PR #5, opened 2015-03-22
Add support to encode data in messages

kouno:
@mattheworiordan

### DIFF
--- a/proto/message.go
+++ b/proto/message.go
@@ -17,6 +17,10 @@ type Message struct {
 	Encoding string `json:"encoding,omitempty" msgpack:"encoding,omitempty"`
 }
 
+const (
+	PaddingStr = "\x0b"
+)
+
 func (m *Message) DecodeData(keys map[string]string) error {
 	encodings := strings.Split(m.Encoding, "/")
 	for i := len(encodings) - 1; i >= 0; i-- {
@@ -119,7 +123,7 @@ func (m *Message) Decrypt(cipherStr string, keys map[string]string) error {
 		blockMode := cipher.NewCBCDecrypter(block, []byte(iv))
 		blockMode.CryptBlocks(out, []byte(m.Data))
 
-		m.Data = strings.TrimRight(string(out), "\x0b")
+		m.Data = strings.TrimRight(string(out), PaddingStr)
 	default:
 		// TODO log wrong keylen
 		// Golang supports only these previous specified keys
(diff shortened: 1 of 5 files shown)


protocol/message.go
```diff
@@ -93,3 +127,55 @@ func (m *Message) Decrypt(cipherStr string, keys map[string]string) error {
 
 	return nil
 }
+
+func (m *Message) Encrypt(cipherStr string, keys map[string]string) error {
```
kouno: `keys map[string]string` probably need to just be `key string`. I thought I would need the `iv` at some point but it is all random.

kouno:
I also hear I need to support hmac for this to be valid. I'll try to find out what I need to do in order to do that.

protocol/message.go
mattheworiordan: Yup, just one string is probably better.  I do have an option to optionally pass in an `iv`, but in reality that's not a critical feature.

mattheworiordan:
The HMAC, as far as I am aware, is only used for token authentication and signing requests.  See https://github.com/ably/ably-ruby/blob/6efdfdc0f484bc0f42f8d39b844e3d8630208343/lib/ably/auth.rb#L355-L368.

Also, I strongly recommend your tests test against the fixture data at https://github.com/ably/ably-ruby/blob/6efdfdc0f484bc0f42f8d39b844e3d8630208343/spec/resources/crypto-data-256.json so that we can ensure binary compatibility between this library and others.

mattheworiordan:
Nice work, getting there

kouno:
@mattheworiordan updated this PR to also cover presence messages. The encode/decode still doesn't happen by default.

protocol/message.go
```diff
@@ -41,55 +52,180 @@ func (m *Message) DecodeData(keys map[string]string) error {
...
 }
 
-func (m *Message) Decrypt(cipherStr string, keys map[string]string) error {
+// EncodeData resets the current Encoding field to an empty string and starts
+// encoding data following the given `encoding` parameter.
+// `encoding` contains slash (/) separated values that EncodeData will read
+// from left to right to encode the current Data string.
+// To encode data using aes, the keys parameter must be present. Otherwise,
+// EncodeData will return an error.
+func (m *Message) EncodeData(encoding string, keys map[string]string) error {
```
rjeczalik: since `(*Message).Encrypt` requires fixed number of args - the `key` and vector `iv` shouldn't the signature be:

```go
func (m *Message) EncodeData(encoding string, key, iv []byte) error
```

rjeczalik:
LGTM

@kouno are you going to copy `crypto-data-*.json` files into ably-go repo? I need to copy `ably-common/protocol/errors.json` and I'm thinking of common place to put them - top-level `testdata` directory?

kouno:
I would usually throw these things in a test folder. Maybe `test/support/resources/`? I don't think I have seen anything in any go library that would point to a convention so it's really up to you :smile:.

rjeczalik:
@mattheworiordan I'll take care of sending errors back to the realtime channels. I'll also take care of error handling - by this I mean creating & exporting error variables out of [the def](https://github.com/ably/ably-common/blob/master/protocol/errors.json).

@kouno How about this PR? If you want to avoid conflicts it'd would be good to merge it before #13. The reason I'm asking a want to change the `Data string -> Data interface{}` so `proto` can handle unmarshaling payloads (#9).

kouno:
I rebased and I will merge it since I won't be able to work on it until later this week or next week :/.

The last piece of this PR which is not done yet is the usage of our common fixture files (so that we can guarantee inter-library compatibility). I think we have not added the submodule yet, but that would only be a refactor of the tests AFAICT.

rjeczalik:
@kouno I'll take care of submodules and testing encoding with common fixtures.

PR #6, opened 2015-04-08
Fix capability

rjeczalik:
Fix for broken Capability marshalling/unmarshalling on master.

### DIFF
--- a/rest/auth.go
+++ b/rest/auth.go
@@ -16,7 +16,7 @@ import (
 
 type Capability map[string][]string
 
-func (c *Capability) String() string {
+func (c Capability) String() string {
 	b, err := json.Marshal(c)
 	if err != nil {
 		panic(err)
@@ -25,9 +25,9 @@ func (c *Capability) String() string {
 }
 
 type Token struct {
-	ID         string      `json:"id"`
-	Key        string      `json:"key"`
-	Capability *Capability `json:"capability"`
+	ID         string `json:"id"`
+	Key        string `json:"key"`
+	Capability string `json:"capability"`
 }
 
 type tokenResponse struct {
@@ -72,7 +72,7 @@ func NewAuth(params config.Params, client *Client) *Auth {
 	return auth
 }
 
-func (a *Auth) CreateTokenRequest(ttl int, capability *Capability) *TokenRequest {
+func (a *Auth) CreateTokenRequest(ttl int, capability Capability) *TokenRequest {
 	req := &TokenRequest{
 		ID:         a.AppID,
 		TTL:        ttl,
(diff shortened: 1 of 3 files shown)


rest/auth.go
```diff
@@ -35,29 +50,31 @@ type tokenResponse struct {
...
-	params := []string{
-		t.ID,
-		strconv.Itoa(t.TTL),
-		t.Capability,
-		t.ClientID,
-		strconv.Itoa(int(t.Timestamp)),
-		t.Nonce,
-	}
-	s := strings.Join(params, "\n") + "\n"
+var nl = []byte{'\n'}
```
kouno: Shouldn't this be within the `Sign` function anyway?

rest/auth.go
```diff
@@ -35,29 +50,31 @@ type tokenResponse struct {
...
+	mac.Write([]byte(strconv.Itoa(t.TTL)))
+	mac.Write(nl)
+	mac.Write([]byte(t.Capability.String()))
+	mac.Write(nl)
+	mac.Write([]byte(t.ClientID))
+	mac.Write(nl)
+	mac.Write([]byte(strconv.Itoa(int(t.Timestamp))))
+	mac.Write(nl)
+	mac.Write([]byte(t.Nonce))
+	mac.Write(nl)
```
kouno: Hum... isn't replacing the array and the string a micro optimization? Also it's less readable IMO.

kouno:
Looks good :+1:.

rest/auth.go
rjeczalik: I was rather aiming for making this readable that optimising - for me series of Writes are easier to follow than []string + strings.Join(..., "\n") + single write.

If it's less readable for you, I'll revert this change.

rest/auth.go
rjeczalik: The nl will be gone after revert.

rest/auth.go
kouno: It would be as readable if that wasn't interleaved with `nl` :).

rjeczalik:
@kouno Please take a look.

rest/auth.go
```diff
@@ -35,29 +51,23 @@ type tokenResponse struct {
...
-
 	mac := hmac.New(sha256.New, []byte(secret))
-	mac.Write([]byte(s))
-
+	fmt.Fprintln(mac, t.ID)
+	fmt.Fprintln(mac, t.TTL)
+	fmt.Fprintln(mac, t.Capability.String())
+	fmt.Fprintln(mac, t.ClientID)
+	fmt.Fprintln(mac, t.Timestamp)
+	fmt.Fprintln(mac, t.Nonce)
```
kouno: Nice :+1:.

kouno:
Nice work.

rest/auth.go
mattheworiordan: It's all ugly to me ;)

rest/auth.go
mattheworiordan: Ahhh, that's nice ;) :smile:

PR #7, opened 2015-04-08
Don't verify hostname if HTTP_PROXY is set

rjeczalik:
Non-invasive CL for tests, to make them work with `HTTP_PROXY` env - useful for inspecting payloads with MITM proxies.

### DIFF
--- a/test/support/test_app.go
+++ b/test/support/test_app.go
@@ -2,11 +2,13 @@ package support
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/ably/ably-go/config"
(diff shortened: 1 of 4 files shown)


test/support/test_app.go
```diff
@@ -131,24 +132,31 @@ func (t *TestApp) Delete() (*http.Response, error) {
...
+				Timeout:   timeout,
+				KeepAlive: timeout,
+			}).Dial,
+			TLSHandshakeTimeout: timeout,
+		},
 	}
 
-	client := http.Client{
-		Transport: &transport,
+	// Allow for HTTPS traffic inspection with MITM proxies - don't verify host names.
```
kouno: I'm not sure we should specify MITM in our comments.

rjeczalik:
I've dropped all references to mitm.

kouno:
Can you rebase this on top of `master` so I can just merge it?

rjeczalik:
Done.

PR #10, opened 2015-04-12
Rework proto.PaginatedResource API

rjeczalik:
According to https://github.com/ably/ably-js/issues/28.

`NewPaginatedResource`is going to be unexported after I merge the `config`, `rest` and `realtime` packages into single `ably` one.

The new API is:

```go
type PaginatedResource struct {
    // contains filtered or unexported fields
}
    PaginatedResource represents a single page coming back from the REST
    API. Any call to create a new page will generate a new instance.

func NewPaginatedResource(typ reflect.Type, path string, params *config.PaginateParams, query QueryFunc) (*PaginatedResource, error)
    NewPaginatedResource returns a new instance of PaginatedResource It
    needs to be a struct implementing ResourceReader which in our case is
    rest.Client.

func (p *PaginatedResource) Items() []interface{}
    Items gives a slice of results of the current page.

func (p *PaginatedResource) Messages() []*Message
    Messages gives a slice of messages for the current page. The method
    panics if the underlying paginated resource is not a message.

func (p *PaginatedResource) Next() (*PaginatedResource, error)
    Next returns the path to the next page as found in the response headers.
    The response headers from the REST API contains a relative link to the
    next resource (Link: <./path>; rel="next").

func (p *PaginatedResource) PresenceMessages() []*PresenceMessage
    PresenceMessages gives a slice of presence messages for the current
    path. The method panics if the underlying paginated resource is not a
    presence message.

func (p *PaginatedResource) Stats() []*Stat
    Stats gives a slice of statistics for the current page. The method
    panics if the underlying paginated resource is not statistics.
```


proto/paginated_resource.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	nextPage, err := p.buildPath(p.path, nextPath)
+	if err != nil {
+		return nil, err
+	}
+	return NewPaginatedResource(p.typ, nextPage, nil, p.query)
+}
+
+// Items gives a slice of results of the current page.
+func (p *PaginatedResource) Items() []interface{} {
```
kouno: We probably don't need to provide this function anymore since we have all types available.

kouno:
Looks good. :+1:

Can you review the two points I raised and I'll merge it ASAP.

rjeczalik:
I've removed `Items()`. I understood we want to allow for generic access to `PaginatedResource`. We can always bring it back later if needed.

proto/paginated_resource.go
rjeczalik: Done.

rjeczalik:
@kouno Please take a look.

mattheworiordan:
@rjeczalik this looks great, nice work.  Can I ask why `items` is private though?  I certainly think that exposing, for example, `PresenceMessages` as a function is indeed correct given generics are not supported in Go, however for consistency with other client libraries, which is one of our goals, should we not provide the generic `Items` function as well even if we don't encourage people to use it?

rjeczalik:
@mattheworiordan Please take a look: #11

PR #11, opened 2015-04-12
Add (*PaginatedResource).Items() for accessing interface{} items

rjeczalik:


mattheworiordan:
Nice, no tests I see, confident coding ;)

rjeczalik:
@mattheworiordan I've added checks for `Items()` to complement existing tests. The real problem is there're no tests for `PaginatedResource`, which is a minor problems, more important are functional tests, and those do only check length of returned slices and no their content. I know it's not an excuse, I'll revisit all tests after completing realtime client. Sounds reasonable?

Moreover I'm not sure what are you using for coverage reporting. If you're open to suggestion, I'd propose setting an account for ably (free, open-source one) on https://coveralls.io and pass me a token for ably-go. Then I could add building reports to Travis and badge to README.md, so we can see the % growing :)

kouno:
@mattheworiordan I'm okay to bring the `Items` getter back for the sake of consistency, but is this the only reason we do so?

@rjeczalik sorry to make you work more for nothing :stuck_out_tongue:

rjeczalik:
@kouno Not an issue at all, it's just a matter of git cherry-pick and 3-way merge :P

>  (...) but is this the only reason we do so?

Sorry for chiming in, but I could imagine a use-case, where user may want to leverage `Items()`, e.g. to build some kind of serialisation on top of `*ably.PaginatedResource`. Normally user would end up with the following interface:

```go
type ResourceSerializer interface {
	Serialize(first *ably.PaginatedResource, typ ResourceType) error
}
```

The `typ` enum is required, so the user know which method to call safely. Now, by having generic `Items()`, the interface and the code could be simplified to:

```go
type ResourceSerializer interface {
	Serialize(first *ably.PaginatedResource) error
}
```

So I feel the `Items()` may be quite orthogonal to the rest of the methods.

kouno:
Fair enough :+1:. I doubt a lot of people would use it, but I don't think I have a valid point to _not_ let them do so.

mattheworiordan:
@kouno, yes the only reason is for API consistency, however as @rjeczalik has said, there is a possibility that someone would like to work with PaginatedResource in a generic way and as such `Items` would be useful.

mattheworiordan:
@rjeczalik I have added coveralls.io support, see https://github.com/ably/ably-ruby/pull/27 and https://coveralls.io/r/ably/ably-go.  @rjeczalik want to add it to see if it works?

mattheworiordan:
@rjeczalik btw please add the following to the Readme:

```[Coverage Status](https://coveralls.io/r/ably/ably-go)```

rjeczalik:
@kouno @mattheworiordan Are we ok with merging this or sth more needs to be done? As it goes for coveralls, I"ll add it to Travis in separate commit.

kouno:
Sounds like it. Thanks. :+1:

PR #13, opened 2015-04-13
Merge config, rest and realtime into one ably package

rjeczalik:
This CL:

- moves config.Params to ably.Params
- moves proto.PaginatedResource to ably.PaginatedResource
- moves {rest,realtime}.Client to ably.{Rest,Realtime}Client to remove
  circular deps on existing ably.New{Rest,Realtime}Client functions


kouno:
Well I guess we must embrace the single package philosophy... :cry:

ably/paginated_resource.go
```diff
@@ -40,8 +40,8 @@ type PaginatedResource struct {
 }
 
 // NewPaginatedResource returns a new instance of PaginatedResource
-// It needs to be a struct implementing ResourceReader which in our case is rest.Client.
-func NewPaginatedResource(typ reflect.Type, path string, params *config.PaginateParams,
+// It needs to be a struct implementing ResourceReader which in our case is rest.RestClient.
```
kouno: The comment still mention `rest` as the package name.

ably/ably_suite_test.go
```diff
@@ -30,7 +29,7 @@ var _ = BeforeSuite(func() {
 
 var _ = BeforeEach(func() {
 	client = ably.NewRestClient(testApp.Params)
-	channel = client.Channel("test")
+	channel = client.RestChannel("test")
```
kouno: It sounds weird to write RestChannel when we already have specified that we instantiated a RestClient. Shouldn't that just be `client.Channel` anyway?

ably/presence.go
```diff
@@ -1,27 +1,22 @@
...
-import (
-	"github.com/ably/ably-go/config"
-	"github.com/ably/ably-go/proto"
-)
+package ably
 
 type Presence struct {
-	client  *Client
-	channel *Channel
+	client  *RestClient
```
kouno: You do have Presence on the realtime side too, so I would probably take the occasion to rename presence `RestPresence` and add a getter on `RestChannel` for `Presence`.

ably/params.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+type ProtocolType string
+
+const (
+	ProtocolJSON    = "json"
+	ProtocolMsgPack = "msgpack"
+)
+
+type Params struct {
```
mattheworiordan: Params is quite generic, should this not perhaps be named in such a way that a developer knows what the `Params` are for?

ably/params.go
```diff
@@ -0,0 +1,137 @@
...
+	RealtimeEndpoint string
+	RestEndpoint     string
+
+	ApiKey       string
+	ClientID     string
+	AppID        string
+	AppSecret    string
```
mattheworiordan: There is no such thing as an `AppSecret`, this is actually `ApiSecret`.  I expect `AppId` is supposed to be `ApiId`

ably/params.go
```diff
@@ -0,0 +1,137 @@
...
+	UseTokenAuth bool
+
+	Protocol ProtocolType
+	Tls      bool
```
mattheworiordan: This should default to true

ably/params.go
```diff
@@ -0,0 +1,137 @@
...
+const (
+	ProtocolJSON    = "json"
+	ProtocolMsgPack = "msgpack"
+)
+
+type Params struct {
+	RealtimeEndpoint string
+	RestEndpoint     string
+
+	ApiKey       string
```
mattheworiordan: This is just called `Key`

kouno:
Can you also update the README  to reflect that change?

mattheworiordan:
@rjeczalik I assume this style of putting everything in the same namespace is more normal for Go?

ably/params.go
kouno: These comments are valid but they all stem from the fact that it is code that was already in that state before the refactor.

@rjeczalik You should address them in another PR. Let's focus on getting this package convention out first.

mattheworiordan: Ah, OK, well I'll create a new issue to address all of these changes + others from https://github.com/ably/ably-js/issues/33

ably/paginated_resource.go
rjeczalik: It mentioned also non-existing `ResourceReader` interface. The `newPaginatedResource` is unexported now, so I've removed the doc.

ably/ably_suite_test.go
rjeczalik: Done.

ably/presence.go
rjeczalik: Done. I'd want to refactor it further - #15.

ably/params.go
rjeczalik: Moved to #15.

ably/params.go
rjeczalik: Moved to #15.

ably/params.go
rjeczalik: Moved to #15.

ably/params.go
rjeczalik: Moved to #15.

rjeczalik:
@kouno @mattheworiordan Please take a look.

> I assume this style of putting everything in the same namespace is more normal for Go?

There's no rule nor practice of putting everything in one package no matter what. The normal for Go is to structure the packages around functionalities - functionalities in ably-go depends on each other, and with four separate packages it was a bit hard to keep the API clean. This single-package API is cleaner, however there's no issue revisiting it after it's complete.

kouno:
> This single-package API is cleaner

Ahah, once again a victory for @lmars who wrote it in a similar way before I jumped on the project. I hate when this guy is right ;).

kouno:
Actually, I said nothing. Seems like @lmars did create the first rest/realtime package too! https://github.com/ably/ably-go/tree/90c6d6341c595270608df67b2622400de63e6359

:confetti_ball: :fireworks:

ably/rest_channel.go
```diff
@@ -1,24 +1,21 @@
...
-func newChannel(name string, client *Client) *Channel {
-	c := &Channel{
+func newRestChannel(name string, client *RestClient) *RestChannel {
+	c := &RestChannel{
 		Name:   name,
 		client: client,
 	}
 
-	c.Presence = &Presence{
+	c.Presence = &RestPresence{
```
kouno: Can you add to the other issue that `Presence` needs to be a getter (`presence()`). This is to restrict the access to the `Presence` object.

rjeczalik: Aleady done, see second bullet here #15.

lmars:
@kouno yep I originally had two packages because there were two clients, and there was a straight forward dependency from `rest` -> `realtime`, but maybe that is no longer the case :wink:.

Feel free to mention me in other PRs if you want another opinion :smile:.

PR #18, opened 2015-04-18
Refactor proto to use []byte for binary data.

rjeczalik:
And fix encoding unpadded binary payloads which happen to be
`n * aes.BlockSize` long.

### DIFF
--- a/ably/proto/message.go
+++ b/ably/proto/message.go
@@ -1,11 +1,11 @@
 package proto
 
 import (
-	"bytes"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
 	"strconv"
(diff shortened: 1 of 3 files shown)


mattheworiordan:
This looks good @rjeczalik, however I do question why all the encoding & encryption functionality is even part of the `Message` proto package.  Surely encoding & decoding is a separate concern and a utility instead?  For example, in the Ruby library I have a set of encoders that are simply applied to any `Message` or `PresenceMessage` objects to apply encoding or decode.  See the [encodeable mixin](https://github.com/ably/ably-ruby/blob/master/lib/ably/modules/encodeable.rb) and the [encoders](https://github.com/ably/ably-ruby/tree/master/lib/ably/models/message_encoders).  I am not suggesting you follow that pattern exactly, but keeping this logic out of the Message itself feels better to me.  @kouno and @lmars what do you think?

rjeczalik:
@mattheworiordan I cannot agree more. If we all agree models & encoders/decoders should be separated, I'll create a separate issue for this.

mattheworiordan:
Certainly :+1: on my side

lmars:
I haven't looked at the code thoroughly, but I would probably expect something like the following to decode messages from a reader:

```go
res := client.Get("/stream")
dec := NewDecoder(res.Body, "json/utf-8")
for {
	var msg proto.Message
	if err := dec.Decode(&msg); err != nil {
		return err
	}
	// do something with msg
}
```

kouno:
Yes. Sounds like a plan. I didn't have time to refactor it into a beautiful separate decoder/encoder, but it would definitely help.

As for the PR, it looks good. I'm merging this and you'll refactor the encoding part later ;).

PR #24, opened 2015-04-25
rest: improve RestClient API

rjeczalik:
This CL cleans up RestClient API, fixing couple of incosistencies and
shortcommings, like:

  - when (*RestClient).Options is changed by the user, changes are
    not reflected in *Auth
  - all the tests were modyfing global client instance, not it's more
    COW like
  - ~~RestClient and ClientOptions are fully zero-value friendly~~
  - ClientOptions properly manages TLS / NoTLS transitions

RestClient would require more work, like support for token auth
(it's using BasicAuth for requests), proper TLS/NoTLS testing. This is
what I've seen from the code itself, possibly there's more if compared
to SPEC.md.

~~Changes in Go API:~~

- ~~added function for initializing RestClient from key:~~

  ```
  client, err := ably.NewRestClient("xxx:xxx")
  ```

- ~~RestClient can be initialized from ClientOptions for more control
  over behaviour: (added correct handling of zero-value struct)~~

  ```
  opts := &ably.ClientOptions{Token: "xxx", Secret: "xxx"}
  client := &alby.RestClient{Options: opts}
  ```

- ~~ClientOptions has no `Key` nor `NoTLS` fields, as it's not possible to
  keep them in sync with dependant fields like Token, Secret for Key or
  RestEndpoint, RealtimeEndpoint for `NoTLS`, e.g.~~

  ```go
  /* old API */
  client := ably.NewRestClient(&ably.Options{Key: "xxx:xxx"})
  // ...
  client.Options.Key = "yyy:yyy"
  // all calls to REST API are going to fail, as Token and Secret fields
  // are still "xxx"

  /* new API */
  client, err := ably.NewRestClient("xxx:xxx")
  // ...
  err := client.Options.SetKey("yyy:yyy")
  // correctly handled here
  ```

~~The same applies to NoTLS, where flipping its value should rewrite URLs
of RestEndpoint and RealtimeEndpoint. With new API it's done with:~~

  ```go
  client.SetTLS(false)
  ```

/cc @kouno Please take a look.

BTW I have trello dashboard for ably-go, as I needed to store somewhere my thoughts while working with current library: https://trello.com/b/pHESio8l/ably-go

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,11 @@ TODO
 All examples assume a client and/or channel has been created as follows:
 
 ```go
-client := ably.NewRestClient(ably.ClientOptions{ApiKey: "xxxx"})
-channel := client.Channel('test')
+client, err := ably.NewRestClient(&ably.ClientOptions{Key: "xxx:xxx"})
+if err != nil {
+	panic(err)
+}
+channel := client.Channel("test")
 ```
 
 ### Publishing a message to a channel
--- a/ably/auth_test.go
+++ b/ably/auth_test.go
@@ -18,7 +18,7 @@ var _ = Describe("Auth", func() {
 
 			Expect(err).NotTo(HaveOccurred())
 			Expect(token.Token).To(ContainSubstring(testApp.Config.ApiID))
-			Expect(token.KeyName).To(Equal(testApp.AppKeyId()))
+			Expect(token.KeyName).To(Equal(testApp.KeyName()))
 			Expect(token.Issued).NotTo(Equal(int64(0)))
 			Expect(token.Capability).To(Equal(req.Capability))
 		})
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -1,6 +1,6 @@
 package ably
 
-func (p *PaginatedResource) BuildPath(base, rel string) (string, error) {
+func (p *PaginatedResult) BuildPath(base, rel string) (string, error) {
 	return p.buildPath(base, rel)
 }
 
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -1,5 +1,13 @@
 package ably
 
-var BuildPath = func(p *PaginatedResource, base, rel string) (string, error) {
+func (p *PaginatedResource) BuildPath(base, rel string) (string, error) {
 	return p.buildPath(base, rel)
+}
+
+func (opts *ClientOptions) RestURL() string {
+	return opts.restURL()
+}
+
+func (opts *ClientOptions) RealtimeURL() string {
+	return opts.realtimeURL()
 }
--- a/ably/options.go
+++ b/ably/options.go
@@ -1,6 +1,7 @@
 package ably
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -14,36 +15,74 @@ const (
 	ProtocolMsgPack = "msgpack"
 )
 
+var errInvalidKey = errors.New("invalid key format")
+
+var DefaultOptions = &ClientOptions{
+	RestHost:     "rest.ably.io",
+	RealtimeHost: "realtime.ably.io",
+	Protocol:     ProtocolJSON, // TODO: make it ProtocolMsgPack
+}
+
 type ClientOptions struct {
-	RealtimeEndpoint string
-	RestEndpoint     string
-	Key              string
-	Token            string
-	Secret           string
-	ClientID         string
-	Protocol         string
-	UseTokenAuth     bool
-	NoTLS            bool
+	RestHost     string
+	RealtimeHost string
+	Key          string
+	ClientID     string
+	Protocol     string // either ProtocolJSON or ProtocolMsgPack
+	UseTokenAuth bool
+	NoTLS        bool
 
 	HTTPClient *http.Client
 }
 
-func (p *ClientOptions) Prepare() {
-	if p.Key != "" {
-		p.parseKey()
+func (opts *ClientOptions) restURL() string {
+	host := opts.RestHost
+	if host == "" {
+		host = DefaultOptions.RestHost
 	}
+	if opts.NoTLS {
+		return "http://" + host
+	}
+	return "https://" + host
 }
 
-func (p *ClientOptions) parseKey() {
-	keyParts := strings.Split(p.Key, ":")
-
-	if len(keyParts) != 2 {
-		Log.Print(LogError, "invalid key format, ignoring")
-		return
+func (opts *ClientOptions) realtimeURL() string {
+	host := opts.RealtimeHost
+	if host == "" {
+		host = DefaultOptions.RealtimeHost
 	}
+	if opts.NoTLS {
+		return "ws://" + host + ":80"
+	}
+	return "wss://" + host + ":443"
+}
 
-	p.Token = keyParts[0]
-	p.Secret = keyParts[1]
+func (opts *ClientOptions) KeyName() string {
+	if i := strings.IndexRune(opts.Key, ':'); i != -1 {
+		return opts.Key[:i]
+	}
+	return ""
+}
+
+func (opts *ClientOptions) KeySecret() string {
+	if i := strings.IndexRune(opts.Key, ':'); i != -1 {
+		return opts.Key[i+1:]
+	}
+	return ""
+}
+
+func (opts *ClientOptions) httpclient() *http.Client {
+	if opts.HTTPClient != nil {
+		return opts.HTTPClient
+	}
+	return http.DefaultClient
+}
+
+func (opts *ClientOptions) protocol() string {
+	if opts.Protocol != "" {
+		return opts.Protocol
+	}
+	return ProtocolJSON // TODO: make it ProtocolMsgPack
 }
 
 // Timestamp returns the given time as a timestamp in milliseconds since epoch.
@@ -68,19 +107,15 @@ func (s *ScopeParams) EncodeValues(out *url.Values) error {
 	if s.Start != 0 && s.End != 0 && s.Start > s.End {
 		return fmt.Errorf("start must be before end")
 	}
-
 	if s.Start != 0 {
 		out.Set("start", strconv.FormatInt(s.Start, 10))
 	}
-
 	if s.End != 0 {
 		out.Set("end", strconv.FormatInt(s.End, 10))
 	}
-
 	if s.Unit != "" {
 		out.Set("unit", s.Unit)
 	}
-
 	return nil
 }
 
@@ -96,7 +131,6 @@ func (p *PaginateParams) EncodeValues(out *url.Values) error {
 	} else if p.Limit != 0 {
 		out.Set("limit", strconv.Itoa(p.Limit))
 	}
-
 	switch p.Direction {
 	case "":
 		break
@@ -106,8 +140,6 @@ func (p *PaginateParams) EncodeValues(out *url.Values) error {
 	default:
 		return fmt.Errorf("Invalid value for direction: %s", p.Direction)
 	}
-
 	p.ScopeParams.EncodeValues(out)
-
 	return nil
 }
--- a/ably/options_test.go
+++ b/ably/options_test.go
@@ -1,57 +1,39 @@
 package ably_test
 
 import (
-	"log"
 	"net/url"
 
 	"github.com/ably/ably-go/ably"
 
 	. "github.com/ably/ably-go/Godeps/_workspace/src/github.com/onsi/ginkgo"
 	. "github.com/ably/ably-go/Godeps/_workspace/src/github.com/onsi/gomega"
-	"github.com/ably/ably-go/Godeps/_workspace/src/github.com/onsi/gomega/gbytes"
 )
 
 var _ = Describe("ClientOptions", func() {
-	var (
-		options *ably.ClientOptions
-		buffer  *gbytes.Buffer
-	)
+	var err error
 
-	BeforeEach(func() {
-		buffer = gbytes.NewBuffer()
+	Context("when Key is valid", func() {
+		var options *ably.ClientOptions
 
-		options = &ably.ClientOptions{
-			Key: "id:secret",
-		}
+		BeforeEach(func() {
+			options = &ably.ClientOptions{Key: "name:secret"}
+			_, err = ably.NewRestClient(options)
+		})
 
-		options.Prepare()
-	})
-
-	It("parses Key into a set of known parameters", func() {
-		Expect(options.Token).To(Equal("id"))
-		Expect(options.Secret).To(Equal("secret"))
+		It("parses it into a set of known parameters", func() {
+			Expect(err).NotTo(HaveOccurred())
+			Expect(options.KeyName()).To(Equal("name"))
+			Expect(options.KeySecret()).To(Equal("secret"))
+		})
 	})
 
 	Context("when Key is invalid", func() {
-		var old *log.Logger
-
 		BeforeEach(func() {
-			old, ably.Log.Logger = ably.Log.Logger, log.New(buffer, "", log.Lmicroseconds|log.Llongfile)
-			options = &ably.ClientOptions{
-				Key: "invalid",
-			}
-
-			options.Prepare()
+			_, err = ably.NewRestClient(&ably.ClientOptions{Key: "invalid"})
 		})
 
-		AfterEach(func() {
-			ably.Log.Logger = old
-		})
-
-		It("prints an error", func() {
-			Expect(string(buffer.Contents())).To(
-				ContainSubstring("ERROR: invalid key format, ignoring"),
-			)
+		It("returns an error", func() {
+			Expect(err).To(HaveOccurred())
 		})
 	})
 })
--- a/ably/paginated_resource_test.go
+++ b/ably/paginated_resource_test.go
@@ -16,7 +16,7 @@ var _ = Describe("PaginatedResource", func() {
 		})
 
 		It("returns a string pointing to the new path based on the given path", func() {
-			newPath, err := ably.BuildPath(&paginatedResource, "/path/to/resource?hello", "./newresource?world")
+			newPath, err := paginatedResource.BuildPath("/path/to/resource?hello", "./newresource?world")
 			Expect(err).NotTo(HaveOccurred())
 			Expect(newPath).To(Equal("/path/to/newresource?world"))
 		})
--- a/ably/rest_client.go
+++ b/ably/rest_client.go
@@ -4,14 +4,16 @@ import (
 	"bytes"
 	_ "crypto/sha512"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"reflect"
 	"sync"
 	"time"
 
-	"github.com/ably/ably-go/Godeps/_workspace/src/gopkg.in/vmihailenco/msgpack.v2"
 	"github.com/ably/ably-go/ably/proto"
+
+	"github.com/ably/ably-go/Godeps/_workspace/src/gopkg.in/vmihailenco/msgpack.v2"
 )
 
 var (
@@ -27,36 +29,31 @@ func query(fn func(string, interface{}) (*http.Response, error)) QueryFunc {
 }
 
 type RestClient struct {
-	Auth *Auth
+	Auth     *Auth
+	Protocol string
+	Host     string
 
-	RestEndpoint string
-	Protocol     string
-
-	HTTPClient *http.Client
-
-	channels map[string]*RestChannel
-	chanMtx  sync.Mutex
+	chansMtx sync.Mutex
+	chans    map[string]*RestChannel
 }
 
-func NewRestClient(options ClientOptions) *RestClient {
-	options.Prepare()
-	client := &RestClient{
-		RestEndpoint: options.RestEndpoint,
-		HTTPClient:   options.HTTPClient,
-		channels:     make(map[string]*RestChannel),
+func NewRestClient(options *ClientOptions) (*RestClient, error) {
+	keyName, keySecret := options.KeyName(), options.KeySecret()
+	if keyName == "" || keySecret == "" {
+		return nil, errors.New("invalid key format")
 	}
-
-	client.Auth = NewAuth(&options, client)
-	client.Protocol = options.Protocol
-
-	return client
-}
-
-func (c *RestClient) httpclient() *http.Client {
-	if c.HTTPClient != nil {
-		return c.HTTPClient
+	c := &RestClient{
+		Protocol: options.protocol(),
+		Host:     options.restURL(),
+		chans:    make(map[string]*RestChannel),
 	}
-	return http.DefaultClient
+	c.Auth = &Auth{
+		options:   *options,
+		client:    c,
+		keyName:   keyName,
+		keySecret: keySecret,
+	}
+	return c, nil
 }
 
 func (c *RestClient) Time() (time.Time, error) {
@@ -66,21 +63,19 @@ func (c *RestClient) Time() (time.Time, error) {
 		return time.Time{}, err
 	}
 	if len(times) != 1 {
-		return time.Time{}, fmt.Errorf("Expected 1 timestamp, got %d", len(times))
+		return time.Time{}, fmt.Errorf("expected 1 timestamp, got %d", len(times))
 	}
 	return time.Unix(times[0]/1000, times[0]%1000), nil
 }
 
 func (c *RestClient) Channel(name string) *RestChannel {
-	c.chanMtx.Lock()
-	defer c.chanMtx.Unlock()
-
-	if ch, ok := c.channels[name]; ok {
+	c.chansMtx.Lock()
+	defer c.chansMtx.Unlock()
+	if ch, ok := c.chans[name]; ok {
 		return ch
 	}
-
 	ch := newRestChannel(name, c)
-	c.channels[name] = ch
+	c.chans[name] = ch
 	return ch
 }
 
@@ -92,27 +87,23 @@ func (c *RestClient) Stats(params *PaginateParams) (*PaginatedResource, error) {
 }
 
 func (c *RestClient) Get(path string, out interface{}) (*http.Response, error) {
-	req, err := http.NewRequest("GET", c.RestEndpoint+path, nil)
+	req, err := http.NewRequest("GET", c.Auth.options.restURL()+path, nil)
 	if err != nil {
 		return nil, err
 	}
 	req.Header.Set("Accept", "application/json")
-	req.SetBasicAuth(c.Auth.Options.Token, c.Auth.Options.Secret)
-	res, err := c.httpclient().Do(req)
-
+	req.SetBasicAuth(c.Auth.keyName, c.Auth.keySecret)
+	res, err := c.Auth.options.httpclient().Do(req)
 	if err != nil {
 		return nil, err
 	}
-
 	if !c.ok(res.StatusCode) {
-		return res, NewRestHttpError(res, fmt.Sprintf("Unexpected status code %d", res.StatusCode))
+		return res, NewRestHttpError(res, fmt.Sprintf("unexpected status code %d", res.StatusCode))
 	}
-
 	if out != nil {
 		defer res.Body.Close()
 		return res, json.NewDecoder(res.Body).Decode(out)
 	}
-
 	return res, nil
 }
 
@@ -121,21 +112,21 @@ func (c *RestClient) Post(path string, in, out interface{}) (*http.Response, err
 	if err != nil {
 		return nil, err
 	}
-	req, err := http.NewRequest("POST", c.RestEndpoint+path, bytes.NewBuffer(buf))
+	req, err := http.NewRequest("POST", c.Auth.options.restURL()+path, bytes.NewBuffer(buf))
 	if err != nil {
 		return nil, err
 	}
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Accept", "application/json")
-	req.SetBasicAuth(c.Auth.Options.Token, c.Auth.Options.Secret)
-	res, err := c.httpclient().Do(req)
+	req.SetBasicAuth(c.Auth.keyName, c.Auth.keySecret)
+	res, err := c.Auth.options.httpclient().Do(req)
 	if err != nil {
 		return nil, err
 	}
 	if !c.ok(res.StatusCode) {
-		return res, NewRestHttpError(res, fmt.Sprintf("Unexpected status code %d", res.StatusCode))
+		return res, NewRestHttpError(res, fmt.Sprintf("unexpected status code %d", res.StatusCode))
 	}
-	if out != nil && c.ok(res.StatusCode) {
+	if out != nil {
 		defer res.Body.Close()
 		return res, json.NewDecoder(res.Body).Decode(out)
 	}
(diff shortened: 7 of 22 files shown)


ably/auth_test.go
```diff
@@ -10,11 +10,11 @@ import (
 var _ = Describe("Auth", func() {
 	Describe("RequestToken", func() {
 		It("gets a token from the API", func() {
-			req := client.Auth.CreateTokenRequest()
+			req := client.Auth().CreateTokenRequest()
```
mattheworiordan: I'm interested to see that Auth is not a property / attribute, but instead a function.  Why is that? That is inconsistent with other client libraries at present because `Auth` is automatically created for every `Client`.

ably/options.go
```diff
@@ -14,36 +15,72 @@ const (
...
+)
+
+func (opts *ClientOptions) SetTLS(enabled bool) {
+	if enabled {
+		opts.RestEndpoint = restTLS.Replace(opts.RestEndpoint)
+		opts.RealtimeEndpoint = realtimeTLS.Replace(opts.RealtimeEndpoint)
+	} else {
+		opts.RestEndpoint = restNoTLS.Replace(opts.RestEndpoint)
+		opts.RealtimeEndpoint = realtimeNoTLS.Replace(opts.RealtimeEndpoint)
 	}
```
mattheworiordan: This feels like a rather cumbersome approach to what should be a simple task.  The end point should be constructed based on whether the connection is over TLS or not, along with the domain.  Why are we doing string replacing here and above when we could far more easily return an end point when requested that is constructed?

ably/options.go
```diff
@@ -14,36 +15,72 @@ const (
...
+	opts.Token = s[0]
+	opts.Secret = s[1]
+	return nil
+}
+
+func (opts *ClientOptions) Key() string {
+	if opts.Token == "" || opts.Secret == "" {
+		return ""
+	}
+	return opts.Token + ":" + opts.Secret
```
mattheworiordan: The token and key are two completely separate things?  `Token` is an expiring token that is created by a [requestToken](http://docs.ably.io/rest-api/#requesttoken) request, whereas a Key is made up of a key name and key secret, although mostly users will not be separating the two and simply passing a `key` option to the library when instanced.

ably/rest_client.go
```diff
@@ -92,27 +99,24 @@ func (c *RestClient) Stats(params *PaginateParams) (*PaginatedResource, error) {
...
+	req, err := http.NewRequest("GET", c.options().RestEndpoint+path, nil)
 	if err != nil {
 		return nil, err
 	}
 	req.Header.Set("Accept", "application/json")
-	req.SetBasicAuth(c.Auth.Options.Token, c.Auth.Options.Secret)
-	res, err := c.httpclient().Do(req)
-
+	// TODO(rjeczalik): add support for token auth
+	req.SetBasicAuth(c.options().Token, c.options().Secret)
```
mattheworiordan: This is not a token, it's an ApiKey and ApiSecret

ably/testutil/test_app.go
```diff
@@ -64,24 +73,32 @@ func (t *App) populate(res *http.Response) error {
 		return err
 	}
 
-	t.Options.Key = t.AppKeyId() + ":" + t.AppKeyValue()
+	t.opts.Token = t.AppKeyId()
```
mattheworiordan: KeyName & KeySecret

ably/options_test.go
```diff
@@ -1,61 +1,14 @@
...
-		buffer = gbytes.NewBuffer()
-
-		options = &ably.ClientOptions{
-			Key: "id:secret",
-		}
-
-		options.Prepare()
-	})
-
-	It("parses Key into a set of known parameters", func() {
```
mattheworiordan: Why are these tests removed?

ably/options.go
rjeczalik: It's just a fix, the REST client was implemented that way the Endpoints are URI instead of hostnames, but API did not provide method to switch between TLS/no TLS; I'll rewrite it to make Endpoints just hostnames.

ably/testutil/test_app.go
rjeczalik: opts.Token should be opts.ApiKey and t.AppKeyID should be t.ApiSecret, is it right?

ably/rest_client.go
rjeczalik: ack

ably/options.go
rjeczalik: the opts.Token should be renamed to opts.ApiKey or opts.KeyName? the same for Secret - ApiSecret or KeySecret?

ably/options_test.go
rjeczalik: Previous version: setting wrong key did not fail but only logged warning - the removed tests checked whether it there was text logged
New version: setting wrong key fails with an error

Ok, I'll add tests for checking whether setting invalid key gives an error.

ably/auth_test.go
rjeczalik: Done, changed back to field.

ably/options.go
rjeczalik: Done, RestEndpoint and RealtimeEndpoint are hostnames now and URLs are constructed internally based on NoTLS property (must be inverted as the only default we can have in Go for bool is false, thus NoTLS).

ably/options.go
rjeczalik: Done, Token renamed to KeyName and Secret to KeySecret (was confused a bit by ApiName / ApiSecret suggestion).

ably/rest_client.go
rjeczalik: Done, renamed to KeyName / KeySecret.

ably/testutil/test_app.go
rjeczalik: Done, renamed to KeyName / KeySecret.

ably/options_test.go
rjeczalik: Done, invalid key test-case now covered.

rjeczalik:
@mattheworiordan @kouno Please take a look.

ably/auth_test.go
kouno: Hum. I thought this was kind of in-line with the change needed for `Presence`. The only way for a property to be immutable outside the package you have to use accessors.

We discussed this for `Presence` to be `Presence()` instead, so if we enforce such a behaviour elsewhere, I would advise to be consistent with this one too.

README.md
```diff
@@ -31,8 +31,13 @@ TODO
...
 
 ```go
-client := ably.NewRestClient(ably.ClientOptions{ApiKey: "xxxx"})
-channel := client.Channel('test')
+options, err := ably.NewClientOptions("xxx:xxx")
+if err != nil {
+	panic(err)
+}
+
+clien := ably.NewRestClient(options)
```
kouno: Spelling mistake :).

ably/options_test.go
```diff
@@ -1,58 +1,33 @@
...
-		})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(opts.KeyName).To(Equal("name"))
+		Expect(opts.KeySecret).To(Equal("secret"))
+	})
 
-		AfterEach(func() {
-			ably.Log.Logger = old
-		})
+	It("when Key is invalid it returns an error", func() {
```
kouno: This still reads as a `Context` not a `It`. A context sets some configuration (in this case the fact that we pass an invalid key), and then you write the expectation in the It block.

ably/rest_client.go
```diff
@@ -27,36 +29,30 @@ func query(fn func(string, interface{}) (*http.Response, error)) QueryFunc {
...
-	RestEndpoint string
-	Protocol     string
+	Options *ClientOptions
+	Auth    *Auth
 
-	HTTPClient *http.Client
-
-	channels map[string]*RestChannel
-	chanMtx  sync.Mutex
+	mtx   sync.Mutex // protects chans
```
kouno: That looks weird. Renaming a variable which obfuscates its purpose and forcing us to do a look up to see that it `// protects chans`?

ably/rest_client.go
```diff
@@ -92,27 +86,24 @@ func (c *RestClient) Stats(params *PaginateParams) (*PaginatedResource, error) {
...
-	req, err := http.NewRequest("GET", c.RestEndpoint+path, nil)
+	req, err := http.NewRequest("GET", c.options().restURL()+path, nil)
 	if err != nil {
 		return nil, err
 	}
 	req.Header.Set("Accept", "application/json")
-	req.SetBasicAuth(c.Auth.Options.Token, c.Auth.Options.Secret)
-	res, err := c.httpclient().Do(req)
-
+	// TODO(rjeczalik): add support for token auth
```
kouno: Remove this and add an issue in the repo :wink:.

ably/rest_client_test.go
```diff
@@ -21,38 +21,39 @@ import (
...
 		}
 	)
 
-	Context("with a failing request", func() {
+	Context("with a failing req", func() {
 		var (
-			statusCode int
-			request    *http.Request
+			req    *http.Request
+			cli    *ably.RestClient
```
kouno: I don't really understand why you want to truncate all variable names like that. `server`, `client` and `request` were acceptable variable names.

ably/rest_client_test.go
```diff
@@ -21,38 +21,39 @@ import (
...
-						return url.Parse(server.URL)
-					},
+					Proxy: func(req *http.Request) (*url.URL, error) { return url.Parse(srv.URL) },
 				},
 			}
 		}
 	)
 
-	Context("with a failing request", func() {
+	Context("with a failing req", func() {
```
kouno: This is supposed to be read as normal english :smile:. You are not supposed to copy variable names in there.

ably/rest_http_error.go
```diff
@@ -15,6 +15,7 @@ type RestHttpError struct {
 func NewRestHttpError(response *http.Response, message string) *RestHttpError {
 	err := &RestHttpError{Response: response, Msg: message}
 	err.readBody()
+	err.Response = nil
```
kouno: If you unset this, maybe `Response` should become private?

ably/testutil/test_app.go
```diff
@@ -64,24 +80,32 @@ func (t *App) populate(res *http.Response) error {
...
 
-	t.Options.Key = t.AppKeyId() + ":" + t.AppKeyValue()
+	t.opts.KeyName = t.KeyName()
+	t.opts.KeySecret = t.KeySecret()
 
 	return nil
 }
 
+func (t *App) httpclient() *http.Client {
+	if t.Options != nil && t.opts.HTTPClient != nil {
```
kouno: `t.Options`? Shouldn't that be `t.opts != nil`?

ably/export_test.go
```diff
@@ -1,5 +1,13 @@
...
 	return p.buildPath(base, rel)
 }
+
+func (opts *ClientOptions) RestURL() string {
+	return opts.restURL()
+}
+
+func (opts *ClientOptions) RealtimeURL() string {
+	return opts.RealtimeURL()
+}
```
kouno: On a side note, this file has been puzzling me for a bit. Isn't this going to be available outside the package anyway?

ably/rest_client_test.go
rjeczalik: http://doc.cat-v.org/bell_labs/pikestyle "Variable names"

- `req` is common way of naming `http.Request` variables (https://golang.org/pkg/net/http/)
- `statusCode` the Code part is redundant as the type says its a code (int) and not a status text (string); acutally the variable is removed, but if it'd stayed it'd be names just `status`
- `cli` I agree is artificial, however it conflicts with global `client`, so for readability I named it differently (imo better than client2); suggestions welcome on this one :)

ably/rest_http_error.go
rjeczalik: It should, I need to rewrite RestHttpError either way as it's a bit useless in current form. Can I do this in separate PR?

ably/export_test.go
rjeczalik: This is defined only for tests - if tests are external (the ones with _test suffix) and for some reason need to access unexported symbols the pattern is to define them exported for tests, usually in `export_test.go` file. As I removed a number of exported functions from the API I needed to export them for tests. I aim to not use unexported functions in rewritten tests when possible.

ably/options_test.go
rjeczalik: > A context sets some configuration (in this case the fact that we pass an invalid key)

This is not configuration, this is actual test - setting an invalid key. But I'll wrap it into Context for the Ginkgo sake ;)

ably/rest_client.go
rjeczalik: Pretty common, you aim for short variables. In stdlib even structs that have multiple mutex use short mutex names:

```
$ grep -irnH 'mu sync' . | grep protects
./net/http/triv.go:33:	mu sync.Mutex // protects n
./net/singleflight.go:28:	mu sync.Mutex       // protects m
./syscall/srpc_nacl.go:23:	outMu sync.Mutex // protects writing to connection
```

or https://golang.org/src/net/rpc/server.go?s=5638:5874#L174. But I'd rename it as I have no strong opinion about it.

ably/auth_test.go
rjeczalik: > The only way for a property to be immutable outside the package is to use accessors.

This is not necessarily true, the same problem is in Java - even though you return a variable with accessor without defensive copying it's still can be modified. So we gain nothing with using accessors. However I very like using accessor vs exported fields (exported fields should be used only for model, not control imho, sorry for offtop), as then I can make them being constructed lazily, which allows for using zero-value `RestClient`, which is preferable in Go.

Comming back to `Presence`, the `RestPresence` does not have any fields exported, so the user can't modify it either way :)

ably/rest_client.go
rjeczalik: This was actually the first thing I've noticed - I've collected a fair amount things that need either to be fixed, missing or polished a bit, I keep them in my trello board as I didn't want to create noise in github issues - https://trello.com/b/pHESio8l/ably-go. Or you feel you want to give feedback on those and I should move them to github isseus?

ably/rest_client.go
rjeczalik: Done, renamed to chansMtx.

ably/options_test.go
rjeczalik: Done, splitted into Context/BeforeEach and It scopes.

README.md
rjeczalik: Done.

ably/rest_client.go
rjeczalik: Done, I've removed those TODOs and the one from auth.go about :query_time.

ably/rest_client_test.go
rjeczalik: Done, this was sloppy sed fail.

ably/rest_http_error.go
rjeczalik: Done, changed to unexported field.

ably/testutil/test_app.go
rjeczalik: Done, thanks.

rjeczalik:
@kouno Another round of fixes, please take a look :)

ably/options.go
```diff
@@ -14,36 +15,86 @@ const (
...
+	if host == "" {
+		host = DefaultOptions.RealtimeEndpoint
+	}
+	if opts.NoTLS {
+		return "ws://" + host + ":80"
+	}
+	return "wss://" + host + ":443"
+}
+
+func (opts *ClientOptions) SetKey(key string) error {
```
mattheworiordan: Sorry, I should have picked up on this earlier, but I'm a little concerned that we're making options mutable.  In Ruby at least, the options are immutable once the client library has been instanced.  This is necessary as if you want a new client with different credentials for example, you'd need to create a new client library.  If I read this correctly, we are making this not only immutable, but in fact exposing public methods making it apparent it is OK to change the options?  Do we mean to do this?

README.md
```diff
@@ -31,8 +31,13 @@ TODO
 All examples assume a client and/or channel has been created as follows:
 
 ```go
-client := ably.NewRestClient(ably.ClientOptions{ApiKey: "xxxx"})
-channel := client.Channel('test')
+options, err := ably.NewClientOptions("xxx:xxx")
```
mattheworiordan: Out of interest, if ClientOptions is effectively just a data type, do you think this style is necessary?  Could we not simply do the err checking in the Client instead and keep ClientOptions as a simple immutable set of options used to instance a Client?

ably/options_test.go
```diff
@@ -1,57 +1,44 @@
...
-		options = &ably.ClientOptions{
-			Key: "id:secret",
-		}
+	Context("when Key is valid", func() {
+		var err error
 
-		options.Prepare()
-	})
+		BeforeEach(func() {
+			err = opts.SetKey("name:secret")
```
mattheworiordan: As mentioned above, this is not an expected way to use the options, perhaps this is more idiomatic Go and we can offer this, but I would have expected that one only ever creates the options in one statement such as `ably.ClientOptions { option: "val" }`

ably/rest_client.go
```diff
@@ -27,36 +29,30 @@ func query(fn func(string, interface{}) (*http.Response, error)) QueryFunc {
...
-	Protocol     string
+	Options *ClientOptions
+	Auth    *Auth
 
-	HTTPClient *http.Client
-
-	channels map[string]*RestChannel
-	chanMtx  sync.Mutex
+	chansMtx sync.Mutex
+	chans    map[string]*RestChannel
```
mattheworiordan: Why change of channels to chans which is less descriptive?

ably/rest_client.go
```diff
@@ -66,21 +62,19 @@ func (c *RestClient) Time() (time.Time, error) {
 		return time.Time{}, err
 	}
 	if len(times) != 1 {
-		return time.Time{}, fmt.Errorf("Expected 1 timestamp, got %d", len(times))
+		return time.Time{}, fmt.Errorf("expected 1 timestamp, got %d", len(times))
```
mattheworiordan: Why lower case "expected" whereas below you have "Unexpected" i.e. sentence case.  I'm not clear on why this commit was even made given it creates inconsistencies in case now?

ably/rest_client_test.go
mattheworiordan: I am with @kouno on this, I don't see the benefit of `srv` of `server` or `req` over `request`.  Whilst perhaps this shorter notation is used in other packages, I would prefer to have our variables descriptive.  `status` is fine, however please note that in our errors we have `status` and `code` where code is our internal error code.

mattheworiordan: Why does `client` conflict with a global `client`?  What is the global `client`?

ably/rest_client_test.go
```diff
@@ -85,58 +87,52 @@ var _ = Describe("RestClient", func() {
...
 
-				err := client.Channel("test").Publish("ping", "pong")
+				err := cli.Channel("test").Publish("ping", "pong")
 				Expect(err).NotTo(HaveOccurred())
 			})
 
 			It("encode the body of the message using msgpack", func() {
 				var anyMsgPack []map[string]interface{}
-				err := msgpack.Unmarshal(buffer, &anyMsgPack)
+				err := msgpack.Unmarshal(buf, &anyMsgPack)
```
mattheworiordan: As above, `buf` and `cli` instead of `buffer` and `client` IMHO does not make this code more readable.

ably/rest_http_error.go
```diff
@@ -6,15 +6,16 @@ import (
 )
 
 type RestHttpError struct {
-	Msg      string
-	Response *http.Response
-
+	Msg          string
 	ResponseBody string
+
+	resp *http.Response
```
mattheworiordan: It feels odd to have an HTTP Error without the status code as part of the type?  Also, note that our responses always follow a generic format in JSON, see https://rest.ably.io/ as an example.  We therefore have an Ably ErrorInfo type, see https://github.com/ably/ably-java/blob/master/src/io/ably/types/ErrorInfo.java.  Should we not follow that format?

ably/rest_client_test.go
rjeczalik: Global client is a test one which is a global shared resource for the ginkgo spec. This is extremely dangerous, as part of the tests modify (!) the global instance, so while it works now, it is really fragile. For now I need to live with this, however I'm planning to drop this after all tests are rewritten to testing.

I'll renamed those variables to their original long form a bit later.

ably/options.go
rjeczalik: The problem the setter is solving is dependant fields: setting key field should set keyName and keySecret to proper values, However if user sets key to different value (e.g. after instantiating client but before making first request or changing the key in some handler code when the old key was wrong) the keyName and keySecret fields have still the old value, which leads to sloppy API. Thus I solved this with setter. If we have better options or we don't care the please give me feedback. 

From the description:

```
/* old API */
client := ably.NewRestClient(&ably.Options{Key: "xxx:xxx"})
// ...
client.Options.Key = "yyy:yyy"
// all calls to REST API are going to fail, as Token and Secret fields
// are still "xxx"

/* new API */
client, err := ably.NewRestClient("xxx:xxx")
// ...
err := client.Options.SetKey("yyy:yyy")
// correctly handled here
```

README.md
rjeczalik: Yes, I wanted to do that in the original version, but in order to make that possible, I needed to have meaningful zero-value for ably.RestClient, so the factory method could take the key string intead of options (Go has no method overloading). The other options is to have to factory methods, one NewRestClient(key string) and second NewRestClientOptions(*CientOptions). If you like the latter idea, I can make it.

The factory method for ClientOptions returns error because it's parsing argument, and parsing can fail.

ably/options_test.go
rjeczalik: Dependant fields problem, if you can tell me how to solve that differently or that we don't care, then I'll change that.

ably/rest_client.go
rjeczalik: Already fixed.

ably/rest_client.go
rjeczalik: I'll change all text messages to have lower-case names a bit latter. This is how idiomatic Go error messages are.

ably/rest_client_test.go
rjeczalik: Sure, will change to longer versions.

ably/rest_http_error.go
rjeczalik: As I already said, I'm going to rewrite RestHttpError in a separate PR, as it's useless in current form. Is that ok?

ably/rest_client_test.go
kouno: @rjeczalik you have a valid point about the `client`. This was easy to manage at first because there was not much tests and the `client` was never really changed, but with the tests around client options, this will inevitably clash.

ably/rest_client.go
kouno: I don't really like Trello in general... But ignoring that, I think that since this is an open source client library, the best is to place these things in our issues and add a label for `features` for transparency.

rjeczalik: Sure, I'll move them later. I was intended only for my personal use, to I keep track of things.

ably/rest_http_error.go
rjeczalik: > Also, note that our responses always follow a generic format in JSON, see (...)

Yes, I aim to have single `ably.Error` type used for all client errors with proper constants and fields to denote Code, internal Code etc. I'm just not describing the design here in order to not draw attention away from this PR :)

ably/rest_client_test.go
mattheworiordan: Ok, thanks for clarifying, well I would change the globalClient if that is the problem then as that part feels like the issue.

ably/options.go
mattheworiordan: I think the problem here is that we're allowing someone to change the key afterwards at all.  Why are we allowing this?  It is unnecessary and introduces these problems.
Also, like you changed with the "realtimeEndPoint", `keyName` and `keySecret` (not Token & Secret) can be determined at runtime from the full Key.

README.md
mattheworiordan: I think the shorthand initialisation with just a key string is causing too many problems and if overloading is not possible, I would simply go for the following instead:

```
ably.NewRestClient(ably.ClientOptions{ Key: "xxxx" })
```

It reduces complexity significantly.

ably/options_test.go
mattheworiordan: Why not:
```
BeforeEach(func() {
   client, err = ably.NewRestClient(ably.ClientOptions{Key:"name:secret"})
}
````

File: ably/options.go (hunk 16 above)
Comment (kouno) [inline ably/options.go]: > I think the problem here is that we're allowing someone to change the key afterwards at all. Why are we allowing this? It is unnecessary and introduces these problems.

Agreed. Once instantiated, the options shouldn't be modifiable. You have to set the options right before instantiating.

Comment (rjeczalik) [inline ably/options.go]: But then *ClientOptions should not be exported / accessible from `RestClient`, and right now it is; I'd love to unexport it, since once ClientOptions were passed from outside to client, they should also be accessed from outside and not from the `RestClient` - thanks to this change user would have no opportunity of altering this field; however I've seen Java library makes this field public. So the question is - are those ClientOptions needed to be accessbile from RestClient? If not, the it solves all the problems.

Comment (mattheworiordan) [inline ably/options.go]: Yes, perhaps Java has left these options as mutable, but this is not our intended design. @paddybyers do you think we should modify the Java library to make `ClientOptions Rest.options` immutable?  I don't see any valid use case where the options should be changed after the library is created.

Comment (kouno) [inline ably/options.go]: I think we have a misunderstanding here.

This struct should be available to developers, but once you pass the ClientOptions instance in the client, you should not expect that any later change to the instance will be reflected on the Client. For example the following code should not work:

```go
options := ably.NewClientOptions()
options.Key = "something"

client, err = ably.NewRestClient(options)

options.Key = "somethingelse"

// client should still use the previous key.
```

mattheworiordan: Indeed, although I am not sure why we even need to provide the write accessor `options.Key = "something"` when one can simply use `ClientOptions { key: "something" }`

paddybyers: > I don't see any valid use case where the options should be changed after the library is created

No, we should not attempt to support modification of the options and if the language provides a convenient way to ensure they are immutable after having been passed to the `Ably` library then we should do that.

However making them `final` in, say, the java lib then gives you an API that is much more awkward to use because they would then all need to be specified in the `ClientOptions` constructor and I don't think we should do that.

kouno: True. I was just using the current API to make myself clear ;).

rjeczalik: But still RestClient uses its own Options field to e.g. read RestEndpoint. If user changes RestClient.Options.RestEndpoint then there's no way to prevent that other than documenting it that it must be accessed only for reading:

```go
options := ably.NewClientOptions()
options.Key = "something"

client, err := ably.NewRestClient(options)

client.Options.RestEndpoint = "invalid" // this is going to break all the requests made by this client
```

README.md
rjeczalik: Do you agree for removing KeyName and KeySecret fields from ClientOptions and have getters instead, which read Key in the runtime?

```go
ops := ably.ClientOptions{Key: "xxx:xxx"}
keyName := opts.KeyName()
keySecret := opts.KeySecret()
```

ably/options_test.go
rjeczalik: This test-case tests options, not RestClient, I'm not convinced we should mixed responsibilities in the tests. However I'm going to revisit this tests after final modifications to this PR are applied according to other discussions.

ably/options.go
kouno: @rjeczalik Well you are not supposed to keep the `Options`. At the moment the library does the right thing for RestEndpoint.

https://github.com/ably/ably-go/blob/master/ably/rest_client.go#L44-L45

Since a string is immutable, modifying `options.RestEndpoint` in my example wouldn't break. It is another story for the http client though...

mattheworiordan: Well we should ideally prevent the options from being editable, so arguably we could make Options not public and expose the options a user may need to access as accessors of the RestClient such as `client.Key()`, but I think you'll need to advise what is possible with Go.

README.md
mattheworiordan: Yup

ably/options.go
rjeczalik: > Well you are not supposed to keep the Options. At the moment the library does the right thing for RestEndpoint.

Not really, there's no difference, user can change both, either `client.ClientOptions.RestEndpoint` or `client.RestEndpoint`.

mattheworiordan: Ok, well I think we should ideally make Options not a public struct type, and instead provide accessors once the library is instanced so that unless someone intentionally pokes into the library and tries to break things, they will not mistakenly be led to believe they can change attributes.

rjeczalik: Done, Options is not kept within RestClient, Endpoint and Protocol fields are back, they're read-only. Will document that fact with in-line code documentation in separate PR.

README.md
rjeczalik: Done, Key field is back with KeyName() and KeySecret().

ably/options_test.go
rjeczalik: Done, tests for ClientOptions use RestClient.

ably/rest_client.go
rjeczalik: Done, error messages start with lower case.

ably/rest_client_test.go
rjeczalik: Done, long names.

rjeczalik:
@mattheworiordan @kouno: Please take a look, pretty hot PR, I've addressed all naming, options-related issues.

rjeczalik:
@mattheworiordan @kouno: Please take a look, pretty hot PR, I've addressed all naming, options-related issues.

kouno:
Looks ok to me. @mattheworiordan any other comments?

ably/options.go
```diff
@@ -14,36 +15,74 @@ const (
 	ProtocolMsgPack = "msgpack"
 )
 
+var errInvalidKey = errors.New("invalid key format")
+
+var DefaultOptions = &ClientOptions{
+	RestEndpoint:     "rest.ably.io",
+	RealtimeEndpoint: "realtime.ably.io",
```
mattheworiordan: Sorry, should have noticed before, but the option name could be considered a bit misleading as an end point is typically a FQDN i.e. "http://rest.ably.io" whereas what we actually allow is simply a custom host.  See https://github.com/ably/ably-java/blob/master/src/io/ably/types/ClientOptions.java#L61-L67 for the Java implementation.  Personally I prefer Rest & Realtime over the Java implementation, but can we use the name host instead please as this is more accurate and consistent, and we'll clean up the Java version too to match this?

ably/options.go
```diff
@@ -14,36 +15,74 @@ const (
...
 	NoTLS            bool
 
 	HTTPClient *http.Client
 }
 
-func (p *ClientOptions) Prepare() {
-	if p.Key != "" {
-		p.parseKey()
+func (opts *ClientOptions) restURL() string {
+	host := opts.RestEndpoint
```
mattheworiordan: as above, this confirms that `RestEndPoint` is indeed just `RestHost`

ably/paginated_resource_test.go
```diff
@@ -16,7 +16,7 @@ var _ = Describe("PaginatedResource", func() {
 		})
 
 		It("returns a string pointing to the new path based on the given path", func() {
-			newPath, err := ably.BuildPath(&paginatedResource, "/path/to/resource?hello", "./newresource?world")
+			newPath, err := paginatedResource.BuildPath("/path/to/resource?hello", "./newresource?world")
```
mattheworiordan: This is my fault because in Ruby I named the class `PaginatedResource`, but in Ably-js and Ably-Ruby it's called `PaginatedResult`.  Sorry, entirely my fault I expect.  See https://github.com/ably/ably-ruby/issues/40

ably/options.go
rjeczalik: `RestHost` and `RealtimeHost`, correct?

mattheworiordan:
@rjeczalik looks good, only one or two small comments, thanks :+1:

ably/options.go
mattheworiordan: Yup, see https://github.com/ably/ably-java/issues/22

rjeczalik:
@mattheworiordan @kouno Please take a look.

PR #32, opened 2015-05-13
realtime client - first working version

rjeczalik:
The realtime client implements:

- event emitters for connection channels
- channel: publish, subscribe, unsubscribe with ack/nack confirmation
- message queueing for connection, channels

@mattheworiordan @kouno Please take a look.


rjeczalik:
@mattheworiordan @kouno I've squashed the history, now it should be easier. Please take a look. I'm still be sending a change for one test and implementation for Close, Ping so don't spend time on looking for those ;) Thanks!

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+	}
+	msg := &proto.ProtocolMessage{Action: proto.ActionAttach, Channel: c.state.channel}
+	if err := c.client.Connection.send(msg, nil); err != nil {
+		return c.state.set(StateChanFailed, newErrorf(90000, "ably: Attach error: %s", err))
+	}
+	return nil
+}
 
-	Err error
+var errDetach = newError(90000, errors.New("ably: Detach on invactive connection"))
```
mattheworiordan: Minor comment, typo

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+	if !c.isActive() {
+		return nil
+	}
+	c.state.set(StateChanDetaching, nil)
+	if !c.client.Connection.isActive() {
+		return c.state.set(StateChanFailed, errDetach)
+	}
+	msg := &proto.ProtocolMessage{Action: proto.ActionDetach, Channel: c.state.channel}
+	if err := c.client.Connection.send(msg, nil); err != nil {
+		c.state.set(StateChanFailed, newErrorf(90000, "ably: Detch error: %s", err))
```
mattheworiordan: Minor comment, typo

mattheworiordan:
It's a monster PR @rjeczalik!  I have skimmed through it all, but unfortunately not got time to look through in detail.  Added to this, I am not that familiar with Go, so don't want to waste too much time asking stupid questions ;)

If we can ensure we have suitable test coverage to match the spec, then I am confident the specific  implementation details are less important.

Thanks @rjeczalik, good work, :+1:

ably/realtime_conn_test.go
```diff
@@ -0,0 +1,74 @@
...
+	return ch, func() { wg.Wait() }
+}
+
+func TestRealtimeConn_Connect(t *testing.T) {
+	client, err := ably.NewRealtimeClient(testutil.Options(t, nil))
+	if err != nil {
+		t.Fatalf("ably.NewRealtimeClient=%v", err)
+	}
+	defer client.Close()
+	record(t, nil, func() bool { return client.Connection.State() == ably.StateConnConnected })
```
kouno: It would be safer to `wait()` for record to end, no?

ably/realtime_conn_test.go
```diff
@@ -0,0 +1,74 @@
...
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case <-stop:
+				t.Fatalf("record timed out after %v (spy=%v)", timeout, spy)
+			case s := <-ch:
+				spy[s.State]++
```
kouno: I know it's quite a simple test so I would be surprised if it happens here, but shouldn't this be an array instead of a `map` where you would have the history of what happened instead of just how many time the connection was in this state?

ably/realtime_conn_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"reflect"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/ably/ably-go/ably"
+	"github.com/ably/ably-go/ably/testutil"
+)
+
+func record(t *testing.T, spy map[int]int, done func() bool) (chan<- ably.State, func()) {
```
kouno: I would have probably called it `recordStateUntil` and renamed `spy` to be `statesCount`.

ably/realtime_conn.go
```diff
@@ -0,0 +1,367 @@
...
+	c.msgSerial = (c.msgSerial + 1) % maxint64
+	if listen != nil {
+		c.pending.Enqueue(msg.MsgSerial, listen)
+	}
+	c.state.Unlock()
+	return c.conn.Send(msg)
+}
+
+func (c *Conn) isActive() bool {
+	c.state.Lock()
```
kouno: It looks like you moved all the locks to be `Mutex` instead of `RWMutex`. I'm not sure why you did that as it is was interesting to see what function was actually trying to write something or just read the value.

Also it avoids having blocking go routines for simple reads.

ably/error.go
```diff
@@ -0,0 +1,263 @@
...
+	return &Error{
+		Code:   err.Code,
+		Status: err.StatusCode,
+		Err:    errors.New(err.Message),
+	}
+}
+
+func checkError(resp *http.Response) error {
+	const maxContentLength = 32 * 1024 // 32 KiB
+	if resp.StatusCode/100 < 3 {
```
kouno: What was wrong with `resp.StatusCode < 300`? :wink:

ably/error.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil
+	}
+	return &Error{
+		Code:   err.Code,
+		Status: err.StatusCode,
+		Err:    errors.New(err.Message),
+	}
+}
+
+func checkError(resp *http.Response) error {
```
kouno: The name of the function is really not that explicit. `checkValidHTTPResponse` or similar would help.

ably/error.go
```diff
@@ -0,0 +1,263 @@
...
+	const maxContentLength = 32 * 1024 // 32 KiB
+	if resp.StatusCode/100 < 3 {
+		return nil
+	}
+	n := resp.ContentLength
+	if n <= 0 || n >= maxContentLength {
+		n = maxContentLength
+	}
+	body := &proto.Error{}
+	if e := json.NewDecoder(io.LimitReader(resp.Body, n)).Decode(body); e != nil {
```
kouno: I feel like we are asking for trouble here. Why not just do a `ioutil.ReadAll(resp.Body)` here, instead of expecting the server to return the right content length?

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+// the returned list would not list a channel that was already released from
+// different goroutine.
+//
+// The returned list is sorted by channel names.
+func (ch *Channels) All() []*RealtimeChannel {
+	ch.mtx.Lock()
+	chans := make([]*RealtimeChannel, 0, len(ch.chans))
+	for _, c := range ch.chans {
+		chans = append(chans, c)
+	}
```
kouno: It sounds like this is doing the same thing as `copy` http://golang.org/pkg/builtin/#copy.

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
 
-func (c *RealtimeChannel) Publish(name string, data string) error {
-	c.Attach()
+// On relays request channel states to c; on state transition
+// connection will not block sending to c - the caller must ensure the incoming
+// values are read at proper pace or the c is sufficiently buffered.
+//
+// If no states are given, c is registered for all of them.
+// If c is nil, the method panics.
+// If c is alreadt registered, its state set is expanded.
```
kouno: `already`

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+func (c *RealtimeChannel) On(ch chan<- State, states ...int) {
+	c.state.on(ch, states...)
+}
+
+// Off removes c from listetning on the given channel state transitions.
```
kouno: `listening`

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+		c.queue.Fail(newErrorProto(msg.Error))
+		// TODO recovery
 	case proto.ActionMessage:
-		c.listenMtx.RLock()
-		defer c.listenMtx.RUnlock()
-		for _, m := range msg.Messages {
-			if l, ok := c.listeners[""]; ok {
+		c.listenMtx.Lock()
+		for _, msg := range msg.Messages {
+			if l, ok := c.listen[""]; ok {
```
kouno: Why did you rename `listeners` to `listen`? This is less readable :/.

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
-		return
-	}
-	c.State = ChanStateAttaching
-}
-
-func (c *RealtimeChannel) setState(s ChanState) {
-	c.stateMtx.Lock()
-	defer c.stateMtx.Unlock()
-	c.State = s
+func (c *RealtimeChannel) isActive() bool {
```
kouno: Reading the code, I understand why `isActive` doesn't use a lock here since you are managing the lock outside of this method. But surely, you should be consistent with the other `isActive` on the connection too and not use locks.

ably/realtime_channel_test.go
```diff
@@ -1,33 +1,118 @@
...
+		if err != nil {
+			return errors.New("nack: " + err.Error())
+		}
+	case <-time.After(timeout):
+		return fmt.Errorf("waiting for Publish confirmation timed out after %v", timeout)
+	}
+	return nil
+}
+
+func publish(fn func(string, string, chan<- error) error, name, data string) func(chan<- error) error {
```
kouno: I understand you want to dry your tests, but this is starting to get crazy :/.

ably/ably.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+func max(i, j int) int {
+	if i > j {
+		return i
+	}
+	return j
+}
+
+func nonil(err ...error) error {
```
lmars: This function seems to be unused

ably/error.go
```diff
@@ -0,0 +1,263 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+
+	"github.com/ably/ably-go/ably/proto"
+)
+
+func tostatus(code int) int {
```
lmars: I think `toHTTPStatus` is more descriptive

paddybyers: We use `statusCode` to represent the HTTP status code, so wouldn't it be `toStatusCode` ?

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+	return nil
+}
 
+// RealtimeChannel represents a single named message channel.
+//
+// The name of the channel is read-only - assigning new name to it does not
+// have any effect.
 type RealtimeChannel struct {
-	Name string
+	Name string // read-only name of the channel
```
lmars: Rather than this comment, how about not exporting it and adding a `Name()` method?

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+
+var errAttach = newError(90000, errors.New("ably: Attach on inactive connection"))
 
-	State    ChanState
-	stateMtx sync.Mutex
+// Attach initiates attach request, which is being processed on a separate
+// goroutine.
+//
+// In order to receive its result register a channel with the On method; upon
+// successful attachment the channel will receive StatChanAttached event.
```
lmars: I think calling this method explicitly should only return when the attach request has been made, and the resulting error indicating wether it is attached or not

```go
ch := channels.Get("foo")
if err := ch.Attach(); err != nil {
  // handle attach failed
}
```

WDYT?

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+// are being dispatched.
+//
+// If no names are given, returned channel will receive all messages.
+// If ch is nil, Subscribes creates new, unbuffered channel.
+//
+// The caller must ensure the messages are read from returned channel at
+// sufficient pace, otherwise they may be dropped. In order to use buffered
+// channel instead of the default unbuffered one, make new channel with requested
+// capacity and pass it as ch; the Subscribe method will return it instead of
+// creating new one.
```
lmars: Rather than being concerned with buffering and slow readers, why not just have a goroutine per subscribed channel?

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
-			{Name: name, Data: data},
-		},
+		Action:   proto.ActionMessage,
+		Channel:  c.Name,
+		Messages: messages,
 	}
-	return c.client.send(msg)
+	switch c.State() {
+	case StateChanInitialized, StateChanAttaching:
+		if c.client.opts.NoQueueing {
```
lmars: It seems a bit of a race to call `Attach` at the top (asynchronously), and then here potentially reject publishing a message if that request is in transit, won't that mean clients who set `NoQueueing` will always have their first message rejected unless they explicitly wait for attach? Is that the expected behaviour?

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+	default:
+		return &Error{Code: 90001}
+	}
+	return c.client.Connection.send(msg, listen)
+}
+
+// State gives current state of the channel.
+func (c *RealtimeChannel) State() int {
+	c.state.Lock()
+	state := c.state.current
```
lmars: You can return this value directly by deferring the unlock

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
 	case proto.ActionAttached:
-		c.setState(ChanStateAttached)
+		c.state.syncSet(StateChanAttached, nil)
+		c.queue.Flush()
 	case proto.ActionDetached:
-		c.setState(ChanStateDetached)
+		c.state.syncSet(StateChanDetached, nil)
 	case proto.ActionPresence:
-		// TODO
+		Log.Printf(LogError, "TODO: implement RealtimePresence: %v", msg)
```
lmars: I think TODOs belong in comments, not in a user's log file

ably/realtime_channel.go
```diff
@@ -2,130 +2,342 @@ package ably
...
+
+// Publish publishes a message on the channel, which is send on separate
+// goroutine. Publish does not block.
+//
+// If listen channel is non-nil it will be used to notify whether the message
+// was published succesfully with nil error value; when the operation failed,
+// the error that caused the failure is going to be sent on the listen channel.
+//
+// This implicitly attaches the channel if it's not already attached.
+func (c *RealtimeChannel) Publish(name string, data string, listen chan<- error) error {
```
lmars: The third argument here seems quite awkward if I want to publish a message and wait for it:

```go
errCh := make(chan error)
if err := c.Publish("foo", "bar", errCh); err != nil {
  return err
}
if err := <-errCh; err != nil {
  return err
}
```

I think it would be clearer with something like:

```go
res, err := c.Publish("foo", "bar")
if err != nil {
  return err
}
res.Wait()
if res.Failed {
  return res.Err
}

File: ably/realtime_channel.go [hunk 25]
```diff
@@ -2,130 +2,342 @@ package ably
...
+			if l, ok := c.listen[msg.Name]; ok {
 				for ch := range l {
-					ch <- m
+					select {
+					case ch <- msg:
+					default:
+						Log.Printf(LogWarn, "dropped %q message due to slow receiver", msg.Name)
+					}
 				}
 			}
```
lmars: Rather than storing channels, I would store a wrapper type which is responsible for sending the messages to the subscriber, e.g.:

```go
type Subscription struct {
	ch    chan *proto.Message
	queue Queue
}

func newSubscription(ch chan *proto.Message) *Subscription {
	s := &Subscription{
		ch:    ch,
		queue: newQueue(),
	}
	go s.loop()
	return s
}

func (s *Subscription) notify(msg *proto.Message) {
	s.queue.Enqueue(msg)
}

func (s *Subscription) loop() {
	for {
		s.ch <- s.queue.Pop()
	}
}
```

then you don't have to worry about slow receivers, and also allows the user to just call something like `subscription.Close()` to unsubscribe / close the channel.

ably/realtime_conn.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+	return []string{"false"}
+}
+
+// Connect is used to connect to Ably servers manually, when the client owning
+// the connection was created with NoConnect option. The connect request is
+// being processed on a separate goroutine.
+//
+// In order to receive its result register a channel with the On method; upon
+// succsufful connection channel will receive StatConnConnected event.
```
lmars: What is the use case where I would explicitly call `Connect()` but not want to know the result straight away? It seems to me like this should return once the connection attempt has been made.

ably/realtime_conn.go
```diff
@@ -0,0 +1,367 @@
...
+
+func (c *Conn) msgloop() {
+	for {
+		me := &msgerr{}
+		me.err = c.conn.Receive(&me.msg)
+		c.connCh <- me
+		if me.err != nil {
+			return // TODO recovery
+		}
+	}
```
lmars: Why two loops? Can this not just be in `eventLoop`?

ably/state.go
```diff
@@ -0,0 +1,374 @@
...
+	"sync"
+
+	"github.com/ably/ably-go/ably/proto"
+)
+
+// StateType specifies which group of states is relevant in given context;
+// either:
+//
+//   - StateConn* group describing Conn states
+//   - StateChan* group describing RealtimeChannel states
```
lmars: Shouldn't these be two separate types? e.g.:

```go
type StateChan string

const (
  StateChanInitialized StateChan = "ably.StateChanInitialized"
  ...
)

File: ably/error.go (hunk 9 above)
Comment (rjeczalik) [inline ably/error.go]: Done.

I’ve rewritten to ioutil.ReadAll to simplify implementation so readability does not suffer.

Just for the record, this does not ask for the length, it uses length if present to preallocate full buffer at-once, falling back to what ioutil.ReadAll does when the ContentLength is unknown.

File: ably/realtime_channel.go (hunk 10 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Can’t touch that, c.chans is map ;)

File: ably/realtime_channel.go (hunk 18 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Changed comment. A word of recap on function vs public field discussions we had so far - the same issue I had with so-called properties like RestClient.Auth, RealtimeClient.Channels, RealtimeClient.Connection etc., even though they can be overwritten we value consistency more with other libraries and that’s why we keep all of them as public fields. Also this is a reason why Name field is a field. I’ll add a statement to README.md that API is not stable and is a subject to change, so we could revisit whole API at the end. Sounds ok?

File: ably/realtime_channel.go (hunk 25 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Async queueing implementation is much more complicated that what you’ve described. See [this code](https://github.com/rjeczalik/ably-go/commit/651694f1e2dc8d20f4d7bce46755d6a238a1f265#diff-a5254638ab4b7f5f9e7482b7a5feed87R173). This is no longer just a wrapper. I don’t want to get rid of slow receivers, I want slow receivers not to leak goroutines started by us. I consider slow receivers as programming error (insufficient buffering, too much blocking operations) on the user side which I would not want to handle.

File: ably/realtime_channel.go (hunk 19 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

Upon Attach return, the attach message is sent and the state is either Attaching or Failed. Moreover to make it more user-friendly, it returns a Result value (in Java library it’s called CompletionListener) which can be used to wait until channel becomes attached or when it transits to a failed / closed state.

(I had no better idea for better name than Result, if you see any please suggest and I’ll change)

File: ably/realtime_conn.go (hunk 26 above)
Comment (rjeczalik) [inline ably/realtime_conn.go]: Done.

Connect waits until connect message is sent and then returns with result. Like above, it also returns Result value (counterpart for Java’s CompletionListener) for full sync.

File: ably/state.go (hunk 28 above)
Comment (rjeczalik) [inline ably/state.go]: You will need to have then two separate State’s, ChanState and ConnState and in order to do that you’ll need to duplicate code for stateEmitter to handle two distinct types. Having this as just int not only allows you to write the code only once, it enables the user to blanket handle all state transitions with single function (see ClientOptions.Listener).

File: ably/realtime_channel.go (hunk 20 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: I’m reluctant to do this - I know that goroutines are cheap etc. but that’s the point - slow receivers indicate user error - which is typically solved by using bigger buffer for user channels or ensuring message handling is not blocked. But that’s not something we’ll fix by starting goroutines, that is yes, messages won’t be dropped, but they won’t be handled faster either.

Exploding goroutines and in worst case leaking them does not sound good to me. It’s worth to mention, that the same facility library exposes to the users for waiting on events or receiving messages is also used by the library internally - it’ll be easier to troubleshoot eventual bugs if sends are dropped right away rather than the library will have hanging goroutines.

File: ably/realtime_channel.go (hunk 21 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Fixed.

I misunderstood how ClientOptions.Queueing is meant to work - this param is used to enable/disable queueing on the connection object, disabling it has no influence on channel, which is always queueing.

File: ably/realtime_channel.go (hunk 24 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

Publish returns Result value, which can be used to await the result, accordingly to Connect and Publish methods; the same API is going to be used for RealtimePresence.

File: ably/realtime_channel.go (hunk 1 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

File: ably/realtime_channel.go (hunk 2 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

File: ably/realtime_conn_test.go (hunk 3 above)
Comment (rjeczalik) [inline ably/realtime_conn_test.go]: Done. The test was rewritten.

File: ably/realtime_conn_test.go (hunk 4 above)
Comment (rjeczalik) [inline ably/realtime_conn_test.go]: Done. The test was rewritten.

File: ably/realtime_conn_test.go (hunk 5 above)
Comment (rjeczalik) [inline ably/realtime_conn_test.go]: Fixed.

File: ably/realtime_conn.go (hunk 6 above)
Comment (rjeczalik) [inline ably/realtime_conn.go]: Firstly RWMutex is slower than Mutex, secondly there's no much to be gain from RLocks on those simple getters - majority of the lock usage is write-access, RWMutex is usefull for single-writer, multiple-readers scenarios.

Before we start to micro-optimize I'd prefer to write benchmarks first so I can be sure about the outcome; probably not going to happen before the functionality is ready first ;)

File: ably/error.go (hunk 7 above)
Comment (rjeczalik) [inline ably/error.go]: Let's pretend no-one seen this. Fixed.

File: ably/error.go (hunk 8 above)
Comment (rjeczalik) [inline ably/error.go]: Done.

File: ably/realtime_channel.go (hunk 11 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

File: ably/realtime_channel.go (hunk 12 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

File: ably/realtime_channel.go (hunk 13 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done. Sorry, old habbits. :(

File: ably/realtime_channel.go (hunk 14 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

File: ably/realtime_channel_test.go (hunk 15 above)
Comment (rjeczalik) [inline ably/realtime_channel_test.go]: Done. The test was rewritten.

File: ably/ably.go (hunk 16 above)
Comment (rjeczalik) [inline ably/ably.go]: Done. I've firstly removed it, but then added it again as it's used.

File: ably/error.go (hunk 17 above)
Comment (rjeczalik) [inline ably/error.go]: Done. "toStatusCode", that is.

File: ably/realtime_channel.go (hunk 22 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

File: ably/realtime_channel.go (hunk 23 above)
Comment (rjeczalik) [inline ably/realtime_channel.go]: Done.

File: ably/realtime_conn.go (hunk 27 above)
Comment (rjeczalik) [inline ably/realtime_conn.go]: Done.

File: ably/realtime_channel.go (hunk 10 above)
Comment (kouno) [inline ably/realtime_channel.go]: :see_no_evil:

File: ably/ably.go (hunk 16 above)
Comment (kouno) [inline ably/ably.go]: Question for you guys: Does Golang have any tool to detect unused private method (dead code)?

It would be nice to have that in our CI too :smile:.

**Comment by kouno** [comment]
Looks like you sorted the feedback, thanks! :+1:

Can you also update the README to mirror the implementation state before I merge this PR?

File: ably/ably.go (hunk 16 above)
Comment (rjeczalik) [inline ably/ably.go]: @kouno If I was to deploy it right now I'll jump on godoc (listing symbols) + oracle (counting references) + bash (exit code 1 if refcount == 0); haven't used anything that works out of a box.

File: ably/realtime_channel.go (hunk 18 above)
Comment (lmars) [inline ably/realtime_channel.go]: SGTM

File: ably/realtime_channel.go (hunk 25 above)
Comment (lmars) [inline ably/realtime_channel.go]: > I don’t want to get rid of slow receivers, I want slow receivers not to leak goroutines started by us

Why would the user being slow lead to them leaking goroutines? The proposed `Subscription` would likely have a `Close()` method which would stop the goroutine, and the user should be encouraged to defer that call, this is a pretty standard pattern:

```go
sub, err := c.Subscribe("foo")
if err != nil {
  return err
}
defer sub.Close()
```

> I consider slow receivers as programming error (insufficient buffering, too much blocking operations) on the user side which I would not want to handle.

You seem to want the user to do extra work when it is not necessary. I'd say it is quite typical for users to want to receive events and then take a non-trivial amount of time to handle those events.

```go
sub, _ := c.Subscribe("foo")
defer sub.Close()
for event := range sub.Events {
  // do some non-trivial processing
}
```

You are expecting the user to do one of two things:

1. Use a buffered channel. How big should this buffer be? I need to predict the rate of incoming events and the time it takes to process an event to determine an appropriate buffer size, and I probably don't know either of those at compile time. Too small a buffer and I start losing messages, which may have serious consequences.

2. Use an unbuffered channel and not be a slow receiver. So if I am processing events in non-trivial time I am going to need to create a goroutine per event. And if I also want to process them in order, I am going to need my own queueing goroutine to avoid blocking the channel, which is more work for me.

I think if you are still against supporting slow receivers, I think you need to make the above considerations clear to the user.

kouno: So I involved @mattheworiordan on this too, and I think that the main problem we have here is that we don't really want to drop any messages at all.

Since Go doesn't provide channels with infinite buffers (AFAIK), it sounds like we will have to go with queues instead.

Another point is that the Queue example that @lmars provided makes the queue private but doesn't define an accessor to know how big the queue is. It would be nice for users to have access to this information too in case they ever run into a situation where they have slow receivers.

rjeczalik:
@lmars I've added `Subscription` implementation + `Len() int` method.

@kouno Please take a look.

mattheworiordan:
@rjeczalik you mentioned this on a commit that is now gone:

> Async queueing implementation is much more complicated that what you’ve described. See [this code](https://github.com/rjeczalik/ably-go/commit/651694f1e2dc8d20f4d7bce46755d6a238a1f265#diff-a5254638ab4b7f5f9e7482b7a5feed87R173). This is no longer just a wrapper. I don’t want to get rid of slow receivers, I want slow receivers not to leak goroutines started by us. I consider slow receivers as programming error (insufficient buffering, too much blocking operations) on the user side which I would not want to handle.

Well yes and no.  It depends on what the side effects are for the client library.  For example, see https://github.com/ably/ably-ruby/blob/master/lib/ably/modules/safe_deferrable.rb#L61-L70 in the Ruby Library. Where you would use channels, in Ruby we are using the [`trigger`](https://github.com/ably/ably-ruby/blob/master/lib/ably/modules/event_emitter.rb#L88) method to emit messages, however as you will see, it uses a [safe method call](https://github.com/ably/ably-ruby/blob/master/lib/ably/modules/safe_yield.rb#L15-L20) so that any exceptions that are raised in the block / Proc that is called is caught by the library, and error is logged, and execution of the library continues.  The point here is that the client library should not break if the developer does something wrong / stupid, it should simply log the problem and move on.

I think the principle is important here as well with this discussion.

ably/realtime_channel.go
rjeczalik: @lmars @kouno Done. Subscribe queues all events.

rjeczalik:
@mattheworiordan Thank you for your explanation. Current API does not call user code, so we don't need to handle runtime panics. The only problem we can have now is the queues grow too large too fast, but I'll add logging for this as well.

(PS the cited comment is [here](https://github.com/ably/ably-go/pull/32#discussion_r30475832))

PR #33, opened 2015-06-17
ably: add RealtimePresence

rjeczalik:
I'm still adding more tests to cover all cases, this CL contains only test for verifying presence event emitting and syncing presence maps between different connections.

More tests will follow in next PRs + after that I'll update features state under https://docs.google.com/spreadsheets/d/1gvN6y0KuzXsN6IOL4LGNbcIeZj-VKZ0U1dAhQrmPz7U/edit#gid=0.

/cc @mattheworiordan @kouno @lmars Please take a look.


ably/realtime_channel.go
```diff
@@ -451,29 +320,27 @@ func (c *RealtimeChannel) notify(msg *proto.ProtocolMessage) {
 	case proto.ActionAttached:
 		c.state.syncSet(StateChanAttached, nil)
 		c.queue.Flush()
+		c.Presence.mu.Lock()
+		c.Presence.syncStart()
+		if msg.Flags&1 == 0 {
+			c.Presence.syncEnd()
```
lmars: Rather than starting then immediately stopping the sync, why not just only start it if `msg.Flags & 1 == 1`?

ably/realtime_channel.go
```diff
@@ -451,29 +320,27 @@ func (c *RealtimeChannel) notify(msg *proto.ProtocolMessage) {
 	case proto.ActionAttached:
 		c.state.syncSet(StateChanAttached, nil)
 		c.queue.Flush()
+		c.Presence.mu.Lock()
```
lmars: I feel like it should be the responsibility of `RealtimePresence` to manage it's own mutex.

ably/realtime_presence.go
```diff
@@ -0,0 +1,230 @@
...
+	if syncSerial == "" {
+		pres.syncEnd()
+	}
+	pres.mu.Unlock()
+	if len(old) != 0 {
+		// Do not announce old presence messages to the subscribers.
+		tmp := make([]*proto.PresenceMessage, 0, len(msg.Presence)-len(old))
+		for i, member := range msg.Presence {
+			if _, ok := old[i]; !ok {
+				tmp = append(tmp, member)
```
lmars: Rather than tracking old ones and excluding them here, you could declare `tmp` further up and just append new messages to it in the above loop.

ably/subscription.go
```diff
@@ -0,0 +1,280 @@
...
+	return ch
+}
+
+// Close unsubscribes from the realtime channel the sub was previously subscribed.
+// It closes the chan returned by C method.
+func (sub *Subscription) Close() error {
+	return sub.close(true)
+}
+
+func (sub *Subscription) dry() {
```
lmars: I think `drain` is a better name for this method.

ably/subscription.go
```diff
@@ -0,0 +1,280 @@
...
+	switch channel := sub.channel.(type) {
```
lmars: Why the switch? Can't you just drain the channel regardless of the type?

README.md
```diff
@@ -39,24 +52,82 @@ for msg := range sub.MessageChannel() {
...
 if err = res.Wait(); err != nil {
 	panic(err)
 }
+```
 
-fmt.Println("message published successfully")
+### Getting all clients present on a channel
+
+```go
+clients := channel.Presence.Get(true)
```
mattheworiordan: This does not quite match our API, see http://www.rubydoc.info/gems/ably/0.8.2/Ably/Realtime/Presence#get-instance_method for an example.  We should be able to provide options, one of which is the `waitForSync` boolean

ably/options.go
```diff
@@ -31,7 +31,7 @@ type ClientOptions struct {
 	RealtimeHost string // optional; overwrite endpoint hostname for Realtime client
 	Environment  string // optional; prefixes both hostname with the environment string
 	Key          string // an authorization key in the 'name:secret' format
-	ClientID     string // optional;
+	ClientID     string // optional; required for realtime presence
```
mattheworiordan: Not entirely true actually, but leave this in.  You can `enterClient` on behalf of another `clientId`.

ably/realtime_channel.go
mattheworiordan: :+1:

ably/realtime_channel.go
```diff
@@ -451,29 +320,27 @@ func (c *RealtimeChannel) notify(msg *proto.ProtocolMessage) {
...
 		c.state.syncSet(StateChanDetached, nil)
+	case proto.ActionSync:
+		syncSerial := ""
+		if i := strings.IndexRune(msg.ChannelSerial, ':'); i != -1 {
+			syncSerial = msg.ChannelSerial[i+1:]
+		}
+		c.Presence.handle(msg, syncSerial)
 	case proto.ActionPresence:
-		// TODO realtime presence
+		c.Presence.handle(msg, "")
```
mattheworiordan: not sure I like the name handle, what exactly is that method doing?  Could you name it more descriptively please?

ably/realtime_presence.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return pres.channel.send(protomsg, result)
+}
+
+func (pres *RealtimePresence) syncStart() {
+	if pres.insync {
+		return
+	}
+	pres.sync.Add(1)
+	pres.insync = true
```
mattheworiordan: `insync` is a bit misleading, I assume `insync` means a sync is in progress and not that the Presence Map is in sync.  Therefore, could we perhaps name it something like `syncInProgress` or `syncComplete` and set it to false?

ably/realtime_presence.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+		}
+		switch member.State {
+		case proto.PresenceUpdate:
+			memberCopy := *member
+			member = &memberCopy
+			member.State = proto.PresencePresent
+			fallthrough
+		case proto.PresencePresent:
+			delete(pres.stale, memberKey)
```
mattheworiordan: I am not sure I follow the logic here.  Why are you deleting the member from the stale list without checking the timestamp?  I would assume that until the sync is complete, we cannot delete stale members?

ably/realtime_presence.go
```diff
@@ -0,0 +1,230 @@
...
+			pres.members[memberKey] = member
+		case proto.PresenceLeave:
+			delete(pres.members, memberKey)
+		}
+	}
+	if syncSerial == "" {
```
mattheworiordan: I don't think this logic is right, see https://github.com/ably/ably-ruby/blob/master/lib/ably/realtime/presence/members_map.rb#L178-L183 for the two variants you can expect

ably/realtime_presence.go
```diff
@@ -0,0 +1,230 @@
...
+		if presence.State == proto.PresenceAbsent {
+			delete(pres.members, memberKey)
+		}
+	}
+	pres.stale = nil
+	pres.insync = false
+	pres.sync.Done()
+}
+
+func (pres *RealtimePresence) handle(msg *proto.ProtocolMessage, syncSerial string) {
```
mattheworiordan: Could this be named better such as `processIncomingMessage` for clarity?

ably/realtime_presence.go
```diff
@@ -0,0 +1,230 @@
...
+				tmp = append(tmp, member)
+			}
+		}
+		msg.Count = len(tmp)
+		msg.Presence = tmp
+	}
+	pres.subs.presenceEnqueue(msg)
+}
+
+func (pres *RealtimePresence) Get(wait bool) []*proto.PresenceMessage {
```
mattheworiordan: As mentioned previously, this is fine for now, but later we need to match the API and allow options

ably/subscription.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+		// Stop subscription if it no longer listens for messages.
+		if count == 0 {
+			sub.close(false)
+		}
+	}
+	subs.mtx.Unlock()
+}
+
+func (subs *subscriptions) messageEnqueue(msg *proto.ProtocolMessage) {
```
mattheworiordan: Could we not simply create an interface called baseMessage that would allow both of these operations (messageEnqueue and presenceEnqueue) to share the same code i.e. everything from line 251 onwards would be the same no?

mattheworiordan:
Great, good progress @rjeczalik

ably/subscription.go
rjeczalik: Unfortunately it’s not possible. You would need to abstract away iteration over items and key retrieval from a single item - it would be more code (one interface + two implementations), it would create intermediate structure for the sake of interface ([]interface{}) and it would be definitely more complex to understand than just two functions with two for loops.

ably/realtime_presence.go
rjeczalik: Done.

ably/realtime_presence.go
rjeczalik: Thanks for checking! Actually It’s ok, the syncSerial is parsed outside, [here](https://github.com/rjeczalik/ably-go/commit/c6980845b958e0b7ea7f0ca0bce17edfbdd57fdb#diff-a5254638ab4b7f5f9e7482b7a5feed87R333).

ably/realtime_presence.go
rjeczalik: When sync is in progress, I remove a member from stale list when it's reported to be present. Then when sync is complete, I remove all stale members from the presence map [here](https://github.com/ably/ably-go/commit/3c65c16491d27c1c5b4c661e4b092bd3e9fdf276#diff-68e24634a97440d2db9491b85323cea2R74).

When sync is completed, this line is a nop, since stale list is a nil, and deleting from nil map is a nop. The stale list is created when we start sync [here](https://github.com/ably/ably-go/commit/3c65c16491d27c1c5b4c661e4b092bd3e9fdf276#diff-68e24634a97440d2db9491b85323cea2R64).

ably/realtime_presence.go
rjeczalik: Done.

ably/realtime_channel.go
rjeczalik: Done.

ably/subscription.go
rjeczalik: Using `reflect` package it's possible. Done.

ably/subscription.go
rjeczalik: Done.

ably/realtime_presence.go
rjeczalik: Done.

ably/realtime_channel.go
rjeczalik: Done.

ably/realtime_channel.go
rjeczalik: Done.

mattheworiordan:
Thanks for addressing all the feedback!

kouno:
Looks good from what I can see :+1:.

PR #35, opened 2015-06-24
fix RealtimePresence race; add StateEnum type

rjeczalik:
Channel may get attached after blocking Get is called on
RealtimePresence - when that happens, Get returns empty presence list.
This CL makes Get also wait for the channel until it's attached.

A StateEnum type was added to have connection/channel states human
readable when printed.

### DIFF
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -115,7 +115,7 @@ var errAttach = newError(90000, errors.New("Attach() on inactive connection"))
 // If channel is already attached, this method is a nop.
 // If sending attach message failed, the returned error value is non-nil.
 // If sending attach message succeed, the returned Result value can be used
-// to wait until result from server is received.
+// to wait until ack from server is received.
 func (c *RealtimeChannel) Attach() (Result, error) {
 	return c.attach(true)
 }
@@ -160,7 +160,7 @@ var errDetach = newError(90000, errors.New("Detach() on inactive connection"))
 // If channel is already detached, this method is a nop.
 // If sending detach message failed, the returned error value is non-nil.
 // If sending detach message succeed, the returned Result value can be used
-// to wait until result from server is received.
+// to wait until ack from server is received.
 func (c *RealtimeChannel) Detach() (Result, error) {
 	return c.detach(true)
 }
@@ -205,7 +205,6 @@ func (c *RealtimeChannel) detach(result bool) (Result, error) {
 func (c *RealtimeChannel) Close() error {
 	err := wait(c.Detach())
 	c.subs.close()
-	c.Presence.stop()
 	if err != nil {
 		return c.state.syncSet(StateChanClosed, newErrorf(90000, "Close() error: %s", err))
 	}
@@ -230,7 +229,7 @@ func (c *RealtimeChannel) Subscribe(names ...string) (*Subscription, error) {
 // Unsubscribe panics if the given sub was subscribed for presence messages and
 // not for regular channel messages.
 //
-// If sub was already unsubscribed the method is a nop.
+// If sub was already unsubscribed, the method is a nop.
 func (c *RealtimeChannel) Unsubscribe(sub *Subscription, names ...string) {
 	if sub.typ != subscriptionMessages {
 		panic(errInvalidType{typ: sub.typ})
@@ -318,12 +317,7 @@ func (c *RealtimeChannel) Reason() error {
 func (c *RealtimeChannel) notify(msg *proto.ProtocolMessage) {
 	switch msg.Action {
 	case proto.ActionAttached:
-		if msg.Flags.Has(proto.FlagPresence) {
-			// since syncStart writes to RealtimePresence's internal WaitGroup
-			// from different goroutine than it's being read we need to lock
-			// it with syncStartLock call instead.
-			c.Presence.syncStartLock()
-		}
+		c.Presence.onAttach(msg.Flags.Has(proto.FlagPresence))
 		c.state.syncSet(StateChanAttached, nil)
 		c.queue.Flush()
 	case proto.ActionDetached:
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -64,7 +64,6 @@ func (ch *Channels) All() []*RealtimeChannel {
 	ch.mtx.Unlock()
 	chanSlice(chans).Sort()
 	return chans
-
 }
 
 // Release closes a channel looked up by the name.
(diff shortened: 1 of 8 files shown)


rjeczalik:
Hey @kouno! I've replaced `sync.WaitGroup` with just a `sync.Mutex` - it even simpler than `sync.Cond`. Please take a look :)

ably/realtime_presence.go
```diff
@@ -11,25 +11,50 @@ import (
...
+			if !pres.observedAttach && !pres.syncInProgress {
+				pres.syncMtx.Unlock() // channel is attached and there's no sync in progress
+				pres.observedAttach = true
+			}
+			pres.mtx.Unlock()
+		}
+	}
+}
+
+func (pres *RealtimePresence) stop() {
```
kouno: Why `stop` and not `close` like we do on the channel level?

ably/realtime_presence.go
```diff
@@ -11,25 +11,50 @@ import (
...
+		event:   make(chan State, 1),
 		members: make(map[string]*proto.PresenceMessage),
 	}
+	pres.syncMtx.Lock() // initially we lock waiting for channel's attached state
+	channel.On(pres.event, StateChanAttached)
+	go pres.eventloop()
+	return pres
+}
+
+func (pres *RealtimePresence) eventloop() {
```
kouno: `processIncomingAttachedEvent`? A bit verbose but `eventloop` is really not that descriptive.

ably/realtime_presence.go
```diff
@@ -156,10 +199,15 @@ func (pres *RealtimePresence) Subscribe(states ...proto.PresenceState) (*Subscri
...
 }
 
+// Unsubscribe removes previous Subscription for the given presence states.
+//
+// If sub was already unsubscribed, the method is a nop.
 func (pres *RealtimePresence) Unsubscribe(sub *Subscription, states ...proto.PresenceState) {
 	pres.subs.unsubscribe(true, sub, statesToKeys(states)...)
 }
 
+// Enter announces presence of the current client altogether with an enter
```
kouno: Not sure the `altogether` is needed.

ably/realtime_presence.go
```diff
@@ -11,25 +11,50 @@ import (
...
+}
+
+func (pres *RealtimePresence) eventloop() {
+	for event := range pres.event {
+		switch event.State {
+		case StateChanAttached:
+			pres.mtx.Lock()
+			if !pres.observedAttach && !pres.syncInProgress {
+				pres.syncMtx.Unlock() // channel is attached and there's no sync in progress
+				pres.observedAttach = true
```
kouno: I'm not sure what the role of `observedAttach` is.

If a channel is attached, then we get detached due to a network issue, this would stay true.

kouno:
Looks ok. Can you address my few comments, especially the one about `observedAttach`?

ably/realtime_presence.go
rjeczalik: The `observedAttach` is there in order to differentiate two attach events - 1) an initial attach (right after the channel gets created), for which we want to block waiting for sync and 2) regular attach event.

But now I see that Get does not return error, when it's called on Detached / Failed channel. I'll address that and see if I can get rid of `observedAttach`.

ably/realtime_presence.go
rjeczalik: Done.

ably/realtime_presence.go
rjeczalik: Done.

ably/realtime_presence.go
rjeczalik: Done.

ably/realtime_presence.go
rjeczalik: Done.

ably/realtime_channel.go
```diff
@@ -318,14 +318,14 @@ func (c *RealtimeChannel) Reason() error {
 func (c *RealtimeChannel) notify(msg *proto.ProtocolMessage) {
 	switch msg.Action {
 	case proto.ActionAttached:
-		c.state.syncSet(StateChanAttached, nil)
-		c.queue.Flush()
 		if msg.Flags.Has(proto.FlagPresence) {
 			// since syncStart writes to RealtimePresence's internal WaitGroup
```
kouno: There is no WaitGroup anymore.

ably/realtime_presence.go
```diff
@@ -25,11 +26,36 @@ type RealtimePresence struct {
...
+		switch event.State {
+		case StateChanAttached:
+			pres.mtx.Lock()
+			if !pres.syncInProgress {
+				// channel is attached and there's no sync in progress
+				pres.syncMtx.Unlock()
+			}
+			pres.mtx.Unlock()
+		case StateChanDetached, StateChanFailed, StateChanClosed:
+			pres.syncMtx.Lock()
```
kouno: You are locking this in a completely separated thread but I don't see where it would get unlocked.

ably/realtime_presence.go
```diff
@@ -25,11 +26,36 @@ type RealtimePresence struct {
...
+	go pres.processIncomingEvents()
+	return pres
+}
+
+func (pres *RealtimePresence) processIncomingEvents() {
+	for event := range pres.event {
+		switch event.State {
+		case StateChanAttached:
+			pres.mtx.Lock()
+			if !pres.syncInProgress {
```
kouno: TBH, it's difficult to get the whole picture from the code.

ably/realtime_presence.go
```diff
@@ -25,11 +26,36 @@ type RealtimePresence struct {
...
 
 func newRealtimePresence(channel *RealtimeChannel) *RealtimePresence {
-	return &RealtimePresence{
+	pres := &RealtimePresence{
 		subs:    newSubscriptions(subscriptionPresenceMessages),
 		channel: channel,
+		event:   make(chan State, 1),
 		members: make(map[string]*proto.PresenceMessage),
 	}
+	pres.syncMtx.Lock() // initially we lock waiting for channel's attached state
```
kouno: I think you should avoid using short comments like this and explain it in a global comment at the top of your function.

kouno:
@rjeczalik I feel like this is not far off, but at the same time, every time I look at the locking system for incoming events, my head starts to spin :/. Especially since you added https://github.com/ably/ably-go/pull/35#discussion-diff-34672316R52.

rjeczalik:
@kouno This lock is direct consequence of the rewrite `sync.WaitGroup` -> `sync.Mutex` and I agree mutex operations are more confusing in the context. I take a step back look the the whole functionality and try to rewrite it to be more understandable.

rjeczalik:
Hey @kouno! Added `syncState` member, now I it's readable why extra blocking for the `initialState` state. I've also took opportunity and fixed erroneous rest test, which tried to marshal `*http.Request` - this is going to fail for go1.5.

Please take a look.

kouno:
:+1: a lot easier to follow thanks.

I added an issue for the channel state check, but I think we should close this PR which has been there for far too long. Can you sort out the incomplete message and merge?

PR #37, opened 2015-07-21
realtime: attach to channel on Presence.Get

### LINKED ISSUE #36: realtime: Presence.Get() should attach instead of raising error
https://github.com/rjeczalik/ably-go/commit/2a02bffb824ec242a7cbdce2c7d6796943bbd378#commitcomment-12249969

rjeczalik:
Fixes #36.

### DIFF
--- a/ably/helpers_test.go
+++ b/ably/helpers_test.go
@@ -18,6 +18,93 @@ func (opts *ClientOptions) RealtimeURL() string {
 	return opts.realtimeURL()
 }
 
+// StateRecorder provides:
+//
+//   * send ably.State channel for recording state transitions
+//   * goroutine-safe access to recorded state enums
+//
+type StateRecorder struct {
+	ch     chan State
+	states []StateEnum
+	wg     sync.WaitGroup
+	done   chan struct{}
+	mtx    sync.Mutex
+}
+
+func NewStateRecorder() *StateRecorder {
+	return newRecorderBuffer(16)
+}
+
+func newRecorderBuffer(buffer int) *StateRecorder {
+	rec := &StateRecorder{
+		ch:     make(chan State, buffer),
+		done:   make(chan struct{}),
+		states: make([]StateEnum, 0, 16),
+	}
+	rec.wg.Add(1)
+	go rec.processIncomingStates()
+	return rec
+}
+
+func (rec *StateRecorder) processIncomingStates() {
+	defer rec.wg.Done()
+	for {
+		select {
+		case state, ok := <-rec.ch:
+			if !ok {
+				return
+			}
+			rec.add(state.State)
+		case <-rec.done:
+			return
+		}
+	}
+}
+
+// Add appends state to the list of recorded ones, used to ensure ordering
+// of the states by injecting values at certain points of the test.
+func (rec *StateRecorder) Add(state StateEnum) {
+	rec.ch <- State{State: state}
+}
+
+func (rec *StateRecorder) add(state StateEnum) {
+	rec.mtx.Lock()
+	rec.states = append(rec.states, state)
+	rec.mtx.Unlock()
+}
+
+func (rec *StateRecorder) Channel() chan<- State {
+	return rec.ch
+}
+
+func (rec *StateRecorder) Options() *ClientOptions {
+	return &ClientOptions{Listener: rec.ch}
+}
+
+// Stop stops separate recording gorouting and waits until it terminates.
+func (rec *StateRecorder) Stop() {
+	close(rec.done)
+	rec.wg.Wait()
+	// Drain listener channel.
+	for {
+		select {
+		case <-rec.ch:
+		default:
+			return
+		}
+	}
+}
+
+// States gives copy of the recorded states, safe for use while the recorder
+// is still running.
+func (rec *StateRecorder) States() []StateEnum {
+	rec.mtx.Lock()
+	defer rec.mtx.Unlock()
+	states := make([]StateEnum, len(rec.states))
+	copy(states, rec.states)
+	return states
+}
+
 // MustRealtimeClient is like NewRealtimeClient, but panics on error.
 func MustRealtimeClient(opts *ClientOptions) *RealtimeClient {
 	client, err := NewRealtimeClient(opts)
@@ -27,9 +114,9 @@ func MustRealtimeClient(opts *ClientOptions) *RealtimeClient {
 	return client
 }
 
-// TestGet is a helper method, which returns attached channel or panics if
+// GetAndAttach is a helper method, which returns attached channel or panics if
 // the attaching failed.
-func (ch *Channels) TestGet(name string) *RealtimeChannel {
+func (ch *Channels) GetAndAttach(name string) *RealtimeChannel {
 	channel := ch.Get(name)
 	if err := Wait(channel.Attach()); err != nil {
 		panic(`attach to "` + name + `" failed: ` + err.Error())
--- a/ably/helpers_test.go
+++ b/ably/helpers_test.go
@@ -1,6 +1,11 @@
 package ably
 
-import "sync"
+import (
+	"fmt"
+	"reflect"
+	"sync"
+	"time"
+)
 
 func Wait(res Result, err error) error {
 	return wait(res, err)
@@ -31,15 +36,21 @@ type StateRecorder struct {
 	mtx    sync.Mutex
 }
 
-func NewStateRecorder() *StateRecorder {
-	return newRecorderBuffer(16)
-}
-
-func newRecorderBuffer(buffer int) *StateRecorder {
+// NewStateRecorder gives new recorder which purpose is to record states via
+// (*ClientOptions).Listener channel.
+//
+// If buffer is > 0, the recorder will use it as a buffer to ensure all states
+// transitions are received.
+// If buffer is <= 0, the recorder will not buffer any states, which can
+// result in some of them being dropped.
+func NewStateRecorder(buffer int) *StateRecorder {
+	if buffer < 0 {
+		buffer = 0
+	}
 	rec := &StateRecorder{
 		ch:     make(chan State, buffer),
 		done:   make(chan struct{}),
-		states: make([]StateEnum, 0, 16),
+		states: make([]StateEnum, 0, buffer),
 	}
 	rec.wg.Add(1)
 	go rec.processIncomingStates()
@@ -77,10 +88,6 @@ func (rec *StateRecorder) Channel() chan<- State {
 	return rec.ch
 }
 
-func (rec *StateRecorder) Options() *ClientOptions {
-	return &ClientOptions{Listener: rec.ch}
-}
-
 // Stop stops separate recording gorouting and waits until it terminates.
 func (rec *StateRecorder) Stop() {
 	close(rec.done)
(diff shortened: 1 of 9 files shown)


rjeczalik:
Btw I've noticed red build on master. Test test that failed was querying ably servers for time and asserting the time is within +/- one second range (assumed latency). Since Travis warns the build was running on old architecture, I've updated the .travis.yml to have builds running on their new infracstructure (more details here http://docs.travis-ci.com/user/migrating-from-legacy/). This should help with the failure on master.

/cc @kouno Please take a look.

ably/helpers_test.go
```diff
@@ -18,6 +18,93 @@ func (opts *ClientOptions) RealtimeURL() string {
...
+
+func (rec *StateRecorder) Options() *ClientOptions {
+	return &ClientOptions{Listener: rec.ch}
+}
+
+// Stop stops separate recording gorouting and waits until it terminates.
+func (rec *StateRecorder) Stop() {
+	close(rec.done)
+	rec.wg.Wait()
+	// Drain listener channel.
```
kouno: After draining, shouldn't we close the channel?

ably/helpers_test.go
```diff
@@ -18,6 +18,93 @@ func (opts *ClientOptions) RealtimeURL() string {
...
+	wg     sync.WaitGroup
+	done   chan struct{}
+	mtx    sync.Mutex
+}
+
+func NewStateRecorder() *StateRecorder {
+	return newRecorderBuffer(16)
+}
+
+func newRecorderBuffer(buffer int) *StateRecorder {
```
kouno: I'm not sure why this needs to be its own function. It would probably be as good to have this directly in NewStateRecorder.

rjeczalik: @kouno Sure, I'll merge it.

ably/helpers_test.go
```diff
@@ -18,6 +18,93 @@ func (opts *ClientOptions) RealtimeURL() string {
...
+	rec.mtx.Lock()
+	rec.states = append(rec.states, state)
+	rec.mtx.Unlock()
+}
+
+func (rec *StateRecorder) Channel() chan<- State {
+	return rec.ch
+}
+
+func (rec *StateRecorder) Options() *ClientOptions {
```
kouno: IMO, the recorder shouldn't provide its own set of ClientOptions. It is handy, but writing `ClientOptions{Listener: rec.ch}` looks more explicit.

ably/realtime_presence_test.go
```diff
@@ -96,3 +104,38 @@ func TestRealtimePresence_Sync250(t *testing.T) {
...
+	ably.StateConnConnected,
+	stateGuard,
+	ably.StateChanAttaching,
+	ably.StateChanAttached,
+}
+
+func TestRealtimePresence_EnsureChannelIsAttached(t *testing.T) {
+	rec := ably.NewStateRecorder()
+	opts := rec.Options()
+	opts.NoConnect = true
```
kouno: See my point above about the ClientOptions in the recorder.

ably/helpers_test.go
rjeczalik: @kouno No, we can't close it, the ownership of the channel is transferred to (*ClientOptions).Listener; the client may still send some events, like `StateConnClosing` etc., but we don't care about them since we stopped recording. We don't need to keep reading the channel because `ably.Client`'s event emitter sends state values in a non-blocking manner.

ably/helpers_test.go
rjeczalik: @kouno Ack

ably/realtime_presence_test.go
rjeczalik: Ack

ably/realtime_presence_test.go
```diff
@@ -96,3 +104,38 @@ func TestRealtimePresence_Sync250(t *testing.T) {
...
+	app, client := testutil.ProvisionRealtime(nil, opts)
+	defer safeclose(t, client, app)
+	channel := client.Channels.Get("persisted:presence_fixtures")
+	if err := ably.Wait(client.Connection.Connect()); err != nil {
+		t.Fatal(err)
+	}
+	rec.Add(stateGuard)
```
kouno: I'm not sure I'm following why this is needed. :(

stateGuard is 0, which as far as I know is not assigned to any state. Why do you need to add this state in your expectation?

ably/realtime_presence_test.go
```diff
@@ -96,3 +104,38 @@ func TestRealtimePresence_Sync250(t *testing.T) {
 		t.Fatalf("members3: %v", err)
 	}
 }
+
+var stateGuard = ^ably.StateEnum(0)
+
+var presTransitions = []ably.StateEnum{
```
kouno: Shouldn't this be within `TestRealtimePresence_EnsureChannelIsAttached` since you are never going to use it outside of this test?

ably/realtime_presence_test.go
```diff
@@ -96,3 +104,38 @@ func TestRealtimePresence_Sync250(t *testing.T) {
 		t.Fatalf("members3: %v", err)
 	}
 }
+
+var stateGuard = ^ably.StateEnum(0)
```
kouno: Same as below.

ably/state.go
```diff
@@ -29,7 +29,7 @@ func (st StateType) String() string {
 	case StateChan:
 		return "channel"
 	default:
-		return ""
+		return "invalid"
```
kouno: I suppose this exists because you added this fake stateGuard in your tests. Empty string looked like the valid approach though.

rjeczalik: Using empty string will hide bugs, user should never see invalid when he/she prints state received from the `ably.Client`. Seeing "invalid" could help troubleshooting. My original version was just wrong.

ably/realtime_presence_test.go
rjeczalik: Event ordering - to ensure I received `StateChanAttached` after calling to `Presence.Get` - if client.Channels.Get automatically attached a channel, the event sequence will be the same, the test would pass but the functionality will be broken - the attach must be initiated here by `Presence.Get`. And inserting marks helps on that :)

ably/realtime_presence_test.go
rjeczalik: Ack

ably/realtime_presence_test.go
rjeczalik: Ack

ably/state.go
kouno: I would have said that in that case, it should be logged instead as a warning or an error. We tend to do that a lot in other client libraries.

The thing is that we have not a proper logger at this point yet. Unless I missed it.

kouno:
This looks good in general. Address the feedback and that should be good to go.

ably/realtime_presence_test.go
kouno: Oh. I see.

In that case, how about doing an expectation here instead? Only check that the first two states are done.

ably/state.go
rjeczalik: @kouno: Logger is implemented and handled in ClientOptions, but I didn't deployed it yet - that is I'm going to add logging after the lib is finished, to not keep reworking the logging.

While I agree exceptional, but non-vital, situations should be logged, we can't log from the inside of the model. At least I think so.

kouno: > we can't log from the inside of the model. At least I think so.

I understand your point, but that would be a default behaviour across our libraries. We always have a global logger to do such things. Take a look at https://github.com/ably/ably-java/blob/870c33991d0a66b182b7f3196d99c21e39d814a1/src/io/ably/http/TokenAuth.java which uses `Log`, a globally accessible logger (https://github.com/ably/ably-java/blob/245a3f20a6dce0d34413ddfed19c5da8ea647422/src/io/ably/util/Log.java).

ably/realtime_presence_test.go
rjeczalik: Hm, sorry, I don't see it. I'll recap: I need to check in the test that StateChanAttached after the call to Presence.Get. Since the event sequence is [Connecting, Connected, Attaching, Attached] I need to somehow differentiate scenarios where I have attached chanel before `Presence.Get` or after `Presence.Get` (where the latter is only correct); by inserting mark after connecting I ensure that I receive either:

-  [Connecting, Connected, Attaching, Attached, mark] for the former
-  [Connecting, Connected, mark, Attaching, Attached] for the latter

I see two options: either I use mark to ensure ordering or I start listening for events after Connect and before `Presence.Get` (but I would need to drain the events first to ensure nothing is hanging asynchronously). I think the latter is more code that's why I went with first option.

If there's other way of implementing the test I don't see it.

kouno: You could verify that `rec.States()` only contains [Connecting, Connected] at this stage instead of adding the marker. That would do exactly the same thing as adding the marker and verifying the whole chain in the end.

kouno: Just to be clear, your implementation of the recorder is fine. I'm just saying that we could drop the marker which makes no sense right away when reading the code.

ably/helpers_test.go
kouno: :+1:

ably/realtime_presence_test.go
rjeczalik: It's more subtle than that. I can't verify if rec.States() contains only [Connecting, Connected] since there may be states values buffered in the chan but not yet added to the slice. States() then will return [Connecting, Connected] and we'll be happy, but as soon as it returns the buffered in the channel states will then be added to the slice before `presence.Get` fires. What makes mark work is that it's sent via the same channel, that is it's enqueued at the end. Obviously it's not 100% accurate test, to be 100% sure we would need to introduce long timeouts before acknowledging the states, but in my opinion it's good enough for our case.

If we really don't want to enforce ordering that way, I surely can rewrite this as you suggested. However this way of testing is much easier to write more complicated tests which assert behaviour of the connection/channel due to various external factors (connection recovery), as it's as simple as compare all recorded events to single slice instead of comparing partial results each step (I'm actually using this more).

Besides the above I have no strong feelings for the mark or absence of it, can rewrite it if you feel it would make maintanence & development of ably-go easier :)

ably/helpers_test.go
rjeczalik: Done.

rjeczalik:
@kouno: All addressed, except whether to mark or not to mark; please update me how we proceed :)

ably/state.go
rjeczalik: This will be covered by #38.

ably/realtime_presence_test.go
```diff
@@ -96,3 +104,38 @@ func TestRealtimePresence_Sync250(t *testing.T) {
...
+		Listener:  rec.Channel(),
+		NoConnect: true,
+	}
+	app, client := testutil.ProvisionRealtime(nil, opts)
+	defer safeclose(t, client, app)
+	channel := client.Channels.Get("persisted:presence_fixtures")
+	if err := ably.Wait(client.Connection.Connect()); err != nil {
+		t.Fatal(err)
+	}
+	rec.Add(beforePresenceGet)
```
kouno: > It's more subtle than that. I can't verify if rec.States() contains only [Connecting, Connected] since there may be states values buffered in the chan but not yet added to the slice. States() then will return [Connecting, Connected] and we'll be happy, but as soon as it returns the buffered in the channel states will then be added to the slice before presence.Get fires. What makes mark work is that it's sent via the same channel, that is it's enqueued at the end. Obviously it's not 100% accurate test, to be 100% sure we would need to introduce long timeouts before acknowledging the states, but in my opinion it's good enough for our case.

Since we are doing parallel processing here, it sounds sensible to have an expectation that `eventually` resolves the states should equal [Connecting, Connected]. In ginkgo that is very easy to do http://onsi.github.io/gomega/#eventually.

kouno: Having tests that rely on such implementation details is bad IMO. We should avoid that.

rjeczalik: @kouno Removed mark, please take a look.

kouno:
Looks good. :+1:

PR #39, opened 2015-08-12
MsgPack support; reworked Stat data type; updated deps

rjeczalik:
During adding full support for MsgPack I've encountered a couple of decoding failures, at first though it may be library problems thus I've updated all deps. Then it turned out to be a problem with decoding maps with custom unmarshallers - the problem is workarounded with this CL. Moreover it fixes one nil ptr deref, reworks Stats data struct fixing problems with MsgPack encoding of the old struct and add ones test.

Originally I tried to aim at adding more tests to have covered fully the Go Spec for realtime client, but eventually found out I need to fix those blockers first.

/cc @mattheworiordan @lmars @kouno Please take a look.

PS Rework of the Go Spec summary is still in progress, I'll update you when I'm done.


mattheworiordan:
Overall looks like it's moving in the right direction, still quite a lot to do though!

rjeczalik:
@mattheworiordan: I've update Travis, please take a look.

The test failure:

```
--- FAIL: TestRealtimePresence_EnsureChannelIsAttached (3.05s)
	realtime_presence_test.go:135: clientID="client_bool" not found in presence map
```

is related to #35, the reworked implementation with sync.Mutex gives flaky results, I'll need to change it back to sync.WaitGroup in seperate PR (@kouno).

ably/paginated_result.go
```diff
@@ -59,7 +62,15 @@ func newPaginatedResult(typ reflect.Type, path string, params *PaginateParams,
 	p.path = builtPath
 	p.links = resp.Header["Link"]
 	v := reflect.New(p.typ)
-	if err = json.NewDecoder(resp.Body).Decode(v.Interface()); err != nil {
+	switch typ := strip(resp.Header.Get("Content-Type"), ';'); typ {
```
lmars: Use [mime.ParseMediaType](https://golang.org/pkg/mime/#ParseMediaType) (and remove the `strip` function).

ably/rest_client.go
```diff
@@ -88,8 +90,18 @@ func (c *RestClient) handleResp(v interface{}, resp *http.Response, err error) (
...
+	proto := c.Auth.options.protocol()
+	if typ := strip(resp.Header.Get("Content-Type"), ';'); typ != "" && typ != protoMIME[proto] {
+		return nil, newError(40000, errors.New("unrecognized Content-Type: "+typ))
+	}
+	switch proto {
+	case ProtocolJSON:
+		err = json.NewDecoder(resp.Body).Decode(v)
+	case ProtocolMsgPack:
+		err = msgpack.NewDecoder(resp.Body).Decode(v)
+	}
```
lmars: Probably worth putting this in a helper function like:

```go
decode(typ string, r io.Reader, v interface{}) error
```

ably/auth.go
```diff
@@ -33,6 +42,18 @@ func (c *Capability) UnmarshalJSON(p []byte) error {
...
+}
+
+func (c *Capability) UnmarshalMsgpack(p []byte) error {
+	var s string
+	if err := msgpack.Unmarshal(p, &s); err != nil {
+		return err
+	}
+	return json.Unmarshal([]byte(s), (*map[string][]string)(c))
+}
+
```
lmars: Why are you implementing these methods? Does msgpack not do the correct thing? If not, this deserves a comment.

ably/auth.go
```diff
@@ -42,21 +63,87 @@ func (c Capability) String() string {
...
+func (tok *Token) UnmarshalMsgpack(p []byte) error {
+	// This method is workaround for msgpack decoder, which tries to decode fields
+	// using its original type and (finding field byte boundary) and then passes
+	// the bytes to custom unmarshaller.
+	//
+	// Decoding *Token directly fails with:
+	//
+	//   msgpack: invalid code ab decoding map length
+	//
```
lmars: I suspect the issue is that `Token.Capability` has type `Capability`, but you have defined the custom `UnmarshalMsgpack` on `*Capability`, so msgpack is using the default unmarshaler for a `map[string][]string` but passing it an encoded JSON string (as that is what `Capability.MarshalMsgpack` returns).

I'm pretty sure if you just drop the custom marshaler methods on `Capability`, this will just work.

rjeczalik: @lmars It would not, Capability is a map[string]string and the value we're receiving is a string, that's why custom unmarshaller.

What I'd do instead is to have `RawCapability string` here so it works like RawQuery in url.URL and I'd drop those unmarshallers.

ably/auth.go
rjeczalik: @lmars: See comment below, Capability is sent as stringified JSON.

ably/paginated_result.go
rjeczalik: Done.

ably/rest_client.go
rjeczalik: Done.

ably/auth.go
lmars: @rjeczalik ah ok, I didn't realise the capabilities were being sent as JSON strings from the API, I assumed they would be msgpack maps.

So you could make `Token.Capability` a `*Capability`, then your Unmarshaler would be called.

ably/auth.go
lmars: Ok, I think it's worth adding a comment that the capabilities are sent as JSON msgpack strings rather than msgpack maps, hence the need for the custom Marshal methods.

ably/auth.go
rjeczalik: @lmars I've already rewritten this getting rid of the marshal / unmarshal dance - https://github.com/rjeczalik/ably-go/commit/247486357d1a8011d1347306c226aed51f607f4a#diff-fe17adcb123ff94418b17414c9d8940dR14. The capabilities are kept as encoded string with method helper `(*TokenRequest).Capability() Capability` - similarly to how `(url.URL).RawQuery` + `(url.URL).Query() url.Values` works.

I'm going to send the rewrite in separate PR. With fix for the flaky test.

ably/auth.go
rjeczalik: @lmars I think it's not needed for now. See comment below.

rjeczalik:
Hey @lmars @mattheworiordan, is this PR ok to merge or you're waiting for me?

ably/auth.go
```diff
@@ -42,21 +63,87 @@ func (c Capability) String() string {
 }
 
 type Token struct {
```
mattheworiordan: This is a `TokenDetails` struct, not a `Token` struct (it is confusing to have a `Token` struct with a `Toke` property). See TokenDetails under http://docs.ably.io/client-lib-development-guide/features/#types.  Can we ensure consistency with other client libraries here?

mattheworiordan:
@rjeczalik sorry, didn't realise you were waiting.  Looks good, nice to see MsgPack support.  I added one more comment, but that could land in a separate PR as it's not necessarily related to this PR. LGTM. :+1:

ably/auth.go
rjeczalik: @mattheworiordan I already got that covered - I've reworked whole Auth [on a separate branch](https://github.com/rjeczalik/ably-go/commit/2f76de11810e7710483a4dd6673a2fd43e0e1b4d#diff-fe17adcb123ff94418b17414c9d8940dR111), the PR will follow after this one.

lmars:
:+1:

PR #61, opened 2018-08-27
Fix resource count  value types

gernest:
- removed unused variable
- changed resource count  values to use `float54`

### DIFF
--- a/ably/proto/stat.go
+++ b/ably/proto/stat.go
@@ -5,9 +5,9 @@ import "time"
 type Stat struct {
 	// The following fields hold statistical information only for Channels
 	// and Connections fields, for the rest they are going to be always zero.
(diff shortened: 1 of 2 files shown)


paddybyers:
Why not change all relevant members to `float64` as part of this change, to align with the declared types in https://github.com/ably/ably-java/blob/develop/lib/src/main/java/io/ably/lib/types/Stats.java ?

gernest:
@paddybyers The goal right now is to get the tests to pass. This is one blocker out of the way.

Eventually, I will add `(TS1) - (TS9)` which will properly group the value in their respective structs. In a nutshell, the other parts are not blocking the tests so I think they are not relevant here yet.

paddybyers:
ok, thanks

PR #65, opened 2018-08-28
Use new  msgpack package

gernest:
This PR introduces github.com/ugorji/go/codec  as mspack encoder/decoder

### DIFF
--- a/ably/internal/ablyutil/msgpack.go
+++ b/ably/internal/ablyutil/msgpack.go
@@ -1,0 +1,38 @@
+package ablyutil
+
+import (
+	"bytes"
+	"io"
+
+	"github.com/ugorji/go/codec"
+)
+
+var handle codec.MsgpackHandle
+
+// Unmarshal decodes the MessagePack-encoded data and stores the result in the
+// value pointed to by v.
+func Unmarshal(data []byte, v interface{}) error {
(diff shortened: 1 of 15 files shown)


ably/realtime_client_test.go
```diff
@@ -90,7 +90,7 @@ func checkUnique(ch chan string, typ string, n int) error {
 }
 
 func TestRealtimeClient_50clients(t *testing.T) {
```
ORBAT: Might as well rename the test

ably/internal/ablyutil/msgpack.go
```diff
@@ -0,0 +1,38 @@
+package ablyutil
+
+import (
+	"bytes"
+	"io"
+
+	"github.com/ugorji/go/codec"
```
ORBAT: Should be vendored

gernest: No, not in this PR. I plan to make Godep directory go away by the time we integrate the integration branch to develop and use the new `go module` to manage dependencies.

But that work will be done later.

ORBAT: Right, makes sense

ORBAT:
As a note, this PR was necessary because the original msgpack decoder would only decode double values to a `float64`, so e.g. `1` would fail to decode

PR #66, opened 2018-08-29
fix ci

gernest:
- start with go 1.5.3
- use make test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,13 @@
 language: go
 
 go:
-  - 1.4.3
   - 1.5.2
 
+before_install:
+  - go get -t -v ./ably/...
+  - go get golang.org/x/tools/cmd/vet
 env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,10 +1,12 @@
...
+  - 1.10.3
 
+before_install:
+  - go get -t -v ./ably/...
 env:
   - GOMAXPROCS=4
-
 sudo: false
+script:
+  - make test
```
paddybyers: Trailing newline needed here?

paddybyers:
Just one meta-comment: please can we have more descriptive names for branches and PRs.

.travis.yml
gernest: nope, that is valid yaml and travis is happy with it. I can add it if you think it will serve any purpose.

PR #67, opened 2018-08-29
RSC7a add version header

gernest:


ably/rest_client_test.go
```diff
@@ -195,3 +197,25 @@ var _ = Describe("RestClient", func() {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Run("must set version header", func(ts *testing.T) {
+		req, err := c.NewHTTPRequest(&ably.Request{})
+		if err != nil {
+			ts.Fatal(err)
+		}
+		h := req.Header.Get("X-Ably-Version")
+		expect := "1.0"
```
paddybyers: Should we use a constant so that this test doesn't break when we change the library version?

ably/rest_client_test.go
```diff
@@ -195,3 +197,25 @@ var _ = Describe("RestClient", func() {
...
+	c, err := ably.NewRestClient(app.Options())
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Run("must set version header", func(ts *testing.T) {
+		req, err := c.NewHTTPRequest(&ably.Request{})
+		if err != nil {
+			ts.Fatal(err)
+		}
+		h := req.Header.Get("X-Ably-Version")
```
paddybyers: Most of the other test suites explicitly check that the header is sent, instead of just checking that it is set on the request. That could be done in two ways:

- start a webserver locally;
- use the echo server (https://github.com/ably/echoserver).

The second option is better but I don't see anything in there at the moment that will echo request headers, although this is easily added.

@ORBAT any thoughts?

gernest: @paddybyers that is not really necessary. Because this is the only place throughout the library where the request is composed.

So, what you see is a stated fact. We are only checking that our facts are correct.

ably/rest_client_test.go
gernest: There are constants, just not exported. I had a different thought on this, but I think your argument is stronger will update to use the constants.

ably/rest_client_test.go
ORBAT: Unit testing `NewHTTPRequest` like this is in my opinion a good idea, but we do probably _also_ want to have tests that actually do HTTP requests. https://golang.org/pkg/net/http/httptest/ would probably be good for building something like the echo server

gernest: @ORBAT I agree with you. What I was trying to say is It will be an overkill to do that with this test. We are only interested to see if the header is set, and it is set within the function. If the header was set by the server that is a different story.

ORBAT: My thoughts exactly

paddybyers:
I should have mentioned that in the other libraries (eg the java one https://github.com/ably/ably-java/blob/develop/lib/src/main/java/io/ably/lib/transport/Defaults.java) we have a large number of constants and defaults that are defined somewhere, but don't pollute the API of the top-level Ably object. I don't see anything similar for ably-go, but perhaps I missed it.

gernest:
@paddybyers  there is no such a thing yet. But we will get there. As you can see I am learning about the library while working on small patches.

Some are not relevant to this PR. For instance `ABLY_LIB_VERSION` , which ill make sense to add at the time of release.

Please let's keep the discussion narrow on the scope of this PR.  I will make one more change, regarding naming, `ABLY_VERSION` and `ABLY_VERSION_HEADER` for consistency.

PR #73, opened 2018-08-31
Update stats object to 1.1 spec

gernest:
This implements

- TS1
- TS2
- TS3
- TS4
- TS5
- TS6
- TS7
- TS8
- TS9

### DIFF
--- a/ably/proto/stat.go
+++ b/ably/proto/stat.go
@@ -2,20 +2,6 @@ package proto
 
 import "time"
 
-type Stats struct {
-	Count         int64           `json:"count" codec:"count"`
-	Unit          string          `json:"unit" codec:"unit"`
-	IntervalID    string          `json:"intervalId" codec:"intervalId"`
-	All           MessageTypes    `json:"all" codec:"all"`
-	APIRequests   RequestCount    `json:"apiRequests" codec:"apiRequests"`
-	Channels      ResourceCount   `json:"channels" codec:"channels"`
-	Connections   ConnectionTypes `json:"connections" codec:"connections"`
-	Inbound       MessageTraffick `json:"inbound" codec:"inbound"`
-	Outbound      MessageTraffick `json:"outbound" codec:"outbound"`
-	Persisted     MessageTypes    `json:"persisted" codec:"persisted"`
-	TokenRequests RequestCount    `json:"tokenRequests" codec:"tokenRequests"`
-}
-
 const (
 	StatGranularityMinute = "minute"
 	StatGranularityHour   = "hour"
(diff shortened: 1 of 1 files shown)


gernest:
@paddybyers @ORBAT I think this is ready for review.

ably/proto/stat.go
```diff
@@ -69,3 +21,125 @@ var (
...
+	Outbound MessageTraffic `json:"outbound" codec:"outbound"`
+}
+
+type XchgMessages struct {
+	All          MessageTypes `json:"all" codec:"all"`
+	ProducerPaid MessageTypes `json:"producerPaid" codec:"producerPaid"`
+	ConsumerPaid MessageTypes `json:"consumerPaid" codec:"consumerPaid"`
+}
+
+type BaseStats struct {
```
paddybyers: This is superflous here, unless you can make `Stats` extend `BaseStats` and remove the common fields.

SimonWoolf:
> what's your view on incuding the exchange-related stats types in the libraries? I'm inclined to include them while we're making updates, unless we're not confident that the structs are stable

Happy for them to be included, unless we want to start telling people to use untyped/flattened stats if they want exchange stats (and add support for that to client libs), per https://github.com/ably/wiki/issues/177

ably/proto/stat.go
```diff
@@ -69,3 +21,125 @@ var (
...
+type MessageDirections struct {
+	All      MessageTypes   `json:"all" codec:"all"`
+	Inbound  MessageTraffic `json:"inbound" codec:"inbound"`
+	Outbound MessageTraffic `json:"outbound" codec:"outbound"`
+}
+
+type XchgMessages struct {
+	All          MessageTypes `json:"all" codec:"all"`
+	ProducerPaid MessageTypes `json:"producerPaid" codec:"producerPaid"`
+	ConsumerPaid MessageTypes `json:"consumerPaid" codec:"consumerPaid"`
```
SimonWoolf: producerpaid and consumerpaid are messagedirections, not messagetypes

ably/proto/stat.go
gernest: my bad, will remove that .

gernest:
@SimonWoolf @paddybyers  I addressed  the comments. Please let me know if there is a need for more changes, thanks.

mattheworiordan:
>  what's your view on incuding the exchange-related stats types in the libraries? I'm inclined to include them while we're making updates, unless we're not confident that the structs are stable

I am dead against this.  This creates a lot more work for developers unnecessarily, given we have no evidence anyone is using the stats from the libs, and more importantly, the libs have no way of getting account level stats anyway.

I would prefer we instead added a new beta API perhaps that allowed access to untyped/flattened stats, and moved us closer towards an API we intend on supporting moving forwards.

PR #74, opened 2018-09-03
call Close on sandbox when exiting TestRSC7 test

gernest:


gernest:
@ORBAT please check this.

PR #77, opened 2018-09-04
use constants for message encoding strings

gernest:


ORBAT:
`"utf-8"` and a couple of other magic string literals are also used in `Message.EncodeData` and `DecodeData`:

https://github.com/ably/ably-go/blob/develop/ably/proto/message.go#L39-L88

PR #78, opened 2018-09-05
add rsc15 support host fallback

gernest:
This implement

 - [x] `rsc15b`
 - [x] `rsc15e`
 - [x] `rsc15a`
 - [x] `rsc15d`
 - [x] `TO3l5`

### DIFF
--- a/ably/options.go
+++ b/ably/options.go
@@ -27,6 +27,7 @@ var defaultOptions = &ClientOptions{
 		"d.ably-realtime.com",
 		"e.ably-realtime.com",
 	},
+	HTTPMaxRetryCount: 3,
 	RealtimeHost:      "realtime.ably.io",
 	TimeoutConnect:    15 * time.Second,
 	TimeoutDisconnect: 30 * time.Second,
@@ -164,6 +165,9 @@ type ClientOptions struct {
 	Recover                 string        // optional; used to recover client state
 	Logger                  LoggerOptions // optional; overwrite logging defaults
 	TransportParams         map[string]string
+
+	// max number of fallback hosts to use as a fallback w
+	HTTPMaxRetryCount int
 
 	NoTLS            bool // when true REST and realtime client won't use TLS
 	NoConnect        bool // when true realtime client will not attempt to connect automatically
--- a/ably/options.go
+++ b/ably/options.go
@@ -237,9 +237,6 @@ func (opts *ClientOptions) restURL() string {
 			host = opts.Environment + "-" + host
 		}
 	}
-	if opts.NoTLS {
-		return opts.scheme() + host
-	}
 	return opts.scheme() + host
 }
 
--- a/ably/options.go
+++ b/ably/options.go
@@ -19,19 +19,23 @@ const (
 )
 
 var defaultOptions = &ClientOptions{
-	RestHost: "rest.ably.io",
-	FallbackHosts: []string{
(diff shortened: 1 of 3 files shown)


ably/options.go
```diff
@@ -148,13 +156,18 @@ func (opts *AuthOptions) KeySecret() string {
...
+	FallbackHostsUseDefault bool
+	FallbackHosts           []string
+	RealtimeHost            string        // optional; overwrite endpoint hostname for Realtime client
+	Environment             string        // optional; prefixes both hostname with the environment string
+	ClientID                string        // optional; required for managing realtime presence of the current client
+	Recover                 string        // optional; used to recover client state
+	Logger                  LoggerOptions // optional; overwrite logging defaults
+	TransportParams         map[string]string
+
+	// max number of fallback hosts to use as a fallback w
```
ORBAT: Typo `w`

ably/rest_client.go
```diff
@@ -140,24 +144,88 @@ func (c *RestClient) do(r *Request) (*http.Response, error) {
...
+					strings.HasSuffix(req.URL.Host, defaultOptions.RestHost)) {
+				fallback := defaultOptions.FallbackHosts
+				if c.opts.FallbackHosts != nil {
+					fallback = c.opts.FallbackHosts
+				}
+				if len(fallback) > 0 {
+					// we copy the hosts to avoid accidentally mutating the array while
+					// running.
+					left := make([]string, len(fallback))
+					copy(left, fallback)
```
ORBAT: The `copy` is unnecessary since slices are value types: simply doing `left := fallback` will create a copy of the `fallback` slice header value. When you assign `left = n` below, it'll overwrite the slice header in `left` but leave `fallback` intact. This is OK because you're not mutating the _contents_ of the slice, only the header

ORBAT: `do()` is getting a bit long, so you might want to break out the host picking code into a function/method of its own

gernest: @ORBAT  I don't think there will be benefit of breaking down `do` I prefer it this way because even if we move host picking code into a function it is only going to be used here so inlining the logic makes sense and also helps when reading and comparing notes to the spec. I addressed the other comment.

gernest:
ping @ORBAT

ably/rest_client_test.go
```diff
@@ -229,3 +228,49 @@ func TestRSC7(t *testing.T) {
...
+	defer app.Close()
+	t.Run("RSC15d must use alternative host", func(ts *testing.T) {
+		var retryCount int
+		var hosts []string
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if retryCount > 0 {
+				hosts = append(hosts, r.Host)
+			}
+			retryCount++
+			fmt.Println(r.Host)
```
ORBAT: Stray `Println`

ORBAT: Stray `Println`

ably/options.go
```diff
@@ -221,9 +238,18 @@ func (opts *ClientOptions) restURL() string {
 		}
 	}
 	if opts.NoTLS {
```
ORBAT: Unneeded `if`

gernest: @ORBAT  why ?

gernest: Oh! I got it sorry

PR #81, opened 2018-09-06
 add more tests for rsc15

gernest:


gernest:
@paddybyers @ORBAT I think this is ready for review. I covered all the features added.

paddybyers:
This is getting there but please:

- in the test code, reference the spec item as a comment: this isn't done for RSC15a;
- you're testing that multiple fallbacks are tried, but the test isn't verifying that there are no repeated requests to the same fallback;
- I think the RSC15a test with a client-specified fallback list should use different hosts altogether; otherwise the test could pass even if that functionality wasn't there. Also, multiple client-specified fallbacks should be tested.

Separately, I think the logic for client-specified primary host (overriding `rest.ably.io`) might be broken - it needs to handle a non-default `environment` as well as a non-default `host` - I think perhaps the code you have will use `rest.ably.io` even if the client specified a non-default `environment`. Perhaps this is a different PR?

PR #84, opened 2018-09-12
add support for  rest channels

gernest:


ably/rest_client_test.go
```diff
@@ -377,3 +377,45 @@ func TestRest_hostfallback(t *testing.T) {
 		}
 	})
 }
+
+func TestRestChannels(t *testing.T) {
```
ORBAT: Generally you should try to have the spec "code" (like `RSN1`) somewhere in the test and/or subtest name, and channel names for each test should be tied to the test name somehow (so e.g. `rsn1_first_channel` or whatever in this case) so that parallel tests (which the library doesn't seem to currently use, but should probably be considered) don't interfere with each other

ably/rest_client.go
```diff
@@ -43,9 +43,54 @@ func query(fn func(string, interface{}) (*http.Response, error)) QueryFunc {
 	}
 }
 
-type RestClient struct {
-	Auth *Auth
+// RestChannels provides an API for managing collection of RestChannel. This is
+// safe for concurrent use.
+type RestChannels struct {
+	cache  *sync.Map
```
ORBAT: Change to normal `map` + `sync.RWMutex` as discussed on Slack

ably/rest_client.go
```diff
@@ -43,9 +43,67 @@ func query(fn func(string, interface{}) (*http.Response, error)) QueryFunc {
...
+	mu     sync.RWMutex
+	client *RestClient
+}
 
+// Range iterates over the channels calling fn on every iteration. If fn returns
+// false then the iteration is stopped.
+func (c *RestChannels) Range(fn func(name string, channel *RestChannel) bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	for k, v := range c.cache {
```
ORBAT: If anybody calling `Range` tries to modify `c` in `fn` (via either `Release` or "indirectly" via `Get`), this would lead to a deadlock. One way to handle this would be to:

* create a temporary map with size `len(c.cache)`
* `RLock`
* "clone" `c.cache` into the temp map with a `range`
* `RUnlock`
* `range` over the temp map and call `fn` at this point

This way concurrent modification from inside `fn` won't cause deadlocks. You should probably also add a test where you `Release` a channel/`Get` a channel that's not in `cache` from inside `fn`

PR #88, opened 2018-09-24
Update message encoding and encrypt channel message

gernest:
- [x] RSL4c
  - [x] RSL4c1
  - [x] RSL4c2
  - [x] RSL4c3
  - [x] RSL4c4
- [x] RSL4d
  - [x] RSL4d1
  - [x] RSL4d2
  - [x] RSL4d3
  - [x] RSL4d4

- [x] RSL5
  - [x] RSL5a


ably/proto/message.go
```diff
@@ -22,12 +27,223 @@ type Message struct {
...
 	ConnectionID string                 `json:"connectionId,omitempty" codec:"connectionID,omitempty"`
 	Name         string                 `json:"name,omitempty" codec:"name,omitempty"`
-	Data         string                 `json:"data,omitempty" codec:"data,omitempty"`
+	Data         *DataValue             `json:"data,omitempty" codec:"data,omitempty"`
 	Encoding     string                 `json:"encoding,omitempty" codec:"encoding,omitempty"`
 	Timestamp    int64                  `json:"timestamp" codec:"timestamp"`
 	Extras       map[string]interface{} `json:"extras" codec:"extras"`
 }
 
+type DataValue struct {
```
ORBAT: If this type is really necessary (see above), document the type and why/where it's used

ably/proto/message.go
```diff
@@ -22,12 +27,223 @@ type Message struct {
 	ClientID     string                 `json:"clientId,omitempty" codec:"clientId,omitempty"`
 	ConnectionID string                 `json:"connectionId,omitempty" codec:"connectionID,omitempty"`
 	Name         string                 `json:"name,omitempty" codec:"name,omitempty"`
-	Data         string                 `json:"data,omitempty" codec:"data,omitempty"`
+	Data         *DataValue             `json:"data,omitempty" codec:"data,omitempty"`
```
ORBAT: What's the reasoning behind having a separate `DataValue` type, instead of having this functionality in `Message`?

ably/proto/message.go
```diff
@@ -44,18 +260,22 @@ func (m *Message) MemberKey() string {
 // was encrypted.
 func (m *Message) DecodeData(opts *ChannelOptions) error {
```
ORBAT: It still feels a bit strange that `EncodeData` and `DecodeData` are exported methods, since I really wouldn't expect users of the package to ever have to call them

ORBAT: What happens if you call `DecodeData` twice on the same `Message`?

ably/proto/message.go
```diff
@@ -94,11 +314,19 @@ func (m *Message) DecodeData(opts *ChannelOptions) error {
 //
 // Any errors encountered in any step will be returned immediately.
 func (m *Message) EncodeData(encoding string, opts *ChannelOptions) error {
```
ORBAT: What happens if `EncodeData` is called with the same parameters on the same message twice?

ably/proto/message.go
```diff
@@ -22,12 +27,273 @@ type Message struct {
 	ClientID     string                 `json:"clientId,omitempty" codec:"clientId,omitempty"`
 	ConnectionID string                 `json:"connectionId,omitempty" codec:"connectionID,omitempty"`
 	Name         string                 `json:"name,omitempty" codec:"name,omitempty"`
-	Data         string                 `json:"data,omitempty" codec:"data,omitempty"`
+	Data         *DataValue             `json:"data,omitempty" codec:"data,omitempty"`
```
ORBAT: As it currently stands, the exposed API around `Message` is a bit clunky and it leaks implementation details to the outside. The fact that it needs a `*DataValue` means that you can't just instantiate them with e.g. `&Message{Data: []byte{1,2,3}}`, and since there's no "constructor" function, instantiating messages is harder than it should be. The `DataValue` type, the various "internal" but public methods etc. are implementation details that users shouldn't have to deal with.

So as before, my suggestion would be:

-  have an exported `Message` type that has `Data interface{}`, with a simple public API for creating new messages and working with them, and no exported "internal" methods. 

- have an `InternalMessage` (or whatever) type (this can even be as simple as `type InternalMessage proto.Message` + methods for `InternalMessage` that convert the data to a `DataValue` etc.) in an internal package under `ably/internal` that provides all the needed shared functionality that we don't want to expose an API for (so marshaling/unmarshaling, encoding/decoding etc.), but need to have shared in different parts of the library itself. It would also need easy conversions from/to `InternalMessage`. All public-facing APIs should only return the exported "dumb" `Message` type

ably/proto/message_test.go
```diff
@@ -306,90 +306,177 @@ func TestDataValue(t *testing.T) {
...
+				}
 			}
-			if !reflect.DeepEqual(value.Data.Value, v.src) {
-				ts.Errorf("expected %v got %v", v.src, value.Data.Value)
+		})
+		ts.Run("with default hints", func(ts *testing.T) {
+			sample := []struct {
+				src    interface{}
+				expect string
+				empty  interface{}
```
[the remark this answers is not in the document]
ORBAT: Generally there seems to be a lot of shared code between the tests. Might want to consider factoring the common bits out and having systematic test cases that share the same payload but have different expectations for JSON and msgpack

ably/proto/message.go
ORBAT: Note that ^ isn't 100% necessary to do in this PR, but I believe it really needs to be done eventually

ably/ablytest/resultgroup.go
```diff
@@ -51,7 +51,9 @@ func (rg *ResultGroup) Add(res ably.Result, err error) {
 	if !rg.check(err) {
 		return
 	}
+	rg.mu.RLock()
 	rg.wg.Add(1)
+	rg.mu.RUnlock()
```
ORBAT: This'll mask the actual problem here. See discussion in slack

ably/proto/message.go
```diff
@@ -18,161 +21,329 @@ const (
...
+		return []byte(v), nil
+	default:
+		return nil, fmt.Errorf("coerceBytes should never get anything but strings or []byte, got %T", i)
 	}
-	return keylen
 }
 
-func (m *Message) decrypt(cipherStr string, opts *ChannelOptions) error {
-	cipher, err := opts.GetCipher()
+func (m Message) ToMap() map[string]interface{} {
```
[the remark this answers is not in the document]
gernest: @ORBAT `FromMap` is used outside the `proto` package, figured I can export `ToMap` as well, it just seemed weird to have  public `FromMap` and private `ToMap`.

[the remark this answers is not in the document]
ORBAT: That makes sense. Let's keep it like this, then

gernest:
@ORBAT I'm going to rebase this on integration if that is okay with you.

PR #93, opened 2018-10-01
prefer the ably error codes generated by scripts/errors.go

gernest:


ably/error.go
```diff
@@ -49,7 +49,7 @@ func newError(code int, err error) *Error {
 		return err
 	case net.Error:
 		if err.Timeout() {
-			return &Error{Code: ErrCodeTimeout, StatusCode: 500, Err: err}
+			return &Error{Code: ErrNotFound, StatusCode: 500, Err: err}
```
ORBAT: Is 500 the correct status for `ErrNotFound`?

PR #95, opened 2018-10-03
rsa10k cache server time offset

gernest:


ably/auth.go
```diff
@@ -284,19 +293,51 @@ func (a *Auth) setDefaults(opts *AuthOptions, req *TokenRequest) error {
...
 		}
+		req.Timestamp = Time(ts)
 	}
 	return nil
 }
 
+//Timestamp returns the timestamp to be used in authorization request.
+func (a *Auth) Timestamp(query bool) (time.Time, error) {
+	var now time.Time
+	if a.Now != nil {
```
paddybyers: I'm trying to understand this. It looks to me like you've declared `Auth:Now()`. So what does this test do?

gernest: @paddybyers `Now` is a helper in case the user wants to supply the time function. The reason behind this is to allow me to write better test coverage of the caching feature. This function allows us to set default values in case `Now` is not supplied. Please see the tests, I think I wouldn't have been able to make them that way without this.

I didn't want to use real clock, also I was avoiding `client.Time()` call because that is testsed somewhere else so not relevant here. The most important bit is caching and calculating time based on cached duration.

paddybyers: > `Now` is a helper in case the user wants to supply the time function

Isn't there a way to add that without making it part of the public API?

PR #103, opened 2018-10-22
add href attribute to ErrorInfo

gernest:
- [x] `T14`
- [x] `T15`

### DIFF
--- a/ably/proto/error.go
+++ b/ably/proto/error.go
@@ -1,6 +1,9 @@
 package proto
 
-import "fmt"
+import (
+	"bytes"
+	"fmt"
+)
 
 // ErrorInfo describes an error returned via ProtocolMessage.
 type ErrorInfo struct {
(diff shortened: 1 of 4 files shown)


ably/proto/error.go
```diff
@@ -26,6 +33,18 @@ func (e *Error) FromMap(ctx map[string]interface{}) {
...
+func (e *ErrorInfo) Error() string {
+	var buf bytes.Buffer
+	buf.WriteString("[ErrorInfo")
+	if e.Message != "" {
+		fmt.Fprintf(&buf, ":%s", e.Message)
+	}
+	if e.StatusCode != 0 {
+		fmt.Fprintf(&buf, ": statusCode=%d", e.StatusCode)
+	}
+	if e.Code != 0 {
```
paddybyers: The spec is awaiting some expanded/clarified requirements for this. In the meantime please look at https://github.com/ably/ably-ruby/commit/ea425be4a855d68649a81f161c4483ce6a38ab99 for what needs to be implemented and tested.

When you receive an ErrorInfo `err`, you determine the error message to display as follows:

```
if `err` contains a non-empty `href` then `error_href = err.href`
else if `err` contains a non-empty `code` then `error_href` = `"https://help.ably.io/error/#{err.code}"`

if `err_href` is defined, and `err.message` does not contain `err_href`, then append `" -> see #{error_href} for help"` to `err.message`.

PR #104, opened 2018-10-23
add Request() API (RSC19)

gernest:
- [x] HttpPaginatedResponse
  - [x] `HP1` 
  - [x] `HP2`
  - [x] `HP3` 
  - [x] `HP4` 
  - [x] `HP5` 
  - [x] `HP6` 
  - [x] `HP7` 
  - [x] `HP8`  

- [x] `RSC19`
  - [x] `RSC19a`
  - [x] `RSC19b`
  - [x] `RSC19c`
  - [x] `RSC19d`
  - [x] `RSC19e`

### DIFF
--- a/ably/paginated_result.go
+++ b/ably/paginated_result.go
@@ -71,7 +71,7 @@ func newPaginatedResult(opts *proto.ChannelOptions, typ reflect.Type, path strin
 		p.respHeaders = make(http.Header)
 	}
 	p.statusCode = resp.StatusCode
-	p.success = 200 < p.statusCode && p.statusCode < 300
+	p.success = 200 <= p.statusCode && p.statusCode < 300
 	copyHeader(p.respHeaders, resp.Header)
 	if h := p.respHeaders.Get(AblyErrorCodeHeader); h != "" {
 		i, err := strconv.Atoi(h)
--- a/ably/paginated_result.go
+++ b/ably/paginated_result.go
@@ -48,7 +48,7 @@ type PaginatedResult struct {
 }
 
 func newPaginatedResult(opts *proto.ChannelOptions, typ reflect.Type, path string, params *PaginateParams,
-	query QueryFunc, log *LoggerOptions) (*PaginatedResult, error) {
+	query QueryFunc, log *LoggerOptions, respCheck func(*http.Response) error) (*PaginatedResult, error) {
 	p := &PaginatedResult{
 		typ:    typ,
 		query:  query,
@@ -63,7 +63,7 @@ func newPaginatedResult(opts *proto.ChannelOptions, typ reflect.Type, path strin
 	if err != nil {
 		return nil, err
 	}
-	if err = checkValidHTTPResponse(resp); err != nil {
+	if err = respCheck(resp); err != nil {
 		return nil, err
 	}
 	defer resp.Body.Close()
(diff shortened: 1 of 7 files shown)


ably/paginated_result.go
```diff
@@ -60,6 +67,22 @@ func newPaginatedResult(opts *proto.ChannelOptions, typ reflect.Type, path strin
 		return nil, err
 	}
 	defer resp.Body.Close()
+	if p.respHeaders == nil {
+		p.respHeaders = make(http.Header)
+	}
+	p.statusCode = resp.StatusCode
+	p.success = 200 < p.statusCode && p.statusCode < 300
```
paddybyers: `200 <= p.statusCode ....`

ably/rest_client.go
```diff
@@ -183,6 +186,31 @@ type Request struct {
...
 	NoRenew bool
+	header  http.Header
+}
+
+// Request sends http request to ably.
+// spec RSC19
+func (c *RestClient) Request(method string, path string, params *PaginateParams, body interface{}, headers http.Header) (*HTTPPaginatedResponse, error) {
+	method = strings.ToUpper(method)
+	switch method {
+	case "GET", "POST", "PUT", "PATCH": // spec RSC19a
```
paddybyers: also `DELETE`: https://github.com/ably/docs/pull/493

ably/paginated_result.go
```diff
@@ -80,11 +74,10 @@ func newPaginatedResult(opts *proto.ChannelOptions, typ reflect.Type, path strin
 			}
 			rs[k] = m
 		}
-		p.typItems = rs
-		return p, nil
+		return rs, nil
 	case presMsgType:
```
paddybyers: I don't really understand why there is special-case code in here to handle `Message` and `PresenceMessage` whereas other types (eg `Stats`) are handled differently

ably/rest_client.go
```diff
@@ -320,13 +341,21 @@ func (c *RestClient) NewHTTPRequest(r *Request) (*http.Request, error) {
 		}
 		body = bytes.NewReader(p)
 	}
-	req, err := http.NewRequest(r.Method, c.opts.restURL()+r.Path, body)
+	path := r.Path
+	if !strings.HasPrefix(path, c.opts.restURL()) {
```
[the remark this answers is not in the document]
gernest: nope, and it is not necessary at all. I will remove it.

ably/rest_client.go
```diff
@@ -426,5 +426,8 @@ func decodeResp(resp *http.Response, out interface{}) error {
 	if err != nil {
 		return err
 	}
-	return decode(typ, resp.Body, out)
+	var buf bytes.Buffer
+	io.Copy(&buf, resp.Body)
```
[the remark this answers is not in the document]
gernest: nothing much was there for debugging, will remove it.

ably/paginated_result.go
gernest: Because `Message` and `PresenceMessage`  requires special handling when decoding.

```go
		var o []map[string]interface{}
		err := decodeResp(resp, &o)
		if err != nil {
			return nil, err
		}
		rs := make([]*proto.Message, len(o))
		for k, v := range o {
			m := &proto.Message{
				ChannelOptions: opts,
			}
			err = m.FromMap(v)
			if err != nil {
				return nil, err
			}
			rs[k] = m
		}
		return rs, nil
```

see, we also have to pass the `ChannelOptions` object in case the message will need decryption. Same applies to `PresenceMessage`.  For other values we fallback to default `json.Unmarshal`.

This is implementation detail so I don't think it will affect the user's API. 
I'm happy to hear other suggestion on a better way to handle this situation.

PR #105, opened 2018-10-29
Add idempotent publishing

gernest:
- [x] `RSL1j`
- [x] `RSL1k`
  - [x] `RSL1k1`
  - [x] `RSL1k2`
  - [x] `RSL1k3`
  - [x] `RSL1k4`
  - [x] `RSL1k5`

### DIFF
--- a/ably/rest_channel.go
+++ b/ably/rest_channel.go
@@ -1,7 +1,10 @@
 package ably
 
 import (
+	"fmt"
 	"strings"
+
+	"github.com/ably/ably-go/ably/internal/ablyutil"
 
 	"github.com/ably/ably-go/ably/proto"
 )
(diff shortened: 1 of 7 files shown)


ably/rest_channel.go
```diff
@@ -59,6 +62,37 @@ func (c *RestChannel) PublishAll(messages []*proto.Message) error {
 			v.ChannelOptions = c.options
 		}
 	}
+	useIdempotent := !c.client.opts.NoIdempotentRestPublishing
+	if useIdempotent {
+		switch len(messages) {
+		case 1:
+			// spec RSL1k2 we preserve the id if we have one message and it contains the
+			// id.
+			if messages[0].ID == "" {
```
paddybyers: I assumed that if there was no `id` present in a message then it would be null, and not the empty string. The empty string isn't a valid id.

Also, I'd not noticed before, but why `ID` and not `Id` ? I realise it needs to have the initial capital, but the field is declared in the spec as `id`, and we should be making the minimum change to the spec to meet the language requirements. Similarly we have `ClientID` instead of `ClientId` etc.

ably/rest_channel_test.go
```diff
@@ -187,4 +196,299 @@ func TestRestChannel(t *testing.T) {
 			}
 		}
 	})
+
+}
+
+func TestIdempotentPublishing(t *testing.T) {
+	t.Parallel()
+	app, err := ablytest.NewSandboxWIthEnv(nil, "idempotent-dev")
```
paddybyers: Once this functionality is integrated we will be removing any special-casing of this environment. Instead of having the custom environment specified in more than one place it would be best please if there can be a single constant or switch we can change to restore the default.

ably/rest_channel_test.go
```diff
@@ -187,4 +196,299 @@ func TestRestChannel(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer app.Close()
+	options := app.Options()
+	client, err := ably.NewRestClient(options)
```
paddybyers: I don't see any of these tests specify `NoIdempotentRestPublishing`. I realise that it defaults to `true`, but I would prefer it to be explicit. It also means there are no tests with `NoIdempotentRestPublishing=false`.

gernest: well unless I missed your point. That code has `NoIdempotentRestPublishing=false`.

ably/rest_channel.go
gernest: strings can not be `nil` so lack of id is an empty string. only pointers or interfaces can be nil(`null`).

PR #110, opened 2018-11-07
[WIP]  remember successful fallback host

gernest:


paddybyers:
> * the cache timeout option should be configurable in ClientOptions timeouts section

See https://github.com/ably/docs/pull/554/commits/0dfd55f1c41dfe6c07eeae1e2224b8d546c4b949

ably/rest_client_test.go
```diff
@@ -388,6 +388,71 @@ func TestRest_hostfallback(t *testing.T) {
...
+		// the same cached host is used again
+		err = channel.Publish("pong", "ping")
+		if err != nil {
+			ts.Fatal(err)
+		}
+		cachedHost = client.GetCachedFallbackHost()
+		if cachedHost != fallbacks[3] {
+			ts.Errorf("expected cached host to be %s got %s", fallbacks[3], cachedHost)
+		}
+		if retryCount != 1 {
```
paddybyers: I don't understand this. It looks like this test is expecting the behaviour to be:

- first call to `publish()`
  - primary host called; returns 500
  - fallback[0] called; returns 500
  - ...
  - fallback[2] called; returns 500
  - fallback[3] called; returns success
- `retry_count = 0`
- second call to `publish()`
  - .... (something happens) *
  - fallback[3] called; returns success
- check `retry_count = 1`

So that means that some failing endpoint must be called at the (something happens) point. But why? The requirement is that it would call fallback[3] without calling anything else.

paddybyers: No, that's wrong. The spec says "Future HTTP requests during that period should use that host". The point of the requirement is that if you have to use a fallback, and that succeeds, then you continue to use the fallback, for all requests, for a period of x minutes before retrying the primary endpoint; the primary is only retried after that time.

gernest: Mmh!  Okay, my bad I misunderstood what the spec meant. I do think `RSC15e` contradicts `RSC15f` though.

paddybyers: > I do think `RSC15e` contradicts `RSC15f` though.

I agree - I'll add a clarification

paddybyers: See https://github.com/ably/docs/commit/d3c7ce0fecea9d73c1f8c8138c22cafae743edb7

ably/rest_client_test.go
```diff
@@ -449,7 +449,7 @@ func TestRest_rememberHostFallback(t *testing.T) {
 			ts.Errorf("expected cached host to be %s got %s", fallbacks[3], cachedHost)
 		}
 		if retryCount != 0 {
-			ts.Errorf("expected to 0 retries got %d retries", retryCount)
+			ts.Errorf("expected  0 retries got %d retries", retryCount)
```
paddybyers: I think there's a double space in here

paddybyers:
> * it feels like we’ve polluted the REST client interface with the new success fallback attribute, unless it’s private

it's `successFallbackHost` which in my understanding is private because it starts with a lower case.

PR #111, opened 2018-11-14
Add 1.1 feature support matrix

paddybyers:


mattheworiordan:
@paddybyers I added @Srushtika and @SimonWoolf to this PR for review.

I think we can do better than this for a few reasons:

* Why is `Feature support` the most important thing shown to a new developer who uses this client library.  That is not the most important thing they need to know about, both because someone new to a library probably wants to know how to use it & install it first, but also because why are we focussing on the negative (of what's missing) first.  
* There is no need to show users what features are supported, only what is not supported. A user is not arriving knowing anything about our release schedules in most cases.  All they care about is the things I mentioned above, and any caveats.  This section should be a caveat.  I do think we should consider a separate table we maintain in one place which keeps a list of all unsupported features across all libraries in one place, but we can do that once the missing feature bit is done for all libs.
* You state "The 1.1 specification includes multiple new features, and this library implements a subset of those features as follows.".  As above, the reader doesn't know about 1.0 or 0.8.  So this is not needed.  
* You state "The 1.1 specification also contains numerous spec clarifications, corrections and minor feature additions.".  The reader does not care.  Who is this README for?

mattheworiordan:
@paddybyers ^?

paddybyers:
@mattheworiordan updated: https://github.com/ably/ably-go/pull/111/commits/001749742a37d28d0b1f5dea78a0c6ff1eee3185

PR #114, opened 2018-12-18
Make IdempotentRestPublishing default to false for 1.1

paddybyers:
In a recent specification change, `IdempotentRestPublishing` will only become `true` by default in library version 1.2. This PR makes the default `false`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
 all: vet build test
 
 vet:
-	go get golang.org/x/tools/cmd/vet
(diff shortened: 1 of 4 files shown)


mattheworiordan:
@paddybyers LGTM, any view on failures?

paddybyers:
> LGTM, any view on failures? They appear to be unrelated

This looks like a flaky test - it passes for me locally, so needs further investigation

PR #116, opened 2019-01-04
Specify msgpack decode option to return strings for appropriate types

paddybyers:
We need to specify `RawToString: true`: https://github.com/ugorji/go/blob/master/codec/msgpack.go#L521

### DIFF
--- a/ably/internal/ablyutil/msgpack.go
+++ b/ably/internal/ablyutil/msgpack.go
@@ -12,6 +12,7 @@ var handle codec.MsgpackHandle
 func init() {
 	handle.Raw = true
 	handle.WriteExt = true
+	handle.RawToString = true
 }
 
 // Unmarshal decodes the MessagePack-encoded data and stores the result in the
--- a/ably/rest_client_test.go
+++ b/ably/rest_client_test.go
@@ -109,15 +109,15 @@ func TestRestClient(t *testing.T) {
 			if err != nil {
 				ts.Fatal(err)
 			}
-			name := anyMsgPack[0]["name"].([]byte)
-			data := anyMsgPack[0]["data"].([]byte)
+			name := anyMsgPack[0]["name"].(string)
+			data := anyMsgPack[0]["data"].(string)
 
 			expectName := "ping"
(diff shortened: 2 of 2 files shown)


paddybyers:
> Now we really dont want this behavior. I considered this before and decided it should not be used because strings in go are `utf-8` meaning this will mangle any raw bytes encoded in msgpack.

OK, but I think you're misreading the intent of the option. In older msgpack versions there was no distinction between binary and string values, and they were labelled as a single "raw" type. Then the spec was changed to introduce separate string and binary types, and there was a migration path where the legacy "raw" types could be interpreted as either a string or binary value. See https://github.com/msgpack/msgpack/blob/master/spec.md#upgrading-messagepack-specification

This option in the library affects the decoding of various opcode ranges, particularly the opcodes that were historically raw types. The opcode for the "client1" member in the message data was `0xa6`, which is in the `FixStr` range (see https://github.com/ugorji/go/blob/master/codec/msgpack.go#L38) - our msgpack encoder in the Ably system correctly uses this opcode for short fixed-length strings, so we need these decoded as strings. The `RawToString` option controls this - see https://github.com/ugorji/go/blob/master/codec/msgpack.go#L520. It does not mean that every binary type is decoded to string type; the binary opcodes are all unconditionally decoded to bytes - see https://github.com/ugorji/go/blob/master/codec/msgpack.go#L528.

paddybyers:
> Please make sure the tests are passing. I see a nasty panic on the ci logs.

> > panic: interface conversion: interface {} is string, not []uint8 [recovered]
	panic: interface conversion: interface {} is string, not []uint8

Agreed, I need to look at this.

paddybyers:
> > I see a nasty panic on the ci logs.
> 
> > > panic: interface conversion: interface {} is string, not []uint8 [recovered]
> > > panic: interface conversion: interface {} is string, not []uint8

See https://github.com/ably/ably-go/pull/116/commits/eb6dba91057e1d29024e4858da6324184b99e27e: this was an error in the test, incorrectly expecting some members of decoded messages to have `[]byte` type.

PR #117, opened 2019-01-04
Use modules to remove vendored websocket library

paddybyers:
@gernest Should this commit include `go.mod` ?


gernest:
@paddybyers yes, `go.mod` should be included

paddybyers:
@gernest final question: is there anything we should add to the README to indicate that this is now set up to use modules? Does it affect usage in any way?

gernest:
@paddybyers nope, this is backward compatible. Although the ones using go modules will benefit the most. The installation process using `go get` will work for both go modules and old go versions ( The one we already support)

paddybyers:
ok, thanks

PR #137, opened 2020-03-25
bug: unpinned range scope variables in tests

audiolion:
Ran https://github.com/kyoh86/scopelint to find unpinned variables and corrected
them.

When a closure is used inside of a goroutine the closure is bound once and the
variable can get out of sync. Variables declared inside the body of a loop are
not shared between iterations, so we pin the value inside the loop body to
prevent this bug from happening. See
https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables
for more details.

### DIFF
--- a/ably/proto/message_test.go
+++ b/ably/proto/message_test.go
@@ -86,6 +86,8 @@ func TestMessage(t *testing.T) {
 	}
 
 	for _, v := range sample {
+		// pin
+		v := v
 		t.Run(v.desc, func(ts *testing.T) {
 			msg := &proto.Message{
 				Data:           v.data,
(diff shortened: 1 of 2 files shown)


audiolion:
> Can you add a comment suffix to each of your new lines with // pin by way of explanation for those less familiar with this particular issue?

Sure, just `// pin` or a longer description or a link to it? They should also be able to view git blame to see the details in my commit message as well that are more fleshed out.

> Could you tell me if you've seen these unpinned variables cause issues at runtime or whether this bug is only validated through the eyes of the linter you were running over the code?

These were found by the linter, and not as a result of a runtime bug. The goal is to ensure if the code is used in an asynchronous fashion that it won't cause subtle bugs.

> Could you also tell me if these were the only issues spotted by the linter?

Yes these were the only three variables that were unpinned.

QuintinWillison:
Looking elsewhere it seems simply `// pin` is sufficient and, indeed, you're right re git blame. Thanks.

PR #138, opened 2020-03-26
bug: TestRestClient Stats failure due to race cond with sandbox

audiolion:
The Stats test sends a [POST request](https://github.com/ably/ably-go/blob/664f3c9e5529cec11429d9d5af02a93b8704f1e6/ably/rest_client_test.go#L184) to the ably sandbox, and then follows up with
a [GET request](https://github.com/ably/ably-go/blob/664f3c9e5529cec11429d9d5af02a93b8704f1e6/ably/rest_client_test.go#L191). If the GET request happens too quickly after the POST, the sandbox
will not have populated data yet and the Stats request will not return data,
which causes the test to fail.

A loop is added that will call client.Stats every 500ms and check for a
nonempty result. If after 10s no result is present the test will timeout.

Failure from running the tests on my local machine:

image

edit: looks like the test failure is in an unrelated part of the codebase, I have seen that test case fail on [other commits](https://travis-ci.org/github/ably/ably-go/jobs/639412366#L547) in the repo, likely a bug that surfaces when `-race` creates the right conditions. The linked build failure was from a changelog update commit so the failure is definitely not related to code changes.

### DIFF
--- a/ably/rest_client_test.go
+++ b/ably/rest_client_test.go
@@ -3,6 +3,7 @@ package ably_test
 import (
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


QuintinWillison:
Thanks for the contribution, @audiolion ... I've asked @tcard to take a look at this one as I'm reluctant to add 'arbitrary' waits to tests unless we've exhausted all other possible avenues to guarantee state synchronisation deterministically.

ably/rest_client_test.go
```diff
@@ -187,6 +187,11 @@ func TestRestClient(t *testing.T) {
 		}
 		res.Body.Close()
 
+		// Artificial delay to provide sandbox time to populate stats data that
+		// was just POST. This does not negatively impact test timings because
+		// tests are run in parallel.
+		<-time.After(time.Second * 5)
```
lmars: Rather than a 5 second sleep, we should call `client.Stats` in a loop with a small sleep that returns when there are some stats or when a timeout is hit. Let me know if you'd like me to provide an example of that.

audiolion: That sounds like a good solution to me!

QuintinWillison: Thanks both. 👍

ably/rest_client_test.go
```diff
@@ -187,22 +188,53 @@ func TestRestClient(t *testing.T) {
...
-			ScopeParams: ably.ScopeParams{
-				Start: ably.Time(longAgo),
-				Unit:  proto.StatGranularityMinute,
-			},
-		})
-		if err != nil {
+		statsCh := make(chan []*proto.Stats, 1)
+		errCh := make(chan error, 1)
+
+		go func() {
```
audiolion: used a goroutine here so that I can return out of the for-select loop, rather than use `break label`

lmars: I actually think a `break` would make this a little more concise:

```go
// poll the stats endpoint for 10 seconds until we get some stats
var stats []*proto.Stats
timeout := 10 * time.Second
for start := time.Now(); time.Since(start) < timeout; time.Sleep(time.Second) {
	longAgo := lastInterval.Add(-120 * time.Minute)

	page, err := client.Stats(&ably.PaginateParams{
		Limit: 1,
		ScopeParams: ably.ScopeParams{
			Start: ably.Time(longAgo),
			Unit:  proto.StatGranularityMinute,
		},
	})
	if err != nil {
		ts.Fatal(err)
	}
	if s := page.Stats(); len(s) > 0 {
		stats = s
		break
	}
}
if len(stats) == 0 {
	ts.Fatal("failed to get stats after %s", timeout)
}
```

tcard:
This should be good to merge, since the failing test is flaky and unrelated to this change.

PR #139, opened 2020-04-01
Fix flaky tests

tcard:
This should get us more steadily green CI builds.


QuintinWillison:
@lmars, if you have a few minutes to take a quick look then that would be appreciated. Thanks!

PR #141, opened 2020-04-01
Implement and test RTC1a, RTC1b.

tcard:
This was done a few weeks back, when we hadn't [formulated a plan](https://ably.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=SDK&selectedIssue=FEA-391) yet and I just went "well let's just implement the spec from top to bottom".

Also, this inaugurates the v1.2 branch. ~A new major version is required due to breaking changes. The ".2" refers to the spec minor version it's supposed to implement: 2.2~ We'll be releasing this as a minor version, in spite of the breaking changes, so that it matches the spec version.

@QuintinWillison This is just renaming a couple variable and fixing a constructor signature.

Going forwards, so that we can get a spec-compliant API out as quickly as possible, my plan is to _append_ the new API alongside/on top of the old one, and, after that's done, just make the old one private in one stroke. We can then clean that up progressively.

### DIFF
--- a/ably/options.go
+++ b/ably/options.go
@@ -23,14 +23,14 @@ const (
 )
 
 var defaultOptions = &ClientOptions{
-	RestHost:             RestHost,
-	FallbackHosts:        DefaultFallbackHosts(),
-	HTTPMaxRetryCount:    3,
-	RealtimeHost:         "realtime.ably.io",
-	TimeoutConnect:       15 * time.Second,
-	TimeoutDisconnect:    30 * time.Second,
-	TimeoutSuspended:     2 * time.Minute,
-	FallbackRetryTimeout: 10 * time.Minute,
+	RestHost:                 RestHost,
+	FallbackHosts:            DefaultFallbackHosts(),
+	HTTPMaxRetryCount:        3,
+	RealtimeHost:             "realtime.ably.io",
+	TimeoutConnect:           15 * time.Second,
+	TimeoutDisconnect:        30 * time.Second,
+	TimeoutSuspended:         2 * time.Minute,
+	FallbackRetryTimeout:     10 * time.Minute,
 	IdempotentRestPublishing: false,
 }
 
@@ -166,8 +166,18 @@ func (opts *AuthOptions) KeySecret() string {
 	return ""
 }
 
+// ClientOptions configures a RestClient or RealtimeClient. Use NewClientOptions
+// to get sensible defaults.
 type ClientOptions struct {
 	AuthOptions
+
+	// In realtime connections, whether publishes are sent back to the
+	// publishing connection.
+	EchoMessages bool
+
+	// In realtime connections, whether the instance should connect right after
+	// being created, without an explicit call to the Connect method.
+	AutoConnect bool
 
 	RestHost                string // optional; overwrite endpoint hostname for REST client
 	FallbackHostsUseDefault bool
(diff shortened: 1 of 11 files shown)


ably/options.go
```diff
@@ -224,11 +232,10 @@ type ClientOptions struct {
 	Trace *httptrace.ClientTrace
 }
 
-func NewClientOptions(key string) *ClientOptions {
+func NewClientOptions() *ClientOptions {
```
QuintinWillison, line 235: Given my lack of Go knowledge this is just an observation, but is the **only** way to enforce the specified default value for these two members via this constructor? Reading [this answer](https://stackoverflow.com/a/37521991) it seems to me that we would need quite a big refactor to control and encapsulate this better. Is that worth doing? Or is this just an unavoidable compromise in this language?

tcard: @QuintinWillison Yes, this is basically the best we can do without getting too weird. It doesn't matter that much in this case since instantiating the library is supposed to run a number of validations on the passed ClientOptions anyway.

paddybyers:
> A new major version is required due to breaking changes. The ".2" refers to the spec minor version it's supposed to implement: 2.2

I'm not sure I agree with this. It's implementing the 1.2 spec. Are we going to deviate from that for the library version?

ably/options.go
lmars: Another convention is not having an exposed `Options` struct but have it be internal and use functional options as explained [here](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) (which gives the library full control over things like default values since the caller can't just do something like `NewClient(&ClientOptions{})` rather than using the constructor).

paddybyers: I would support this, but it does go against our preference for similarity across the APIs on different platforms.

tcard:
> I'm not sure I agree with this. It's implementing the 1.2 spec. Are we going to deviate from that for the library version?

I think it's a smaller price to pay compared to breaking semver. That could cause real problems to users, as tooling assumes minor versions are backwards-compatible, so asking for 1.1 and getting 1.2 is expected to work.

ably/options.go
tcard: Something akin to functional options do seem a good idea here. I think it's not so much a deviation from the spec as it's a particular way to implement it.

How about this?

```go
type ClientOptions []ClientOption

type ClientOption func(*clientOptions)

func (o ClientOptions) Key(key string) ClientOptions {
    return append(o, func(o *clientOptions) {
        o.Key = key
    })
}

options := ably.ClientOptions{}.
    Key("...").
    AuthCallback(auth).
    AutoConnect(false)

ably.NewRealtime(options...)

// And we could maybe provide top-level shortcuts to match the constructor overloads defined in the spec:   
ably.NewRealtime(ably.Key("..."))
ably.NewRealtime(ably.Token("..."))
```

I wouldn't say this deviates from the spec, provided the translation from the abstract IDL to Go is consistent with naming and types.

mattheworiordan:
> I think it's a smaller price to pay compared to breaking semver. 

But the realtime part of this library is at 0.8.

Having inconsistency in one library is a far bigger problem. 

Having to communicate a breaking change to a small set of users is easy.  Changing all of our libraries, docuemntation, protocol specs, is far harder.

tcard:
OK, base branch renamed to `v1.2`.

ably/options.go
paddybyers: I support doing this ^. I know it's a breaking change, but it's a better API. Just allowing clients to create the zero value `ClientOptions{}` is clearly not going to work, and if you're going to force the use of a `NewClientOptions()` factory method, you might as well do it properly.

lmars: @tcard I would probably only export the option function type and the option methods to lower the surface area of the API, for example:

```go
type clientOptions struct {
	key string
}

type ClientOption func(*clientOptions)

func Key(key string) ClientOption {
	return func(opts *clientOptions) {
		opts.key = key
	}
}

func NewClient(opts ...ClientOption) *Client {
	options := newClientOptions(opts...)
}

client := ably.NewClient(ably.Key("..."))
```

tcard:
Closing, as we'll do ClientOptions as [functional options](https://github.com/ably/ably-go/pull/141#discussion_r401953562) instead.

ably/options.go
tcard: @lmars I'd swear I responded to this at the time, but apparently I didn't (or something ate it).

The problem with that approach is that we have _a lot_ of options, so putting them all in the same top-level namespace is problematic for discoverability and potential name conflicts.

My proposal above has been implemented in #145 and #146 (not that we can't still change it).

tcard: Ah, I see now that I replied [in the pull request](https://github.com/ably/ably-go/pull/145#issuecomment-610958179).

PR #146, opened 2020-04-09
v1.2 ChannelOptions

tcard:
Using functional options in a similar way to ClientOptions.

I decided to also expose `ChannelOption` so that we can pass `...ChannelOption` to `Get`. This way, we can do:

```go
channel.Get("foo")
```

Instead of:

```go
channel.Get("foo", ably.ChannelOptions{})
```

(This isn't useful for `ClientOptions`, as you always must pass a non-empty `ClientOptions`.)

Full usage looks like this:

```go
channel.Get("foo", ably.ChannelOptions{}.WithCipherKey(myKey))
```

The spec's [`Crypto::getDefaultParams` (RSE1)](https://docs.ably.io/client-lib-development-guide/features/#RSE1) isn't provided as it's redundant with the `WithCipherKey` option, which already sets defaults. Also, [Ruby-ish hash-merging of params (TB2b2)](https://docs.ably.io/client-lib-development-guide/features/#TB2b2) isn't implemented; you have to provide a full params struct to `WithCipher`.

### DIFF
--- a/ably/proto/crypto.go
+++ b/ably/proto/crypto.go
@@ -45,6 +45,7 @@ func (c CipherMode) String() string {
 const (
 	DefaultKeyLength       = 256
 	DefaultCipherAlgorithm = AES
+	DefaultCipherMode      = CBC
 )
 
 // CipherParams  provides parameters for configuring encryption  for channels.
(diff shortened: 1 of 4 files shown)


tcard:
Merging in spite of the failing build to avoid conflicts with other work. Might have to fix the tests later.

PR #147, opened 2020-04-09
A few v1.2 easy API fixes

tcard:
Just some straightforward wrapping of the existing API. Context parameters added to a few REST methods.


tcard:
Closing because we changed the approach to get to v1.2 —no more wrapping, just fix APIs as we add spec-associated tests.

PR #148, opened 2020-04-20
RTN23a: Receive from WebSocket with a timeout.

tcard:
> If a transport does not receive any indication of activity on a
> transport for a period greater than the sum of the maxIdleInterval
> (which will be sent in the connectionDetails of the most recent
> CONNECTED message received on that transport) and the
> realtimeRequestTimeout, that transport should be disconnected.

### DIFF
--- a/ably/realtime_conn_test.go
+++ b/ably/realtime_conn_test.go
@@ -191,6 +191,7 @@ func TestRealtimeConn_ReceiveTimeout(t *testing.T) {
 	}
 
 	// TODO: Should be Disconnected, not Failed
+	// Part of https://ably.atlassian.net/browse/FEA-391
 	if expected, got := ably.StateConnFailed, state.State; expected != got {
 		t.Fatalf("expected %v, got %v", expected, got)
 	}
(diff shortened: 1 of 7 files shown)


ably/realtime_conn_test.go
```diff
@@ -139,3 +140,58 @@ func TestRealtimeConn_AuthError(t *testing.T) {
...
+	}
+
+	leeway := 10 * time.Millisecond
+	select {
+	case state = <-states:
+	case <-time.After(realtimeRequestTimeout + time.Duration(maxIdleInterval)*time.Millisecond + leeway):
+		t.Fatal("didn't receive state change event")
+	}
+
+	// TODO: Should be Disconnected, not Failed
```
paddybyers, line 193: Is there an issue for this? Can we link it here?

tcard:
```suggestion
	// TODO: Should be Disconnected, not Failed
	// Part of https://ably.atlassian.net/browse/FEA-391
```

QuintinWillison:
for future reference: the [internal support discussion](https://ably-real-time.slack.com/archives/CPEP9K69Y/p1587409037209000) that generated this PR for a specific customer.

PR #149, opened 2020-04-21
Don't read from connection on inactive state.

tcard:
After the connection gets to an inactive state (ie. not CONNECTING,
CONNECTED, CLOSING), no further messages are expected. Instead, moving
back to an active state will start the loop again by calling setConn.

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -312,7 +312,9 @@ func (c *Conn) isActive() bool {
 	return c.state.current == StateConnConnecting || c.state.current == StateConnConnected
 }
 
-func (c *Conn) canReceiveMessages() bool {
+func (c *Conn) lockCanReceiveMessages() bool {
+	c.state.Lock()
+	defer c.state.Unlock()
 	return c.state.current == StateConnConnecting || c.state.current == StateConnConnected || c.state.current == StateConnClosing
 }
 
@@ -334,7 +336,7 @@ func (c *Conn) logger() *LoggerOptions {
 func (c *Conn) eventloop() {
 	var receiveTimeout time.Duration
 
-	for c.canReceiveMessages() {
+	for c.lockCanReceiveMessages() {
 		var deadline time.Time
 		if receiveTimeout != 0 {
 			deadline = time.Now().Add(receiveTimeout) // RTN23a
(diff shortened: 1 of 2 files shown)


ably/realtime_conn.go
```diff
@@ -312,6 +312,12 @@ func (c *Conn) isActive() bool {
 	return c.state.current == StateConnConnecting || c.state.current == StateConnConnected
 }
 
+func (c *Conn) lockCanReceiveMessages() bool {
+	c.state.Lock()
+	defer c.state.Unlock()
+	return c.state.current == StateConnConnecting || c.state.current == StateConnConnected || c.state.current == StateConnClosing
```
gernest, line 318: This is  duplicate of `(*Conn) lockIsActive` which calls this
```
func (c *Conn) isActive() bool {
	return c.state.current == StateConnConnecting || c.state.current == StateConnConnected
}
```

So probably adding `|| c.state.current == StateConnClosing` to `(*Conn) isActive` and then calling `lockIsActive()` makes more sense in this situation.

From your pr message you can receive message when in active state so probably `lockIsActive()` alone will do?

ably/realtime_conn.go
```diff
@@ -330,7 +336,7 @@ func (c *Conn) logger() *LoggerOptions {
 func (c *Conn) eventloop() {
 	var receiveTimeout time.Duration
 
-	for {
+	for c.lockCanReceiveMessages() {
```
gernest, line 339: I think `c.lockIsActive()` makes more sense here.

ably/realtime_conn.go
tcard: `isActive` determines whether you can do "active" things (ie. send messages), which you can't if you're closing. But you still can receive messages while closing, hence the different condition.

ably/realtime_conn.go
tcard: See https://github.com/ably/ably-go/pull/149#discussion_r412226818.

ably/realtime_conn.go
gernest: @tcard so  you are describing `c. isActive()||c.state.current == StateConnClosing` ?

tcard: Yes, that would be equivalent.

tcard:
@gernest Thanks! I'm not sure it does, but anyway, I'd expect this to be seriously refactored as we implement the rest of the recovery logic.

PR #152, opened 2020-04-21
RTN15a: Reconnect after networking error.

tcard:
This implements RTN15a in a kind of ad-hoc way, pending the rest
of the recovery logic.

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -3,6 +3,8 @@ package ably
 import (
 	"errors"
 	"fmt"
+	"io"
+	"net"
 	"net/url"
 	"strconv"
 	"time"
@@ -80,6 +82,18 @@ var connectResultStates = []StateEnum{
 }
 
 func (c *Conn) connect(result bool) (Result, error) {
+	return c.connectWithRecovery(result, "", 0)
+}
+
+func (c *Conn) reconnect(result bool) (Result, error) {
+	c.state.Lock()
+	connKey := c.details.ConnectionKey
+	connSerial := c.serial
+	c.state.Unlock()
+	return c.connectWithRecovery(result, connKey, connSerial)
+}
+
+func (c *Conn) connectWithRecovery(result bool, connKey string, connSerial int64) (Result, error) {
 	c.state.Lock()
 	defer c.state.Unlock()
 	if c.isActive() {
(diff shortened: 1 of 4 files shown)


tcard:
@paddybyers Now that I think of it, we should do that on any error, right? Application-level errors aren't errors here, but protocol messages with an ERROR action. Any other error here should be recoverable.

ably/realtime_conn.go
```diff
@@ -118,6 +130,10 @@ func (c *Conn) connect(result bool) (Result, error) {
 	if err := c.auth.authQuery(query); err != nil {
 		return nil, c.setState(StateConnFailed, err)
 	}
+	if connKey != "" {
+		query.Set("resume", connKey)
+		query.Set("connectionSerial", fmt.Sprint(connSerial))
```
gernest, line 135: Can you add comment here to mention that these are actually for `(RTN15b1)` and `(RTN15b2)`

Also we need tests for these, ensuring they are properly set.

Do you think maybe its a good idea to mention the spec id in tests so we can easily track down which tests have been covered.

paddybyers:
> Now that I think of it, we should do that on any error, right? Application-level errors aren't errors here, but protocol messages with an ERROR action. Any other error here should be recoverable.

Any error that's not an `ErrorInfo` is implicitly a 500-style error and retriable.

Any `ErrorInfo` with a 5xx statusCode is retriable.

Some 4xx errors (specifically token errors) can also be retried, but after renewing a token.

ably/realtime_conn.go
tcard: @gernest I actually meant to focus on [RTN15a](https://docs.ably.io/client-lib-development-guide/features/#RTN15a), which doesn't specify _how_ resume happens, just that it happens.

But anyway, it's true that the test wasn't really testing that the connection was recovered, just that there was a reconnection. 3404051 adds such to the test.

At 513b37f I've moved the test to a new file, where spec-oriented tests should be put. I did this already at #144, and I really intended this to be done for new API tests only, but since that's not happening yet, I think we should just do it now.

gernest: 👍

QuintinWillison:
future breadcrumbs: [the internal conversation](https://ably-real-time.slack.com/archives/CPEP9K69Y/p1587481283213000) on this, for a specific customer, and their enthusiastic confirmation that this fixed the issue they were facing. Nice work @tcard! 👏

Issue #154, opened 2020-04-22
Lib failing to retrying on 5xx if it can't parse the body


SimonWoolf:
[Customer using ably-go 1.1.2](https://support.ably.io/a/tickets/106111) reporting 400 errors. On investigation this was due to cloudfront returning rendered HTML for 502 errors; ably-go is then unable to parse the body, and turns it into a 400 error.

But this is incorrect - the library should be retrying for any response with an http status code 500-504 (specific spec violation: https://docs.ably.io/client-lib-development-guide/features/#RSC15d), doesn't matter what the body is. It still has the HTTP status code, and can decide to retry based on that.

Possibly introduced by https://github.com/ably/ably-go/pull/129/files (though that may have been fixing even-more-broken older behaviour)

SimonWoolf:
Thanks for the quick response @gernest. Poke me or quintin when there's a release with this fix in it and we'll let the customer know

PR #156, opened 2020-04-23
Use ClientOptions.TLSPort for tls connections

gernest:


tcard:
Could you provide a test for this? Ideally one that map to the relevant spec item.

gernest:
@tcard I only came to this accidentally from the diff of PR #129  mentioned on #154 it looks like it slipped past unnoticed. I don't know which part of the spec @paddybyers was addressing I think It slipped past review by accident.

tcard:
@gernest The PR looks good to me, so feel free to merge. I just think that as a general practice it's nice that fixes come with tests to avoid regressions, even if they're unlikely.

PR #157, opened 2020-04-23
Properly set  Error.StatusCode

### LINKED ISSUE #154: Lib failing to retrying on 5xx if it can't parse the body

gernest:
fixes #154

newError was completely ignoring  response status code.
It was wrongly deriving it from error code.

This patch ensures Error,StatusCode is set with correct response status.

### DIFF
--- a/ably/error_test.go
+++ b/ably/error_test.go
@@ -1,19 +1,14 @@
 package ably_test
 
 import (
-	"crypto/tls"
-	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"strconv"
-	"sync/atomic"
 	"testing"
-	"time"
 
 	"github.com/ably/ably-go/ably"
-	"github.com/ably/ably-go/ably/ablytest"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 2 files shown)


ably/error_test.go
```diff
@@ -54,3 +59,83 @@ func TestIssue127ErrorResponse(t *testing.T) {
 	assert.NotNil(t, err)
 	assert.Contains(t, err.Error(), errMsg)
 }
+
+func TestIssue_154(t *testing.T) {
```
tcard, line 63: This tests a number of things that have nothing to do with #154. We just want to check that a HTTP response with 400 status code turns into an ErrorInfo with 400 status code.

There are other tests that test retries and fallbacks.

gernest:
@tcard  PTAL  I added a simpler test where I only check to see the StatusCode is properly set in Error on not json/msgpack mime types.

PR #159, opened 2020-04-27
add implementation for  RTN15b-c spec

gernest:
Covers

- [x] [(RTN15b) ](https://docs.ably.io/client-lib-development-guide/features/#RTN15b)
  - [x] [(RTN15b1) ](https://docs.ably.io/client-lib-development-guide/features/#RTN15b1)
  - [x] [(RTN15b2) ](https://docs.ably.io/client-lib-development-guide/features/#RTN15b2)
- [ ] [(RTN15c) ]()
  - [x] [(RTN15c1) ](https://docs.ably.io/client-lib-development-guide/features/#RTN15c1)
  - [x] [(RTN15c2) ](https://docs.ably.io/client-lib-development-guide/features/#RTN15c2)
  - [x] [(RTN15c3) ]()
  - [x] [(RTN15c4) ]()
  - [ ] [(RTN15c5) ]()

### DIFF
--- a/ably/realtime_client.go
+++ b/ably/realtime_client.go
@@ -79,12 +79,12 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 
 	case proto.ActionError:
 		// (RTN15c4)
+
 		for _, ch := range c.Channels.All() {
-			ch.state.Lock()
-			if ch.state.current == StateChanAttached {
-				ch.state.set(StateChanFailed, msg.Error)
+			if ch.State() == StateChanAttached {
+				ch.state.syncSet(StateChanFailed, msg.Error)
 			}
-			ch.state.Unlock()
+
 		}
 	}
 }
--- a/ably/realtime_client.go
+++ b/ably/realtime_client.go
@@ -1,6 +1,7 @@
 package ably
 
 import (
+	"net/http"
 	"sync"
 	"time"
 
@@ -71,26 +72,8 @@ func (c *RealtimeClient) onChannelMsg(msg *proto.ProtocolMessage) {
 func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 	switch msg.Action {
 	case proto.ActionConnected:
-		if c.Connection.id == msg.ConnectionID {
-			c.Connection.queue.Flush()
-			if msg.Error != nil {
-				// (RTN15c2)
-				c.Connection.state.Lock()
-				c.Connection.setState(StateConnConnected, msg.Error)
-				c.Connection.state.Unlock()
-				return
-			}
-			// (RTN15c1)
-			return
-		}
 		if msg.Error != nil {
 			// (RTN15c3)
-			c.Connection.state.Lock()
-			c.Connection.setState(StateConnConnected, msg.Error)
-			//TODO (gernest): add Conn.ErrorReason
-			c.Connection.msgSerial = 0
-			c.Connection.state.Unlock()
-
 			for _, ch := range c.Channels.All() {
 				ch.state.Lock()
 				current := ch.state.current
@@ -102,13 +85,20 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 				}
 			}
 		}
-	default:
-		// (RTN15c5)
+	case proto.ActionError:
 		// (RTN15c4)
-		c.Connection.state.Lock()
-		c.Connection.setState(StateConnFailed, msg.Error)
-		c.Connection.state.Unlock()
+		for _, ch := range c.Channels.All() {
+			ch.state.Lock()
+			if ch.state.current == StateChanAttached {
+				ch.state.set(StateChanFailed, msg.Error)
+			}
+			ch.state.Unlock()
+		}
 	}
+}
+
+func tokenError(err *proto.ErrorInfo) bool {
+	return err.StatusCode == http.StatusUnauthorized && (40140 <= err.Code && err.Code < 40150)
 }
 
 func (c *RealtimeClient) onConnStateChange(state State) {
--- a/ably/realtime_client.go
+++ b/ably/realtime_client.go
@@ -72,16 +72,15 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 	switch msg.Action {
 	case proto.ActionConnected:
 		if c.Connection.id == msg.ConnectionID {
+			c.Connection.queue.Flush()
 			if msg.Error != nil {
 				// (RTN15c2)
 				c.Connection.state.Lock()
 				c.Connection.setState(StateConnConnected, msg.Error)
 				c.Connection.state.Unlock()
-				c.Connection.queue.Flush()
 				return
 			}
 			// (RTN15c1)
-			c.Connection.queue.Flush()
 			return
 		}
 		if msg.Error != nil {
--- a/ably/realtime_client.go
+++ b/ably/realtime_client.go
@@ -37,7 +37,9 @@ func NewRealtimeClient(opts *ClientOptions) (*RealtimeClient, error) {
 	c.rest = rest
 	c.Auth = rest.Auth
 	c.Channels = newChannels(c)
-	conn, err := newConn(c.opts(), rest.Auth, c.onChannelMsg, c.onReconnectMsg, c.onConnStateChange)
+	conn, err := newConn(c.opts(), rest.Auth, connCallbacks{
+		c.onChannelMsg, c.onReconnectMsg, c.onConnStateChange,
+	})
 	if err != nil {
 		return nil, err
 	}
@@ -72,10 +74,6 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 		if c.Connection.id == msg.ConnectionID {
 			if msg.Error != nil {
 				// (RTN15c2)
-
-				// I'm a bit confused here as the spec says to set CONNECTED event but
-				// there is no event mechanism only state changes and we are not supposed to
-				// change the state here.
 				c.Connection.state.Lock()
 				c.Connection.setState(StateConnConnected, msg.Error)
 				c.Connection.state.Unlock()
--- a/ably/realtime_client.go
+++ b/ably/realtime_client.go
@@ -72,37 +72,42 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 		if c.Connection.id == msg.ConnectionID {
 			if msg.Error != nil {
 				// (RTN15c2)
+
+				// I'm a bit confused here as the spec says to set CONNECTED event but
+				// there is no event mechanism only state changes and we are not supposed to
+				// change the state here.
 				c.Connection.state.Lock()
 				c.Connection.setState(StateConnConnected, msg.Error)
-
-				// According to the spec we need to set Connection#errorReason but Conn
-				// doesn't have errorReason field.However it has Reason method that
-				// returns the last know error which in our case will be be msg.Error so
-				// it will be conforming to spec.
-				//
-				// Adding Conn.errorReason won't work in our case as it will be private
-				// according to Go conventions. So, I'm choosing to only set state with
-				// appropriate reason and assuming users will use Conn.Reason to access
-				// the error.
-				//
-				// TODO: Add Conn.ErrorReason  field.
 				c.Connection.state.Unlock()
+				c.Connection.queue.Flush()
 				return
 			}
 			// (RTN15c1)
+			c.Connection.queue.Flush()
 			return
 		}
 		if msg.Error != nil {
 			// (RTN15c3)
-			// process all queued messages. No state changes here.
-			c.Channels.client.Connection.queue.Flush()
-			return
+			for _, ch := range c.Channels.All() {
+				ch.state.Lock()
+				current := ch.state.current
+				ch.state.Unlock()
+				switch current {
+				case StateConnSuspended, StateChanAttaching, StateChanAttached:
+					// The spec doesn't say we should wait for response.
+					ch.attach(false)
+				}
+			}
+			c.Connection.state.Lock()
+			c.Connection.msgSerial = 0
+			c.Connection.state.Unlock()
 		}
-	case proto.ActionError:
-		// (RTN15c5)
-		// (RTN15c4) ?
 	default:
-		// We have received unexpected message here. We are in failure state
+		// (RTN15c5)
+		// (RTN15c4)
+		c.Connection.state.Lock()
+		c.Connection.setState(StateConnFailed, msg.Error)
+		c.Connection.state.Unlock()
 	}
 }
 
--- a/ably/realtime_client.go
+++ b/ably/realtime_client.go
@@ -37,7 +37,7 @@ func NewRealtimeClient(opts *ClientOptions) (*RealtimeClient, error) {
 	c.rest = rest
 	c.Auth = rest.Auth
 	c.Channels = newChannels(c)
-	conn, err := newConn(c.opts(), rest.Auth, c.onChannelMsg, c.onConnStateChange)
+	conn, err := newConn(c.opts(), rest.Auth, c.onChannelMsg, c.onReconnectMsg, c.onConnStateChange)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


ably/realtime_conn.go
```diff
@@ -32,6 +32,10 @@ type Conn struct {
 	auth          *Auth
 	onChannelMsg  func(*proto.ProtocolMessage)
 	onStateChange func(State)
+	// This tracks if we have issued reconnection quest. If we receive any message
```
tcard, line 35:
```suggestion
	// reconnecting tracks if we have issued a reconnection request. If we receive any message
```

ably/realtime_conn.go
```diff
@@ -90,6 +94,9 @@ func (c *Conn) reconnect(result bool) (Result, error) {
 	connKey := c.details.ConnectionKey
 	connSerial := c.serial
 	c.state.Unlock()
+	defer func() {
```
tcard, line 97: Why the defer? Please add a comment if there's a reason for it, otherwise remove it.

Plus, it's not true we're reconnecting if `c.connectWithRecovery` returns an error.

tcard: Done at ca27f152968950cd4bb5ecab15a9b33aea75d5c3.

ably/realtime_conn.go
```diff
@@ -372,6 +379,30 @@ func (c *Conn) eventloop() {
...
+					continue
+				}
+				if msg.Error != nil {
+					// (RTN15c3)
+					continue
+				}
+			case proto.ActionError:
+				// (RTN15c5)
+				// (RTN15c4) ?
+			}
```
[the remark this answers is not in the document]
tcard: Done at 432d365205848fd9c20d3102e663777141f45cdb.

ably/realtime_conn.go
```diff
@@ -393,6 +393,21 @@ func (c *Conn) eventloop() {
...
+						// doesn't have errorReason field.However it has Reason method that
+						// returns the last know error which in our case will be be msg.Error so
+						// it it be conforming to spec.
+						//
+						// Adding Conn.errorReason won't work in our case as it will be private
+						// according to Go conventions. So, I'm choosing to only set state with
+						// appropriate reason and assuming users will use Conn.Reason to access
+						// the error.
+						//
+						// TODO: Add Conn.ErrorReason  field.
```
tcard: This is the kind of API change that should go into the 1.2 branch.

tcard: Maybe a sensible thing to do here is:

- Get this change into master.
- Merge master onto the 1.2 branch.
- Implement this TODO as a PR on top of that.

ably/realtime_conn.go
```diff
@@ -18,27 +18,32 @@ var (
...
+	err          error
+	conn         proto.Conn
+	opts         *ClientOptions
+	state        *stateEmitter
+	stateCh      chan State
+	pending      pendingEmitter
+	queue        *msgQueue
+	auth         *Auth
+	onChannelMsg func(*proto.ProtocolMessage)
+	// This callback is called when we get a response from reconnect request. We
```
tcard, line 35: Totally unimportant, but I'd prefer for "doc comments"'s first word to be the name of the documented thing, [as established](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences), even private ones. Please indulge me :)

```suggestion
	// onReconnectMsg is called when we get a response from reconnect request. We
```

ably/realtime_conn.go
```diff
@@ -18,27 +18,32 @@ var (
...
+	onReconnectMsg func(*proto.ProtocolMessage)
+	onStateChange  func(State)
 	// reconnecting tracks if we have issued a reconnection request. If we receive any message
 	// with this set to true then its the first message/response after issuing the
 	// reconnection request.
 	reconnecting bool
 }
 
-func newConn(opts *ClientOptions, auth *Auth, onChannelMsg func(*proto.ProtocolMessage), onStateChange func(State)) (*Conn, error) {
+func newConn(opts *ClientOptions, auth *Auth, onChannelMsg, onReconnectMsg func(*proto.ProtocolMessage), onStateChange func(State)) (*Conn, error) {
```
tcard, line 54: Suggestion: when we're at three callbacks, maybe it's time to take a single "delegate" struct-of-callbacks instead?

ably/realtime_client.go
```diff
@@ -94,6 +94,8 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 		}
 		if msg.Error != nil {
 			// (RTN15c3)
+			// process all queued messages. No state changes here.
```
tcard: I interpret [RTN15c3](RTN15c3) as also moving to CONNECTED; the item doesn't really say otherwise, and it's arguably implied by the "CONNECTED event" it talks about. Since it doesn't specify any other transition, and also because it makes sense (we're definitely CONNECTED, although on a different connection), the default transition to CONNECTED should apply here.

Plus, this is missing a lot from RTN15c3. Just commenting this to avoid losing track of that, since there's no placeholder comment for all that.

ably/realtime_client.go
```diff
@@ -72,37 +72,42 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 		if c.Connection.id == msg.ConnectionID {
 			if msg.Error != nil {
 				// (RTN15c2)
+
+				// I'm a bit confused here as the spec says to set CONNECTED event but
+				// there is no event mechanism only state changes and we are not supposed to
+				// change the state here.
```
tcard, line 78: But we are; the `setState` below will transition from CONNECTING to CONNECTED.

ably/realtime_client.go
```diff
@@ -72,37 +72,42 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 		if c.Connection.id == msg.ConnectionID {
 			if msg.Error != nil {
 				// (RTN15c2)
+
+				// I'm a bit confused here as the spec says to set CONNECTED event but
```
tcard, line 76: The events thing is being revamped for the spec-compliant 1.2 release. See #144.

ably/realtime_client.go
```diff
@@ -72,37 +72,42 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
...
-				// appropriate reason and assuming users will use Conn.Reason to access
-				// the error.
-				//
-				// TODO: Add Conn.ErrorReason  field.
 				c.Connection.state.Unlock()
+				c.Connection.queue.Flush()
 				return
 			}
 			// (RTN15c1)
+			c.Connection.queue.Flush()
```
tcard, line 86: Since this is also what happens on a normal transition to CONNECTED, and there's also some other common logic between first-time connection and reconnection, I'm worried we're duplicating too much. Maybe we should refactor so that reconnection includes that common logic?

ably/realtime_conn.go
```diff
@@ -389,15 +397,16 @@ func (c *Conn) eventloop() {
...
 			// We have already issued the reconnecting request. So we are in the
 			// (RTN15c)  territory now.
-			c.reconnecting = false
+			c.reconnecting.Store(false)
 			if c.onReconnectMsg != nil {
 				c.onReconnectMsg(msg)
 			}
 			return
 		}
+		c.state.Unlock()
```
tcard: Why this unlock?

ably/realtime_conn.go
```diff
@@ -40,7 +41,14 @@ type Conn struct {
 	// reconnecting tracks if we have issued a reconnection request. If we receive any message
 	// with this set to true then its the first message/response after issuing the
 	// reconnection request.
-	reconnecting bool
+	reconnecting atomic.Value
```
tcard, line 44: I'm not sure we should introduce another lock when we already have one to cover the data accessed by both the connection event loop and calls from the outside, namely `c.state`. This is the lock that we hold on `eventloop` below for data we access when when a message arrives, which is also the case here.

ably/realtime_conn.go
```diff
@@ -372,6 +379,30 @@ func (c *Conn) eventloop() {
...
+					// (RTN15c3)
+					continue
+				}
+			case proto.ActionError:
+				// (RTN15c5)
+				// (RTN15c4) ?
+			}
+
+		}
 		if msg.ConnectionSerial != 0 {
```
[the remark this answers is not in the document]
tcard: @gernest This seems to be pending.

ably/realtime_conn.go
tcard: Done at f575c81464759aa2fa48e3221b5845eb8acc3ce6.

ably/realtime_conn.go
```diff
@@ -403,7 +404,9 @@ func (c *Conn) eventloop() {
 		if c.isReconnecting() {
 			// We have already issued the reconnecting request. So we are in the
 			// (RTN15c)  territory now.
-			c.reconnecting.Store(false)
+			c.state.Lock()
+			c.reconnecting = true
```
tcard, line 408:
```suggestion
			c.reconnecting = false
```

ably/realtime_conn.go
tcard: Done at 909ef4b07af24f08e2151b7866e6d7342af9a948. But really, we should revisit locking here, we're unlocking and locking several times in the same sequence which logically is atomic.

ably/realtime_client.go
```diff
@@ -72,16 +72,15 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
 	switch msg.Action {
 	case proto.ActionConnected:
 		if c.Connection.id == msg.ConnectionID {
+			c.Connection.queue.Flush()
```
tcard, line 75: Sorry, this is wrong now, we should only start processing the queue when the state is CONNECTED. It was right in that sense before.

If you look at [realtime_conn.go](https://github.com/ably/ably-go/blob/feature/RTN15b-c/ably/realtime_conn.go#L440-L457), when CONNECTED arrives while not reconnecting, there's a lot of logic that also must apply when we _are_ reconnecting. Some of that logic is replicated here.

I'd suggest that that block of code should be reached whether we're reconnecting or not. And then, only if we're reconnecting, we call `onReconnect(err)` in order to reattach channels and _only_ to reattach channels. The rest of the logic is common.

gernest: That is very nice observation.

gernest: I moved as much logic to `Conn.eventLoop` only changes touching channels remain on the call back https://github.com/ably/ably-go/pull/159/commits/9ac3c19fa787933298376d891ebb9796e97d60d8

ably/realtime_conn_spec_test.go
```diff
@@ -121,8 +123,147 @@ func (c protoConnWithFakeEOF) Receive(deadline time.Time) (*proto.ProtocolMessag
...
+	}
+
+	// Publish a message to the channel through REST. If connection recovery
+	// succeeds, we should then receive it without reattaching.
+
+	rest, err := ably.NewRestClient(app.Options())
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = rest.Channels.Get("channel", nil).Publish("name", "data")
```
tcard, line 195: There's a race here between this publish and the reconnection. It could happen that, before this publish, the connection already reconnected, in which case we wouldn't know if we recovered the connection state or not.

I'd suggest putting a channel send/rececive in the `Dial` function above and do this:

- Wait for Dial to be called; let it blocked there.
- Do the publish.
- Let Dial to finish connection.

Kind of like:

```go
gotDial := make(chan chan struct{})

Dial: func ... {
    ...
    goOn := make(chan struct{})
    gotDial <- goOn
    <-goOn
   ....
}

goOn := <-gotDial
rest.Publish(...)
close(goOn)
```

ably/realtime_conn_spec_test.go
```diff
@@ -562,3 +563,95 @@ func TestRealtimeConn_RTN15c3(t *testing.T) {
...
+		t.Fatalf("expected transition to %v, got %v", expected, got)
+	}
+	err = client.Connection.Reason()
+	if err == nil {
+		t.Fatal("expected reason to be set")
+	}
+	reason := err.(*ably.Error)
+	if reason.StatusCode != errInfo.StatusCode {
+		t.Errorf("expected %d got %d", errInfo.StatusCode, reason.StatusCode)
+	}
```
tcard, line 656: Missing this:

> The client should transition to the FAILED state

(Here "client" means the connection.)

ably/realtime_client.go
```diff
@@ -102,15 +85,22 @@ func (c *RealtimeClient) onReconnectMsg(msg *proto.ProtocolMessage) {
...
-	default:
-		// (RTN15c5)
+	case proto.ActionError:
 		// (RTN15c4)
-		c.Connection.state.Lock()
-		c.Connection.setState(StateConnFailed, msg.Error)
-		c.Connection.state.Unlock()
+		for _, ch := range c.Channels.All() {
+			ch.state.Lock()
+			if ch.state.current == StateChanAttached {
```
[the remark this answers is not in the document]
gernest: addressed fae539220e0a4dd3da778c74ffba04860091e4cb

ably/realtime_conn.go
```diff
@@ -452,6 +452,7 @@ func (c *Conn) eventloop() {
 					c.callbacks.onReconnectMsg(msg)
 				}
 			} else {
+				// preserve old bahavior.
```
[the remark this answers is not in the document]
gernest: addressed 445ac96727c0479eb112afa1d674370b42fcfa7a

ably/realtime_conn_spec_test.go
gernest: addressed e0b5de36d73a7913fea09ab1e190ce32838f3f9b

ably/realtime_conn_spec_test.go
```diff
@@ -267,3 +267,80 @@ func recent(msgs []*proto.ProtocolMessage, action proto.Action) *proto.ProtocolM
...
+		t.Fatal("didn't transition on EOF")
+	}
+	if expected, got := ably.StateConnDisconnected, state; expected != got.State {
+		t.Fatalf("expected transition to %v, got %v", expected, got)
+	}
+	rest, err := ably.NewRestClient(app.Options())
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = rest.Channels.Get("channel", nil).Publish("name", "data")
```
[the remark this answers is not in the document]
gernest: addressed e0b5de36d73a7913fea09ab1e190ce32838f3f9b

ably/realtime_conn_spec_test.go
```diff
@@ -468,3 +468,97 @@ func TestRealtimeConn_RTN15c2(t *testing.T) {
...
+	case state = <-stateChanges:
+	case <-time.After(50 * time.Millisecond):
+		t.Fatal("didn't reconnect")
+	}
+	if expected, got := ably.StateConnConnecting, state; expected != got.State {
+		t.Fatalf("expected transition to %v, got %v", expected, got)
+	}
+	select {
+	case state = <-chanStateChanges:
+	case <-time.After(80 * time.Millisecond):
```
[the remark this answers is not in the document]
gernest: addressed cd26ac4b930f0c576b01da120a39064b8765f668

I was getting channel state changes too late, this was an attempt to wait a bit more. I have figured out a way to properly wait for channel state changes now.

ably/realtime_conn_spec_test.go
```diff
@@ -340,3 +340,103 @@ func TestRealtimeConn_RTN15c1(t *testing.T) {
...
+		t.Fatal("didn't reconnect")
+	}
+	if expected, got := ably.StateConnConnecting, state; expected != got.State {
+		t.Fatalf("expected transition to %v, got %v", expected, got)
+	}
+	select {
+	case state = <-chanStateChanges:
+		t.Fatal("expected no channel state changes")
+	case <-time.After(50 * time.Millisecond):
+		// (RTN15c1)
```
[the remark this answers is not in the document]
gernest: addressed 1069fa893f143aaa46e7b3edcb384deaf3f95ef3

ably/realtime_conn_spec_test.go
gernest: addressed 402872e3efa6fcf4b3eea62a349a58ffe1932fa5

ably/realtime_conn_spec_test.go
```diff
@@ -468,3 +468,97 @@ func TestRealtimeConn_RTN15c2(t *testing.T) {
...
+	if expected, got := ably.StateConnConnecting, state; expected != got.State {
+		t.Fatalf("expected transition to %v, got %v", expected, got)
+	}
+	select {
+	case state = <-chanStateChanges:
+	case <-time.After(80 * time.Millisecond):
+		t.Fatal("didn't change state")
+	}
+	// we are testing to make sure we have initiated a new attach for channels
+	// ATTACHING or ATTACHED state.
```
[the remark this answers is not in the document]
gernest: addressed ee38fc368afab5cb2425d1bd6fe95415952f2af5

ably/realtime_conn_spec_test.go
```diff
@@ -144,12 +144,17 @@ func TestRealtimeConn_RTN15b(t *testing.T) {
...
 
 	var metaList []*meta
-
+	gotDial := make(chan chan struct{})
 	app, client := ablytest.NewRealtimeClient(&ably.ClientOptions{
 		NoConnect: true,
 		Dial: func(protocol string, u *url.URL) (proto.Conn, error) {
 			m := &meta{dial: u}
 			metaList = append(metaList, m)
+			if len(metaList) > 1 {
```
tcard, line 153: What's the purpose of this condition? Also, won't it be always true, given we're just appending to `metaList` in the line just above this one?

ably/realtime_conn_spec_test.go
```diff
@@ -467,6 +467,7 @@ func TestRealtimeConn_RTN15c2(t *testing.T) {
 		t.Fatal("expected message after connection recovery; got none")
 	}
 
+	<-stateChanges
```
tcard, line 470: If we wait on a channel without a timeout, the test will just hang forever if we ever have a regression.

(You might want to backport [these helpers](https://github.com/ably/ably-go/commit/6abf09a4a86d237e7c006fe761516b6c0ce67a02) I added to the 1.2 branch. So you could do e. g. `ablytest.Soon.Recv(t, nil, stateChanges, t.Fatal)` instead of a big, noisy `select`.)

ably/realtime_conn_spec_test.go
gernest: Dial is called twice, and the first time we don't want to block the channel. We want to block with `goOn` channel when dialling the second time where `len(metaList)==2`

ably/realtime_conn_spec_test.go
gernest: I think since we will be on `1.2` branch after this PR I will address this there in a separate PR to cleanup the tests.

gernest:
The test that is failing it a flaky one due to sandbox timeout. Nothing to do with this PR.

PR #165, opened 2020-06-22
 Add rtn16 

gernest:
- [x] RTN16a
- [x] RTN16b
- [x] RTN16c
- [x] RTN16d
- [x] RTN16e
- [x] RTN16f

### DIFF
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -118,3 +118,9 @@ func (c *Connection) RemoveKey() {
 	defer c.state.Unlock()
 	c.key = ""
 }
+
+func (c *Connection) MsgSerial() int64 {
+	c.state.Lock()
+	defer c.state.Unlock()
+	return c.msgSerial
+}
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -107,12 +107,6 @@ func (c *Connection) OffState(ch chan<- State, states ...StateEnum) {
 	c.offState(ch, states...)
 }
 
-func (c *Connection) RecoveryKey() string {
-	c.state.Lock()
-	defer c.state.Unlock()
-	return c.recoveryKey
-}
-
 func (c *Connection) RemoveKey() {
 	c.state.Lock()
 	defer c.state.Unlock()
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -106,3 +106,15 @@ func (c *Connection) OnState(ch chan<- State, states ...StateEnum) {
 func (c *Connection) OffState(ch chan<- State, states ...StateEnum) {
 	c.offState(ch, states...)
 }
+
+func (c *Connection) RecoveryKey() string {
+	c.state.Lock()
+	defer c.state.Unlock()
+	return c.recoveryKey
+}
+
+func (c *Connection) RemoveKey() {
+	c.state.Lock()
+	defer c.state.Unlock()
+	c.key = ""
+}
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -18,19 +18,20 @@ var (
 // Connection represents a single connection Realtime instantiates for
 // communication with Ably servers.
 type Connection struct {
-	details   proto.ConnectionDetails
-	id        string
-	serial    int64
-	msgSerial int64
-	err       error
-	conn      proto.Conn
-	msgCh     chan *proto.ProtocolMessage
-	opts      *ClientOptions
-	state     *stateEmitter
-	stateCh   chan State
-	pending   pendingEmitter
-	queue     *msgQueue
-	auth      *Auth
+	id          string
+	key         string
+	recoveryKey string
+	serial      int64
+	msgSerial   int64
+	err         error
+	conn        proto.Conn
+	msgCh       chan *proto.ProtocolMessage
+	opts        *ClientOptions
+	state       *stateEmitter
+	stateCh     chan State
+	pending     pendingEmitter
+	queue       *msgQueue
+	auth        *Auth
 
 	callbacks    connCallbacks
 	reconnecting bool
@@ -100,7 +101,7 @@ func (c *Connection) connect(result bool) (Result, error) {
 
 func (c *Connection) reconnect(result bool) (Result, error) {
 	c.state.Lock()
-	connKey := c.details.ConnectionKey
+	connKey := c.key
 	connSerial := c.serial
 	c.state.Unlock()
 	r, err := c.connectWithRecovery(result, connKey, connSerial)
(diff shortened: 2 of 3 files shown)


ably/export_test.go
```diff
@@ -106,3 +106,21 @@ func (c *Connection) OnState(ch chan<- State, states ...StateEnum) {
 func (c *Connection) OffState(ch chan<- State, states ...StateEnum) {
 	c.offState(ch, states...)
 }
+
+func (c *Connection) RecoveryKey() string {
+	c.state.Lock()
+	defer c.state.Unlock()
+	return c.recoveryKey
```
tcard, line 113: This isn't right. Per [RTN16b](https://docs.ably.io/client-lib-development-guide/features/#RTN16b):

> Connection#recoveryKey is an attribute composed of the connectionKey, and the latest connectionSerial received on the connection, and the current msgSerial

But this seems to be returning the `resume` option passed to the library, if any.

ably/export_test.go
```diff
@@ -106,3 +106,21 @@ func (c *Connection) OnState(ch chan<- State, states ...StateEnum) {
...
+}
+
+func (c *Connection) RemoveKey() {
```
tcard, line 116: This at least shouldn't be public, but I don't understand why it exists at all. It doesn't seem to be called from anywhere.

ably/export_test.go
```diff
@@ -106,3 +106,21 @@ func (c *Connection) OnState(ch chan<- State, states ...StateEnum) {
...
+	c.state.Lock()
+	defer c.state.Unlock()
+	c.key = ""
+}
+
+func (c *Connection) MsgSerial() int64 {
```
tcard, line 122: This isn't part of the public API defined in the spec. If you want to expose it for tests, you can put it in `export_test.go`.

ably/realtime_conn.go
```diff
@@ -15,26 +16,40 @@ var (
...
+	msgCh       chan *proto.ProtocolMessage
+	opts        *ClientOptions
+	state       *stateEmitter
+	stateCh     chan State
+	pending     pendingEmitter
+	queue       *msgQueue
+	auth        *Auth
+
+	callbacks connCallbacks
+	mode      ConnectionMode
```
tcard, line 50: I don't think this belongs here. What you call connection mode only affects connection establishment, and can be encapsulated there.

ably/realtime_conn.go
```diff
@@ -151,13 +164,45 @@ func (c *Connection) connectWithRecovery(result bool, connKey string, connSerial
...
 	if err := c.auth.authQuery(query); err != nil {
+		return nil, err
+	}
+	switch mode {
+	case ResumeMode:
+		query.Set("resume", c.key)
+		query.Set("connectionSerial", fmt.Sprint(c.serial))
+	case RecoveryMode:
+		m := strings.Split(c.recoveryKey, ":")
+		if len(m) > 0 {
```
tcard, line 175: We should error out if the recovery key is malformed. It _needs_ to have three parts, separated by `:`.

Also, we're missing this ([RTN16f](https://docs.ably.io/client-lib-development-guide/features/#RTN16f)):

> The msgSerial component of the recoveryKey [...] is used to set the library internal msgSerial. (If the recover fails, the counter should be reset to 0 per RTN15c3 )

ably/realtime_conn.go
```diff
@@ -50,12 +65,13 @@ type connCallbacks struct {
...
 func newConn(opts *ClientOptions, auth *Auth, callbacks connCallbacks) (*Connection, error) {
 	c := &Connection{
-		opts:      opts,
-		msgCh:     make(chan *proto.ProtocolMessage),
-		state:     newStateEmitter(StateConn, StateConnInitialized, "", auth.logger()),
-		pending:   newPendingEmitter(auth.logger()),
-		auth:      auth,
-		callbacks: callbacks,
+		opts:        opts,
+		recoveryKey: opts.Recover,
```
tcard, line 69: Why create a new field when we can get it from `opts`?

ably/realtime_conn.go
```diff
@@ -480,44 +531,50 @@ func (c *Connection) eventloop() {
...
 				c.state.Unlock()
 				if id != msg.ConnectionID {
-					// (RTN15c3)
-					// we are calling this outside of locks to avoid deadlock because in the
-					// RealtimeClient client where this callback is implemented we do some ops
-					// with this Conn where we re acquire Conn.state.Lock again.
-					c.callbacks.onReconnectMsg(msg)
+					// (RTN16a) its up to client to explicitly attach channels when in
+					// RecoveryMode
+					if mode == ResumeMode {
```
tcard, line 560: Note that if we're recovering instead of resuming, we can also call `onReconnectMsg` anyway since there won't be any channels to reattach, so the behavior is the same. Given this, the only bit of logic that diverges here between `ResumeMode` and `RecoveryMode` disappears, so we can just go back to the `reconnecting` boolean flag.

ably/export_test.go
gernest: This is helper in `export_test.go` file they are only used in tests

ably/export_test.go
gernest: it is in `export_test.go` file already

ably/realtime_conn.go
gernest: addressed in 4449023b755bb04b26b63f4dab7c83ccd5dba667

ably/realtime_conn.go
gernest: addressed in 4449023b755bb04b26b63f4dab7c83ccd5dba667

ably/realtime_conn.go
gernest: addressed in 4449023b755bb04b26b63f4dab7c83ccd5dba667

ably/realtime_conn.go
gernest: addressed in 4449023b755bb04b26b63f4dab7c83ccd5dba667

ably/export_test.go
gernest: addressed in 4449023b755bb04b26b63f4dab7c83ccd5dba667

ably/realtime_conn.go
```diff
@@ -15,11 +16,23 @@ var (
 	errQueueing = errors.New("unable to send messages in current state with disabled queueing")
 )
 
+// ConnectionMode is the mode in which the connection is operating
+type ConnectionMode uint
```
tcard, line 20: This and associated constants shouidn't be public.

gernest: addressed in 71dafc5cc4d75fd14c09e66ad2907a30bc2c08ef

gernest:
@tcard PTAL

PR #166, opened 2020-06-30
ProtocolMessage: add the timestamp to each message on receipt

paddybyers:
This adds `Message.Timestamp`


tcard:
@paddybyers Should we merge this?

Also, after merging, could you please to merge `master` into `integration/1.2`?

paddybyers:
> Should we merge this?

This was a quick+dirty change that we needed because we're running load tests with the ably-go client. I think a real PR should be created that has a spec reference, checks for other related things we might need to do in the same area, and has tests.

QuintinWillison:
Timestamp property will ultimately end up being added, in aggregate, under #322.

PR #169, opened 2020-07-09
Add RTN23- heartbeats

gernest:
- [x] RTN23a
- [x] RTN23b

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -70,6 +70,10 @@ func newConn(opts *ClientOptions, auth *Auth, callbacks connCallbacks) (*Connect
 }
 
 func (c *Connection) dial(proto string, u *url.URL) (proto.Conn, error) {
+	// (RTN23b)
+	query := u.Query()
+	query.Add("heartbeats", "true")
+	u.RawQuery = query.Encode()
 	if c.opts.Dial != nil {
 		return c.opts.Dial(proto, u)
 	}
@@ -419,7 +423,6 @@ func (c *Connection) logger() *LoggerOptions {
 
 func (c *Connection) eventloop() {
 	var receiveTimeout time.Duration
-
 	for c.lockCanReceiveMessages() {
 		var deadline time.Time
 		if receiveTimeout != 0 {
--- a/ably/realtime_conn_spec_test.go
+++ b/ably/realtime_conn_spec_test.go
@@ -1081,24 +1081,19 @@ func sameConnection(a, b string) bool {
 
 type wrapProtoMessage struct {
 	proto.Conn
-	ok chan *messageWithDeadline
-}
-
-type messageWithDeadline struct {
-	msg      *proto.ProtocolMessage
-	deadline time.Time
+	ok chan *proto.ProtocolMessage
 }
 
 func (w *wrapProtoMessage) Receive(deadline time.Time) (*proto.ProtocolMessage, error) {
 	msg, err := w.Conn.Receive(deadline)
-	w.ok <- &messageWithDeadline{msg: msg, deadline: deadline}
+	w.ok <- msg
 	return msg, err
 }
 
 func TestRealtimeConn_RTN23(t *testing.T) {
 	t.Parallel()
 	var query url.Values
-	ok := make(chan *messageWithDeadline, 2)
+	ok := make(chan *proto.ProtocolMessage, 2)
 	timeout := 30 * time.Millisecond
 	app, c := ablytest.NewRealtime(&ably.ClientOptions{
 		RealtimeRequestTimeout: timeout,
@@ -1117,8 +1112,8 @@ func TestRealtimeConn_RTN23(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	v := <-ok // consume the connect message
-	maxIdleInterval := time.Duration(v.msg.ConnectionDetails.MaxIdleInterval) * time.Millisecond
+	msg := <-ok // consume the connect message
+	maxIdleInterval := time.Duration(msg.ConnectionDetails.MaxIdleInterval) * time.Millisecond
 	receiveTimeout := timeout + maxIdleInterval
 
 	{ // RTN23b
@@ -1129,15 +1124,15 @@ func TestRealtimeConn_RTN23(t *testing.T) {
 	}
 	{ //RTN23a
 		select {
-		case v := <-ok:
+		case msg := <-ok:
 			// The test here is ensuring we receive a heartbeat within the timeout bounds
 			// as per rtn23a because the connection is idle since receiving connection
 			// message from ably.
-			if v.msg == nil {
+			if msg == nil {
 				t.Fatal("expected to receive heartbeat message")
 			}
-			if v.msg.Action != proto.ActionHeartbeat {
-				t.Errorf("expected action %v got %v", proto.ActionHeartbeat, v.msg.Action)
+			if msg.Action != proto.ActionHeartbeat {
+				t.Errorf("expected action %v got %v", proto.ActionHeartbeat, msg.Action)
 			}
 		case <-time.After(receiveTimeout):
 			t.Error("no heartbeat was sent by ably")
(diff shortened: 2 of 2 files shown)


ably/realtime_conn.go
```diff
@@ -84,6 +84,10 @@ func newConn(opts *ClientOptions, auth *Auth, callbacks connCallbacks) (*Connect
 }
 
 func (c *Connection) dial(proto string, u *url.URL) (proto.Conn, error) {
+	// (RTN23b)
+	query := u.Query()
+	query.Add("heartbeats", "true")
```
tcard, line 89: RTN23b says that you may rely on the WebSocket's heartbeat mechanism, which we [already do by setting the connection's read deadline](https://github.com/ably/ably-go/blob/e362d92f93b0b04408c1b4299ae0ea34a0d64f93/ably/internal/ablyutil/websocket.go#L25). So it shouldn't be necessary to send the `heartbeat=true` param.

ably/realtime_conn_spec_test.go
```diff
@@ -1078,3 +1078,64 @@ func TestRealtimeConn_RTN16(t *testing.T) {
 func sameConnection(a, b string) bool {
 	return strings.Split(a, "-")[0] == strings.Split(b, "-")[0]
 }
+
+type wrapProtoMessage struct {
```
tcard, line 1082: `wrapProtoMessage` doesn't really express what this does. 

* The wrapped thing is the `proto.Conn`, not the protocol message.
* The conn is wrapped to add a hook on incoming protocol messages.

Suggestion: How about `protoConnWithReceiveHook`?

ably/realtime_conn_spec_test.go
```diff
@@ -1078,3 +1078,64 @@ func TestRealtimeConn_RTN16(t *testing.T) {
...
+	ok chan *proto.ProtocolMessage
+}
+
+func (w *wrapProtoMessage) Receive(deadline time.Time) (*proto.ProtocolMessage, error) {
+	msg, err := w.Conn.Receive(deadline)
+	w.ok <- msg
+	return msg, err
+}
+
+func TestRealtimeConn_RTN23(t *testing.T) {
```
tcard, line 1093: There's a very similar test called TestRealtimeConn_ReceiveTimeout. Please consider unifying both.

ably/realtime_conn_spec_test.go
gernest: addressed d6515f72862ebc2d48deea576e03f249056e4c2b

ably/realtime_conn.go
gernest: I have checked our websocket library to see how it handles the Ping message.

https://github.com/golang/net/blob/ab34263943818b32f575efc978a3d24e80b04bd7/websocket/hybi.go#L291-L303

See there, this only receives ping and sends pong message without resetting the deadline.
Meaning the deadline which we only set once on every Receive call applies to all `Read` on conn and hence it will timeout even though we received a Ping from ably.

In short the deadline is never reset after receiving ping, so reads will eventually timeout based on the deadline set once on every Receive.

There is no api for custom handling of ping for this library so its impossible to rely on websocket pings

ably/realtime_conn_spec_test.go
gernest: I feel like they are different because we aren't really testing receive timeouts here, we are testing heartbeats.

ably/realtime_conn.go
tcard: The deadline is resetted when the connection receives _any_ data (which is the behavior we want), not at the `net/websocket` layer, but at the underlying `net` layer.

ably/realtime_conn_spec_test.go
tcard: Per my other comments, we don't need heartbeats: we can rely on WebSocket pings and `SetReadDeadline`. Then, a receive timeout is the equivalent of a missing heartbeat.

ably/realtime_conn.go
gernest: > 
	// SetDeadline sets the read and write deadlines associated
	// with the connection. It is equivalent to calling both
	// SetReadDeadline and SetWriteDeadline.
	//
	// A deadline is an absolute time after which I/O operations
	// fail with a timeout (see type Error) instead of
	// blocking. The deadline applies to all future and pending
	// I/O, not just the immediately following call to Read or
	// Write. After a deadline has been exceeded, the connection
	// can be refreshed by setting a deadline in the future.
	
@tcard documentation for `net.Conn` says otherwise. All future calls will timeout when deadline is reached, you need to manually call `SetDeadline` again to reset it.

So, it isn't reset at all. Reads from `Ping` won't reset it.

tcard: You're right, sorry for the confusion. I agree we should activate Ably heartbeats. We could go lower-level and handle incoming frames ourselves but I don't think it's worth the complexity.

ably/realtime_conn_spec_test.go
tcard: See https://github.com/ably/ably-go/pull/169#discussion_r456460786.

ably/realtime_conn_spec_test.go
```diff
@@ -1078,3 +1078,64 @@ func TestRealtimeConn_RTN16(t *testing.T) {
...
+			// The test here is ensuring we receive a heartbeat within the timeout bounds
+			// as per rtn23a because the connection is idle since receiving connection
+			// message from ably.
+			if msg == nil {
+				t.Fatal("expected to receive heartbeat message")
+			}
+			if msg.Action != proto.ActionHeartbeat {
+				t.Errorf("expected action %v got %v", proto.ActionHeartbeat, msg.Action)
+			}
+		case <-time.After(receiveTimeout):
```
tcard, line 1137: Sorry for being late here, but a few comments.

This test is very flaky and fails a lot on my machine with:

```
$ go test -race -count=1 -v -run='TestRealtimeConn_RTN23' ./...
=== RUN   TestRealtimeConn_RTN23
=== PAUSE TestRealtimeConn_RTN23
=== CONT  TestRealtimeConn_RTN23
    TestRealtimeConn_RTN23: realtime_conn_spec_test.go:1146: no heartbeat was sent by ably
```

Also, we're missing a case here. [RTN23a says](https://docs.ably.io/client-lib-development-guide/features/#RTN23a):

> If a transport does not receive any indication of activity on a transport for a period greater than the sum of the maxIdleInterval (which will be sent in the connectionDetails of the most recent CONNECTED message received on that transport) and the realtimeRequestTimeout, that transport should be disconnected. 

But we're not testing this case. We're testing the case in which Ably _does_ send a heartbeat in time. We're missing a test that checks that we disconnect if we don't get a heartbeat.

Also, note that `maxIdleInterval` is quite big (15 s). I think we should mock a shorter `maxIdleInterval`, so that tests don't take too long. (We also avoid flakiness this way.)

tcard: PS. I said "on my machine" but it's failing a lot in CI too.

gernest: @tcard its weird, that test is passing , it takes a long time though

```
$ go test -v -run TestRealtimeConn_RTN23 ./ably/
=== RUN   TestRealtimeConn_RTN23
=== PAUSE TestRealtimeConn_RTN23
=== CONT  TestRealtimeConn_RTN23
--- PASS: TestRealtimeConn_RTN23 (25.46s)
PASS
ok      github.com/ably/ably-go/ably    25.477s
```

tcard: See a [recent occurrence](https://travis-ci.org/github/ably/ably-go/jobs/713709989) of this test failing, and [another](https://travis-ci.org/github/ably/ably-go/jobs/713709990). The second one with a different message:

```
--- FAIL: TestRealtimeConn_RTN23 (37.50s)
    realtime_conn_spec_test.go:1140: expected to receive heartbeat message
```

Also, what about this?

> Also, we're missing a case here. [RTN23a says](https://docs.ably.io/client-lib-development-guide/features/#RTN23a):
> 
> > If a transport does not receive any indication of activity on a transport for a period greater than the sum of the maxIdleInterval (which will be sent in the connectionDetails of the most recent CONNECTED message received on that transport) and the realtimeRequestTimeout, that transport should be disconnected.
> 
> But we're not testing this case. We're testing the case in which Ably _does_ send a heartbeat in time. We're missing a test that checks that we disconnect if we don't get a heartbeat.

tcard: I've disabled the test at #178 to keep CI stable. Should be enabled again once this is addresses.

PR #170, opened 2020-07-18
add RTN19 - Transport state side effect

gernest:
-  [x] RTN19a
- [x] RTN19b)

### DIFF
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -242,11 +242,21 @@ func (c *RealtimeChannel) Detach(ctx context.Context) error {
 func (c *RealtimeChannel) detach() (Result, error) {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
-	switch {
-	case c.state == ChannelStateFailed:
-		return nil, channelStateError(ChannelStateFailed, errDetach)
-	case !c.isActive():
+	if !c.isActive() {
 		return nopResult, nil
+	}
+	return c.detachUnsafe()
+}
+
+func (c *RealtimeChannel) detachSkipVerifyActive() (Result, error) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.detachUnsafe()
+}
+
+func (c *RealtimeChannel) detachUnsafe() (Result, error) {
+	if c.state == ChannelStateFailed {
+		return nil, channelStateError(ChannelStateFailed, errDetach)
 	}
 	if !c.client.Connection.lockIsActive() {
 		return nil, c.lockSetState(ChannelStateFailed, errDetach)
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -217,11 +217,7 @@ func (c *RealtimeChannel) lockAttach(err error) (Result, error) {
 		Action:  proto.ActionAttach,
 		Channel: c.Name,
 	}
-	err = c.client.Connection.send(msg, nil)
-	if err != nil {
-		return nil, c.setState(ChannelStateFailed, err)
-	}
-
+	c.client.Connection.send(msg, nil)
 	return res, nil
 }
 
@@ -267,10 +263,7 @@ func (c *RealtimeChannel) detachUnsafe() (Result, error) {
 		Action:  proto.ActionDetach,
 		Channel: c.Name,
 	}
-	err := c.client.Connection.send(msg, nil)
-	if err != nil {
-		return nil, c.lockSetState(ChannelStateFailed, err)
-	}
+	c.client.Connection.send(msg, nil)
 	return res, nil
 }
 
@@ -443,9 +436,7 @@ func (c *RealtimeChannel) send(msg *proto.ProtocolMessage) (Result, error) {
 	}
 
 	res, listen := newErrResult()
-	if err := c.client.Connection.send(msg, listen); err != nil {
-		return nil, err
-	}
+	c.client.Connection.send(msg, listen)
 	return res, nil
 }
 
--- a/ably/realtime_client.go
+++ b/ably/realtime_client.go
@@ -81,6 +81,8 @@ func (c *Realtime) onReconnected(isNewID bool) {
 		// TODO: SUSPENDED
 		case ChannelStateAttaching, ChannelStateAttached:
 			ch.mayAttach(false)
+		case ChannelStateDetaching:
+			ch.detachSkipVerifyActive()
 		}
 	}
 }
--- a/ably/realtime_client.go
+++ b/ably/realtime_client.go
@@ -73,6 +73,8 @@ func (c *Realtime) onReconnected(isNewID bool) {
 		for _, ch := range c.Channels.All() {
 			ch.queue.Flush()
 		}
+		//RTN19a
+		c.Connection.resendPending()
 		return
 	}
 
@@ -85,6 +87,8 @@ func (c *Realtime) onReconnected(isNewID bool) {
 			ch.detachSkipVerifyActive()
 		}
 	}
+	//RTN19a
+	c.Connection.resendPending()
 }
 
 func (c *Realtime) onReconnectionFailed(err *proto.ErrorInfo) {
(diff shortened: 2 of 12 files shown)


ably/realtime_client.go
```diff
@@ -73,8 +73,11 @@ func (c *Realtime) onReconnectMsg(msg *proto.ProtocolMessage) {
 				switch ch.State() {
 				case StateConnSuspended:
 					ch.attach(false)
-				case StateChanAttaching, StateChanAttached:
+				case StateChanAttaching, //(RTN19b)
+					StateChanAttached:
 					ch.mayAttach(false, false)
+				case StateChanDetached: //(RTN19b)
+					ch.mayDetach(false, false)
```
tcard, line 80: RTN19 says "when a transport is upgraded or disconnected for any reason", but here we're reconnecting, not disconnecting. ("Upgraded" isn't really well-defined, so I'm just ignoring it for now; I've opened a Slack conversation about this.)

gernest: I think its proper to do it here because  se resend these on new connection. We reconnect after a disconnect meaning we have a new connection by this time.

ably/realtime_channel.go
```diff
@@ -236,12 +236,16 @@ var detachResultStates = []StateEnum{
 }
 
 func (c *RealtimeChannel) detach(result bool) (Result, error) {
+	return c.mayDetach(result, true)
+}
+
+func (c *RealtimeChannel) mayDetach(result bool, checkActive bool) (Result, error) {
```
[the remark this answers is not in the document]
gernest: done 147afc6d10e0a6f5d3e64d2a173c957171bc1a00

ably/realtime_client.go
tcard: You're right, I wasn't making much sense. We need a connection to send protocol messages!

ably/realtime_client.go
```diff
@@ -71,12 +71,14 @@ func (c *Realtime) onReconnected(err *proto.ErrorInfo, isNewID bool) {
...
 	}
-
 	for _, ch := range c.Channels.All() {
 		switch ch.State() {
 		// TODO: SUSPENDED
-		case StateChanAttaching, StateChanAttached:
+		case StateChanAttaching, //(RTN19b)
+			StateChanAttached:
 			ch.mayAttach(false, false)
+		case StateChanDetached: //(RTN19b)
```
tcard, line 80: RTN19b: "If there are any pending channels i.e. in the ATTACHING or **DETACHING** state"

```suggestion
		case StateChanDetaching: //(RTN19b)
```

lmars:
@gernest what is the status of this PR? This came up in https://github.com/ably/ably-go/pull/208#discussion_r519712866, would be good to get an ETA on when we expect this to be delivered.

ably/realtime_conn_spec_test.go
```diff
@@ -2144,3 +2144,217 @@ func TestRealtimeConn_RTN2g(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	changes := make(ably.ConnStateChanges, 2)
+	off := c.Connection.On(ably.ConnectionEventConnected, changes.Receive)
+	defer off()
+	c.Connect()
+	connIDs <- "1"
+	ablytest.Soon.Recv(t, nil, changes, t.Fatalf)
+	attaching := c.Channels.Get("attaching")
+	go attaching.Attach(context.Background())
```
tcard, line 2181: As commented in `sendAndWait`, AFAICT this `Attach` will block forever. If, instead of just `go`ing, you checked that ultimately those operations returned, this test would've caught that bug.

As a general pattern, `go` in tests should never happend and we should use `ablytest.ResultGroup.GoAdd` instead, in order to ensure we don't have goroutine leaks.

gernest: done c47d51b

ably/realtime_conn.go
```diff
@@ -664,6 +664,34 @@ func (c *Connection) setSerial(serial int64) {
...
+	lg.Debugf("resending %d messages waiting for ACK/NACK", len(cx))
+	for _, v := range cx {
+		go c.sendAndWait(v.msg)
+		close(v.ch)
+	}
+}
+
+func (c *Connection) sendAndWait(msg *proto.ProtocolMessage) {
+	lg := c.logger().sugar()
+	res, listen := newErrResult()
```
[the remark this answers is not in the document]
gernest: my bad! fixed at d174733

ably/realtime_conn.go
```diff
@@ -664,6 +664,34 @@ func (c *Connection) setSerial(serial int64) {
...
+func (c *Connection) resendPending() {
+	lg := c.logger().sugar()
+	c.mtx.Lock()
+	cx := make([]msgCh, len(c.pending.queue))
+	copy(cx, c.pending.queue)
+	c.pending.queue = []msgCh{}
+	c.mtx.Unlock()
+	lg.Debugf("resending %d messages waiting for ACK/NACK", len(cx))
+	for _, v := range cx {
+		go c.sendAndWait(v.msg)
```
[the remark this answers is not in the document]
gernest: we don't need it if we use the pending listener, fixed at d174733

ably/realtime_conn.go
```diff
@@ -664,6 +664,22 @@ func (c *Connection) setSerial(serial int64) {
...
+	c.mtx.Lock()
+	cx := make([]msgCh, len(c.pending.queue))
+	copy(cx, c.pending.queue)
+	c.pending.queue = []msgCh{}
+	c.mtx.Unlock()
+	lg.Debugf("resending %d messages waiting for ACK/NACK", len(cx))
+	for _, v := range cx {
+		err := c.send(v.msg, v.ch)
+		if err != nil {
+			lg.Errorf("failed to re send message %v with error:%v", v.msg, err)
```
lmars, line 678: Should the message not be re-queued?

tcard: `send` [requeues the message and returns nil](https://github.com/ably/ably-go/blob/2c4d62974472f5f8178dc20ba0e90935aa13ec2b/ably/realtime_conn.go#L584-L585), following [RTL6c](https://docs.ably.io/client-lib-development-guide/features/#RTL6c).

_But_ that only works if the connection is already disconnected. If it gets disconnected as a result of the send operation failing, the message won't be requeued, nor enqueued in the first place.

It seems to me that `send` shouldn't return an error at all. The error feedback will be sent to the channel it takes, eventually. But its job can't fail —it either succeeds sending the message, succeeds enqueueing it or succeeds sending an error through the channel (that is handled elsewhere).

This problem predates this PR, so maybe it can be fixed in another PR?

lmars: Ok, but I asked the same question [here](https://github.com/ably/ably-go/pull/208#discussion_r519712866) and you said this PR covers fixing it, so it's fine if this PR _doesn't_ fix it, but we really should fix it 😉

gernest: @lmars  I think you bring up a good point here. This is a direct send .

@tcard I think we  might need to address it here. The logic was supposed to be same as `RealtimeChannel.send` which takes care of queuing and address @lmars point. We just don't have reference to the original Channel that sent the message at this point , need to figure out out how to pass that reference.

gernest: Mmh! Thinking about it again, I think @tcard is right, I misunderstood the issue. 

@lmars 
>Should the message not be re-queued?
It depends, you should know that, by the time we resending stuff, we are in a new transport meaning we have acrive connection. So the message might be probably either failed  to `queue` or `sent`

But, I also think `send` should not return an error at all. Any error should be  sent to the `listen` channel

lmars: Ok, from a high level we just shouldn't be dropping messages that fail to send, they should be re-queued and resent.

lmars: @gernest have you been able to make any progress on this?

lmars: LGTM thanks 👍

lmars (approved):

PR #172, opened 2020-07-23
add rtn14 - Connection opening failures:

gernest:
- [x] RTN14a
- [x] RTN14b
- [x] RTN14c
- [x] RTN14d
- [x] RTN14e
- [x] RTN14f
- [x] RTN14g


ably/ablytest/recorders.go
```diff
@@ -422,7 +422,7 @@ type HostRecorder struct {
 
 	mu         sync.Mutex
 	httpClient *http.Client
-	dialWS     func(string, *url.URL) (proto.Conn, error)
+	dialWS     func(string, *url.URL, time.Duration) (proto.Conn, error)
```
tcard, line 425: Suggestion: How about taking a `context.Context` instead? It's the established way of passing a timeout or deadline.

ably/realtime_conn.go
```diff
@@ -85,10 +85,16 @@ func (c *Connection) dial(proto string, u *url.URL) (proto.Conn, error) {
 	query := u.Query()
 	query.Add("heartbeats", "true")
 	u.RawQuery = query.Encode()
+	// (RTN14c) we are setting timeout to connect.
+	timeout := c.opts.RealtimeRequestTimeout
+	if timeout == 0 {
+		// use defined default value
+		timeout = defaultOptions.RealtimeRequestTimeout
```
tcard, line 95: If you rebase this on top of latest `integration/1.2`, this isn't necessary, as the default values get applied [earlier](https://github.com/ably/ably-go/blob/43dafd3246e3ef14b3091c8b3ea45ddd9867f11b/ably/rest_client.go#L139).

ably/ablytest/recorders.go
gernest: We can't, the underlying `x/websocket` library uses `Dial` instead of `DialContext` to connect. So the only way to have the timeout is is through `Dialer` configuration.

ably/realtime_conn.go
gernest: Thanks . Done. I did a `git push -f` after a rebase so can't reference the commit.

tcard:
Looking good so far.

ably/realtime_conn.go
```diff
@@ -86,23 +90,78 @@ func newConn(opts *clientOptions, auth *Auth, callbacks connCallbacks) (*Connect
...
 	if c.opts.Dial != nil {
-		return c.opts.Dial(proto, u)
+		return c.opts.Dial(protocol, u, timeout)
+	}
+	return ablyutil.DialWebsocket(protocol, u, timeout)
+}
+
+// recoverable returns true if err is recoverable, err is from making a
+// connection
+func recoverable(err error) bool {
```
gernest, line 155: Need to define what a recoverable error must be.

ably/export_test.go
```diff
@@ -103,6 +103,12 @@ func (os ClientOptions) RealtimeRequestTimeout(d time.Duration) ClientOptions {
 	})
 }
 
+func (os ClientOptions) ConnectionStateTTL(d time.Duration) ClientOptions {
```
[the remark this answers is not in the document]
tcard: PS. Now that I think of it, you won't have a ConnectionDetails given RTN14 is about establishing a connection. It really doesn't make sense that [RTN14e](https://docs.ably.io/client-lib-development-guide/features/#RTN14e) is under RTN14, because it seems to be about recovering an existing connection. Connections not yet established don't have a state! I'll bring this up.

tcard (commented):
Almost there! Mainly suggestions, but I strongly think we should simplify the retry logic.

But also note that I don't think RTN14 should deal with SUSPENDED at all. I've brought that up internally, so let's stand by that.

ably/export_test.go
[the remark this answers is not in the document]
tcard: @gernest The situation here has been clarified in https://github.com/ably/docs/pull/974. Ignore my previous comments; RTN14e and SUSPENDED is _not_ (only) about connection state recovery. So it makes sense it can be set also as an option for tests.

(Although I think it's preferable to mock the timeout code to control it from tests, instead of running real timeouts with artificially small durations.)

ably/realtime_conn.go
```diff
@@ -86,23 +90,78 @@ func newConn(opts *clientOptions, auth *Auth, callbacks connCallbacks) (*Connect
...
+	start := time.Now()
+	lg.Debugf("Dial protocol=%q url %q ", proto, u.String())
 	// (RTN23b)
 	query := u.Query()
 	query.Add("heartbeats", "true")
 	u.RawQuery = query.Encode()
+	// We first try a single dial to see if we are successful. There is no need to
+	// have this only in the for loop because most cases we will succeed after the
+	// first attempt, let's be prudent here.
+	conn, err = c.dialInternal(proto, u, c.opts.realtimeRequestTimeout())
```
lmars, line 114: I think this should be using [`httpOpenTimeout`](https://www.ably.io/documentation/client-lib-development-guide/features#TO3l3)

ably/realtime_conn.go
```diff
@@ -86,23 +90,78 @@ func newConn(opts *clientOptions, auth *Auth, callbacks connCallbacks) (*Connect
...
+	if err != nil {
+		lg.Debugf("Dial Failed in %v with %v", time.Since(start), err)
+		return nil, err
```
lmars, line 117: Should there not be a retry here (or in the caller of `dial`) rather than returning the dial error to the user when the instantiate the client?

ably/realtime_conn_spec_test.go
```diff
@@ -1810,3 +1811,181 @@ func TestRealtimeConn_RTN14(t *testing.T) {
 		}
 	}
 }
+func TestRealtimeConn_RTN14a(t *testing.T) {
+	t.Parallel()
+
+	{ // missing key
```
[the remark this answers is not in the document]
gernest: done 3425b1b55a4be68e59814a685eb8444c5ec67c65

ably/realtime_conn_spec_test.go
```diff
@@ -1810,3 +1811,181 @@ func TestRealtimeConn_RTN14(t *testing.T) {
...
+		tokenError := &proto.ErrorInfo{
+			StatusCode: http.StatusUnauthorized,
+			Code:       40140,
+		}
+		in <- &proto.ProtocolMessage{
+			Action:            proto.ActionError,
+			ConnectionDetails: &proto.ConnectionDetails{},
+			Error:             tokenError,
+		}
+		change := make(chan ably.ConnectionStateChange, 1)
```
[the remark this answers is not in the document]
gernest: done 3f7ab824145f35bda33f8f54abb9d9e5565f0f0f and 0c27cd4c696aff31a78d016a5ac7039e03065050

ably/realtime_conn.go
```diff
@@ -86,23 +90,78 @@ func newConn(opts *clientOptions, auth *Auth, callbacks connCallbacks) (*Connect
...
-		if _, err := c.connect(false); err != nil {
+		if _, err := c.connect(connArgs{}); err != nil {
 			return nil, err
 		}
 	}
 	return c, nil
 }
 
-func (c *Connection) dial(proto string, u *url.URL) (proto.Conn, error) {
+func (c *Connection) dial(proto string, u *url.URL, once bool) (conn proto.Conn, err error) {
```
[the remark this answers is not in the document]
gernest: removed in 8e05397d44441b084ecaa59b9ae8ecc74f3bcc1e

ably/realtime_conn.go
```diff
@@ -208,45 +280,151 @@ func (c *Connection) params(mode connectionMode) (url.Values, error) {
 	return query, nil
 }
 
-func (c *Connection) connectWith(result bool, mode connectionMode) (Result, error) {
+func (c *Connection) connectWith(arg connArgs) (Result, error) {
```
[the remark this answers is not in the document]
gernest: done dc4366d1f8b23c9b9b3b34b95b3b81d2f4920c0b

ably/realtime_conn.go
```diff
@@ -208,45 +280,151 @@ func (c *Connection) params(mode connectionMode) (url.Values, error) {
...
+					lg.Debug("exiting dial retry loop")
+					return nopResult, nil
+				case <-suspend:
+					// (RTN14f)
+					lg.Debug("Reached SUSPENDED state while reconnecting")
+					return c.connetAfterSuspension(arg)
+				case <-next.C:
+					lg.Debug("Attemting to dial")
+					res, err := c.connectWithInternal(arg)
+					if err != nil {
```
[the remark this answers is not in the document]
gernest: done 5249085d5ba1d4276053f98d92425ebd3c16c0ae

ably/realtime_conn.go
```diff
@@ -208,45 +280,151 @@ func (c *Connection) params(mode connectionMode) (url.Values, error) {
 	return query, nil
 }
 
-func (c *Connection) connectWith(result bool, mode connectionMode) (Result, error) {
+func (c *Connection) connectWith(arg connArgs) (Result, error) {
+	lg := c.logger().Sugar()
+	res, err := c.connectWithInternal(arg)
+	if err != nil {
```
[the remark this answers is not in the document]
gernest: done afbb52cf00bb4de143af0de3d38813f826b1d1b9

ably/realtime_conn.go
gernest: done 01b4ab2f6e0a0c19d7da4bd465dcc42ab6b2f810

ably/realtime_conn.go
gernest: I had changed the way where retry happen but forgot to update comments. See `Connection.connectWithRetryLoop`  there is no need for `dialInternal` anymore I removed it  in 15287625ff558b4dd8033516056dd9a62f6931c6

ably/realtime_conn_spec_test.go
```diff
@@ -1810,3 +1811,181 @@ func TestRealtimeConn_RTN14(t *testing.T) {
...
+		var state ably.ConnectionStateChange
+		ablytest.Instantly.Recv(t, &state, change, t.Fatalf)
+		if expect, got := ably.ConnectionStateDisconnected, state.Current; expect != got {
+			t.Errorf("expected %v got %v", expect, got)
+		}
+		if expect, got := bad, c.Connection.ErrorReason(); !strings.Contains(got.Error(), expect) {
+			t.Errorf("expected %v to contain %q", got, expect)
+		}
+		n := reauth.Load().(int)
+		if n != 1 {
```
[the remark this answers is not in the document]
gernest: fixed in d53c13df635d39f71691ae14d416fa10e84a8771

ably/realtime_conn_spec_test.go
```diff
@@ -1989,3 +1989,59 @@ func TestRealtimeConn_RTN14b(t *testing.T) {
...
+				w, err := ablytest.MessagePipe(in, out)(protocol, u, timeout)
+				if err != nil {
+					return nil, err
+				}
+				ls = &closeConn{Conn: w}
+				return ls, nil
+			}))
+		c.Connect()
+		// Get the connection to CONNECTED.
+		tokenError := &proto.ErrorInfo{
```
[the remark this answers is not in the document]
gernest: nice catch.  fixed in 898cbca2cd7384247c076fbb7fdd0846470d4643

ably/realtime_conn.go
```diff
@@ -86,23 +90,78 @@ func newConn(opts *clientOptions, auth *Auth, callbacks connCallbacks) (*Connect
...
-func (c *Connection) dial(proto string, u *url.URL) (proto.Conn, error) {
+func (c *Connection) dial(proto string, u *url.URL, once bool) (conn proto.Conn, err error) {
+	lg := c.logger().Sugar()
+	start := time.Now()
+	lg.Debugf("Dial protocol=%q url %q ", proto, u.String())
 	// (RTN23b)
 	query := u.Query()
 	query.Add("heartbeats", "true")
 	u.RawQuery = query.Encode()
+	// We first try a single dial to see if we are successful. There is no need to
```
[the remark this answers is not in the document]
gernest: apologies, these are comments left from my previous implementation.  removed in 15287625ff558b4dd8033516056dd9a62f6931c6

ably/realtime_conn.go
```diff
@@ -86,23 +90,78 @@ func newConn(opts *clientOptions, auth *Auth, callbacks connCallbacks) (*Connect
...
+}
+
+// recoverable returns true if err is recoverable, err is from making a
+// connection
+func recoverable(err error) bool {
+	if info, ok := err.(*ErrorInfo); ok {
+		err = info.err
+	}
+	switch err {
+	case context.DeadlineExceeded:
```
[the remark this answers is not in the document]
gernest: done d43301acae97e924b28ed857000264b420bda373

ably/realtime_conn.go
lmars: @gernest sorry I don't think I was clear in what I said.

What I mean is that I don't think this should be returning an error if the initial dial attempt fails with a recoverable error, instead it _should_ be retrying.

To demonstrate, when I run the following code to connect to `192.168.0.200` which I expect to timeout (and after moving [this code](https://github.com/ably/ably-go/blob/feature/conn-open-failures-rtn14/ably/realtime_client.go#L36-L38) to after the error check to avoid a panic):

```go
log.Println("Connecting to 192.168.0.200")

opts := ably.ClientOptions{}.
	Key("xxx:xxx").
	RESTHost("192.168.0.200").
	RealtimeHost("192.168.0.200").
	HTTPOpenTimeout(1 * time.Second)

if _, err := ably.NewRealtime(opts); err != nil {
	log.Fatal(err)
}
```

I get an error after one second and the program exits:

```
2020/10/18 12:23:09 Connecting to 192.168.0.200
2020/10/18 12:23:10 [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603020189482: dial tcp 192.168.0.200:443: i/o timeout code=80000 statusCode=0] See https://help.ably.io/error/80000
exit status 1
```

whereas I'd expect that to enter into a retry loop as a dial timeout is a recoverable error.

I think this relates to the design I proposed in https://github.com/ably/ably-go/pull/144#discussion_r404785636 where `NewRealtime` doesn't return an error and has a separate `Start` method that enters an event loop that emits errors to registered handlers (including connection errors), which would look something like:

```go
realtime := ably.NewRealtime(opts)

realtime.Connection.On(ably.CONNECTED, func() {
	// do something now we know we are connected
})

// start enters a blocking event loop, attempts connections
// transparently manages reconnecting / resuming / recovering
if err := realtime.Start(); err != nil {
	// start only returns an error when the connection is unrecoverable
}
// calling realtime.Stop() causes the event loop to gracefully exit
```

cc @tcard, did you consider this design?

paddybyers: The spec says that unless `autoConnect == false`, then the connection sequence starts automatically on instantiating the library; how do we incorporate this into this API? `NewRealtime()` shouldn't return an error unless the given `ClientOptions` are invalid.

lmars: > the connection sequence starts automatically on instantiating the library; how do we incorporate this into this API?

Ah ok I didn't realise that, I proposed starting the loop as a separate step to avoid a race between the library starting to do stuff and the user registering handlers for stuff, although that could be handled by allowing handlers to be registered via the options too, for example:

```go
ably.NewRealtime(
	ably.WithConnectionHandler(func() {
		// handle connection change events
	}),
)
```

tcard: > although that could be handled by allowing handlers to be registered via the options too

Should we move that discussion to the spec?

---

To reiterate your original point so that it doesn't get lost:

> I don't think this should be returning an error if the initial dial attempt fails with a recoverable error, instead it _should_ be retrying.

This totally should be fixed.

lmars: > Should we move that discussion to the spec?

I think this is a language issue rather than a spec issue. For example, I don't believe this is a problem in JavaScript as it will not pre-emptively schedule the establishment of connections before the handlers have been registered, so requiring the handlers be registered in the options there doesn't seem necessary.

I may be wrong though, and even if I'm right, it still may be better to make this clear in the spec. @paddybyers WDYT?

> This totally should be fixed.

I agree, but it would be good to agree on how we think it should be fixed, and if we think we should design the API to avoid races between connections being made and handlers being registered or not, as I think now is the correct time to make that design decision (happy to discuss elsewhere though to not block this PR).

gernest: @lmars 

The spec makes it easier for a callback based API's (JS etc). Go on the other hand looks much of API's aren't callback based.

It is much convenient to keep blocky part around and let users execute them in goroutines to fit the style of the app they are building. `Realtime` is safe for concurrent use.

```
ably.NewRealtime(
	ably.WithConnectionHandler(func() {
		// handle connection change events
	}),
```

Isn't this already covered by event emitter api? we already have `Realtime.Connection.On()` where users can watch for events as they happen.

lmars: > Isn't this already covered by event emitter api? we already have Realtime.Connection.On() where users can watch for events as they happen.

It is covered by that API yes, but it presents a race condition between the user registering the handler and the library emitting the events.

For example, let's say I'm interested in tracking the connection lifecycle, then under the current design I would write something like:

```go
realtime, err := ably.NewRealtime(...)
if err != nil {
	return err
}
realtime.Connection.On(func(state ably.State) {
	if state == ably.StateConnConnected {
		// handle being connected
	} else {
		// handle not being connected
	}
})
```

But given the `NewRealtime` method has already created the connection, there is no guarantee that my handler is going to receive the `StateConnConnected`, so my program may think it's not connected when it is.

We may of course decide to ignore this race condition, but given we have a choice right now to eliminate it, perhaps we should.

lmars: Sorry I was unaware of the `autoConnect` option (which I now realise Paddy mentioned above), so we can document the use of that for those that need to avoid the race condition (i.e. instantiate the library with `AutoConnect(false)`, register the handlers, call `realtime.Connect()` to start the connect loop).

@gernest sorry to add that noise to this thread, so the action here is to start the connection loop here in `NewRealtime` and not return dial errors to the user.

ably/realtime_conn_spec_test.go
```diff
@@ -1810,3 +1811,181 @@ func TestRealtimeConn_RTN14(t *testing.T) {
...
+		if expect, got := "missing key", c.Connection.ErrorReason(); !strings.Contains(got.Error(), expect) {
+			t.Errorf("expected %v to contain %q", got, expect)
+		}
+	}
+	{ // invalid key
+		in := make(chan *proto.ProtocolMessage, 1)
+		out := make(chan *proto.ProtocolMessage, 16)
+
+		c, _ := ably.NewRealtime(ably.ClientOptions{}.
+			Token("fake:token").
```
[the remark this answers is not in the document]
gernest: If we remove this, then `c` will be nil because we call `NewREST` inside `NewRealtime` which will fail with `invalid key` error. I am testing for the `Realtime` and not `REST`

ably/realtime_conn.go
```diff
@@ -208,45 +280,151 @@ func (c *Connection) params(mode connectionMode) (url.Values, error) {
...
 }
 
+func (c *Connection) connectionStateTTL() time.Duration {
+	if c.arg.connDetails != nil && c.arg.connDetails.ConnectionStateTTL != 0 {
+		return time.Duration(c.arg.connDetails.ConnectionStateTTL)
+	}
+	return c.opts.connectionStateTTL()
+}
+
+func (c *Connection) watchDisconnect(ctx context.Context) {
```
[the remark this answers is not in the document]
gernest: done 154688153de80abebf3a32ca9e16f110017e31a8

gernest:
@tcard I addressed majority of the comments. I see we have new API's on the integration branch, Can you PTAL before I rebase?

ably/realtime_conn_spec_test.go
```diff
@@ -1810,3 +1811,181 @@ func TestRealtimeConn_RTN14(t *testing.T) {
...
+		change := make(chan ably.ConnectionStateChange, 1)
+		c.Connection.OnAll(func(ev ably.ConnectionStateChange) {
+			change <- ev
+		})
+		var state ably.ConnectionStateChange
+		ablytest.Instantly.Recv(t, &state, change, t.Fatalf)
+		if expect, got := ably.ConnectionStateDisconnected, state.Current; expect != got {
+			t.Errorf("expected %v got %v", expect, got)
+		}
+		if expect, got := "missing key", c.Connection.ErrorReason(); !strings.Contains(got.Error(), expect) {
```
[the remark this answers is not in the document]
tcard: @gernest This seems to be still pending.

ably/realtime_conn_spec_test.go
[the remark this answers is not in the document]
tcard: That's a good point, yet I'd say this test relies on a bug. Both REST and Realtime should validate _all_ the ClientOptions ([RSC1b](https://docs.ably.io/client-lib-development-guide/features/#RSC1b)). The key should be validated on `NewRealtime` even if a token is passed too.

We'll get to fix that in the future, but for now, the only way to conciliate RSC1b and RTN14a is that RTN14a is about _well-formed_ (so it passes ClientOptions validation) and yet _invalid_ (non-existent, for example?) keys.

tcard:
@gernest Done. From my part, it's only these outstanding:

* https://github.com/ably/ably-go/pull/172#discussion_r508339538 (mainly this one)
* https://github.com/ably/ably-go/pull/172#discussion_r499518352
* https://github.com/ably/ably-go/pull/172#discussion_r509351999

gernest:
thanks

ably/realtime_conn.go
gernest: Hi @lmars , I fixed this in 5c77665eddc7b21c8bc955a6b2ac503e41b84019

add `LogLevel(ably.LogDebug).` in options to see what is happening with the retries.

lmars: @gernest it looks like the behaviour is now to block in `NewRealtime` until a connection is established, is that the desired behaviour? I thought it would just validate the options, start the event loop and return.

gernest: @lmars  You can't real do anything without a connection. If you don't care about the connection you can set `autoConnect` to `false` and manually call `Connect`.

This just make sense to me as a sane choice,  however I'm open to new suggestion. You should be aware that even `Connect` will block until a connection is established.

lmars: > You can't real do anything without a connection

You can, for example, publish messages that will be queued and sent when the connection is established?

I just find it surprising that we're going for an async event loop with events delivered through registered handlers, but starting with a blocking call to `NewRealtime`? As you say, if I care about the connection lifecycle and waiting for the connection to be established (either initially or when the client reconnects), then I can use `autoConnect = false` + register a handler + `client.Connect()`, but if I don't care about the connection lifecycle and just want to instantiate the library and start publishing the messages I want to publish, why do I need to wait?

A non-blocking `NewRealtime` also matches other event-loop based client libraries like [ably-ruby](https://github.com/ably/ably-ruby/blob/v1.1.2/lib/ably/realtime/connection/connection_manager.rb#L30-L33) and [ably-js](https://github.com/ably/ably-js/blob/1.2.3/common/lib/client/connection.js#L38) that also don't block waiting for a connection when instantiating the client (just trigger the start of the connection loop and return).

In summary, I think as long as the behaviour is clearly documented, I'm not too concerned either way, but I was just surprised this was the behaviour when starting an async event loop kind of thing.

gernest: @lmars @tcard 

> You can, for example, publish messages that will be queued and sent when the connection is established?

Won't this be poor design? I think it makes sense if the `tcp` socket is ready then sending messages even if they are queued (waiting for ably to send `CONNECTED`)  It won't take forever.  If there is no socket available the queue will fill up and waste resources.  In case of reconnection it make sense, because we have historical records of successful obtaining the socket.

Sending messages before even having the socket  that might take even forever to be established should be discouraged. however as you mentioned its fine either way. I mean waiting or not waiting based on the library consumer.  I think we don't need to spin up unnecessary goroutines, library consumers can try to do that at their own discretion based on their app programming models.

> I was just surprised this was the behaviour when starting an async event loop kind of thing.

We seem to differ on what we think about  async aspects of the client. I think in terms of `async` communication. In which we have a `eventLoop` for that. However socket programming is not async in Go so I believe establishing sockets should kinda stay blocky to discourage bad designs.

If we want to spin up goroutines ourselves then we need to rethink some of the `api's` and pass `context.Context` in relevant places for control.

So, in the patch, we are introducing connection opening loop, I think this should not be mistaken with the `eventloop`. Because the later reads from the socket established by the former. Some state events might be emitted with the former is happening but will strictly be one of `CONNECTING`, `DISCONNECTED`, `FAILED`, or `SUSPENDED` but they are happening outside of the main event loop.

what do you think about this?

I'm fine with `NewRealtime` not to block even with `autoConnect=true`. My concerns are

- Should `NewRealtime` also accept `context.Context` for user to be able to terminate the `connectWithRetryLoop` function ?
- Is it okay to encourage users to publish messages with clients that will never be able to open connections  connect (assuming something is wrong)?

lmars: > > You can, for example, publish messages that will be queued and sent when the connection is established?

> Won't this be poor design? I think it makes sense if the tcp socket is ready then sending messages even if they are queued (waiting for ably to send CONNECTED) It won't take forever.

The spec is explicit that messages are queued whilst the connection is in the `INITIALIZED` state (see [RTL6c](https://www.ably.io/documentation/client-lib-development-guide/features#RTL6c)), so whilst you may consider this to be bad design, I think we have to stick to the spec here, and discuss changes to that design separately.

Also, it looks like this design has already been adopted by ably-go (see [here](https://github.com/ably/ably-go/blob/integration/1.2/ably/realtime_conn.go#L407-L412) where messages are queued when the state is `ConnectionStateInitialized`), so I don't even think we're talking about a change in design here.

> Sending messages before even having the socket that might take even forever to be established should be discouraged

Again, I think we should make a distinction between implementing features to match the 1.2 spec, and discussing potential design flaws in the spec, because, as I understand it, it is expected that messages are queued whilst the connection is not active, regardless of whether that's because it's the first time the connection is being established or the Nth time, since customers should have every expectation that their connection _will_ establish to Ably in a timely manner, and should not be discouraged from having that expectation.

So, I think our opinions differ, but it seems both of us are willing to compromise, so I'll leave it to @tcard to cast his vote and then let's go with his decision (i.e. `NewRealtime` either blocks until initial connection is established, or just starts the connection loop in a goroutine and returns).

tcard: I don't think it's on the Go library to decide either way. The spec makes it clear that the user-requested operations on the Realtime instance is abstracted from the current state of the connection. That is, attaching, publishing messages, etc. should work (by queueing operations) even if the connection is transiently in a disconnected state. This is a core feature in terms of usability, and pretty much the whole reason we have such complex SDKs.

For this reason, `connect` is only a one-way signal for the library to _start_ the connection loop, same as `close` is a signal for the opposite. It makes sense, given we need to abstract over transient disconnections anyway. As any design decision, as you point out @gernest, it comes with drawbacks, but for users that prefer to avoid them we also offer the `queueMessages` option and the connection event emitter for increased control.

So, while I understand @gernest objections, I think we don't really have an option here but to implement the intent of the spec.

(We also can find a [precedent in the other multithreaded SDK: Java](https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L749-L755).)

> Should `NewRealtime` also accept `context.Context` for user to be able to terminate the `connectWithRetryLoop` function ?

That's what `close` is for.

tcard: > but for users that prefer to avoid them we also offer the `queueMessages` option and the connection event emitter for increased control

I should also add: we don't queue messages indefinitely, but up to the default `connectionStateTtl`. After that, queued and subsequent publishes fail.

gernest: @lmars @tcard Thanks for your thorough observation and review.
I made the relevant changes which address the points discussed here i.e async connection opening for RealtimeClient

The changes are here 4fb3f431260b7b45b91d0c2f90ef7ea09fdaa1cc PTAL

ably/realtime_conn_spec_test.go
```diff
@@ -1837,6 +1835,7 @@ func TestRealtimeConn_RTN14a(t *testing.T) {
 		change := make(ably.ConnStateChanges, 1)
 		c.Connection.OnAll(change.Receive)
 		var state ably.ConnectionStateChange
+		ablytest.Instantly.Recv(t, nil, change, t.Fatalf) // skip CONNECTING
```
tcard, line 1838: This is racy. To ensure we'll get the CONNECTING state into the `change` channel, the `c.Connect` must be below the `OnAll` listener registration.

ably/realtime_conn_spec_test.go
```diff
@@ -1870,6 +1869,7 @@ func TestRealtimeConn_RTN14a(t *testing.T) {
...
```
tcard, line 1872: This is racy. To ensure we'll get the CONNECTING state into the `change` channel, the `c.Connect` must be below the `OnAll` listener registration.

ably/realtime_conn_spec_test.go
```diff
@@ -1913,6 +1913,7 @@ func TestRealtimeConn_RTN14b(t *testing.T) {
 		change := make(ably.ConnStateChanges, 1)
 		c.Connection.OnAll(change.Receive)
 		var state ably.ConnectionStateChange
+		ablytest.Instantly.Recv(t, nil, change, t.Fatalf) // Skip CONNECTING
```
tcard, line 1916: This is racy. To ensure we'll get the CONNECTING state into the `change` channel, the `c.Connect` must be below the `OnAll` listener registration.

ably/realtime_conn_spec_test.go
```diff
@@ -1967,6 +1968,7 @@ func TestRealtimeConn_RTN14b(t *testing.T) {
 			Error:             bad,
 		}
 		var state ably.ConnectionStateChange
+		ablytest.Instantly.Recv(t, nil, change, t.Fatalf) // skip CONNECTING
```
tcard, line 1971: This is racy. To ensure we'll get the CONNECTING state into the `change` channel, the `c.Connect` must be below the `OnAll` listener registration.

ably/realtime_conn_spec_test.go
```diff
@@ -2040,6 +2042,7 @@ func TestRealtimeConn_RTN14g(t *testing.T) {
 		change := make(ably.ConnStateChanges, 1)
 		c.Connection.OnAll(change.Receive)
 		var state ably.ConnectionStateChange
+		ablytest.Instantly.Recv(t, nil, change, t.Fatalf) // Skip CONNECTING
```
tcard, line 2045: This is racy. To ensure we'll get the CONNECTING state into the `change` channel, the `c.Connect` must be below the `OnAll` listener registration.

ably/realtime_conn_test.go
```diff
@@ -120,8 +120,9 @@ func TestRealtimeConn_AuthError(t *testing.T) {
 	if err = ablytest.ConnWaiter(client, client.Connect, ably.ConnectionEventConnected).Wait(); err == nil {
 		t.Fatal("Connect(): want err != nil")
 	}
-	if state := client.Connection.State(); state != ably.ConnectionStateFailed {
-		t.Fatalf("want state=%s; got %s", ably.ConnectionStateFailed, state)
+	// Token errors are recoverable so we stay in DISCONNECTED and RTN14 kicks in
```
tcard, line 123: This test also seems to be directed to the production env. I'd say it's too low-value to fix anyway. I'd suggest just removing it; we test (or will do) auth errors exhaustively elsewhere.

ably/realtime_conn.go
```diff
@@ -87,34 +91,91 @@ func newConn(opts *clientOptions, auth *Auth, callbacks connCallbacks) (*Connect
...
+		return res, nil
 	}
-	return ablyutil.DialWebsocket(proto, u)
+}
+
+// recoverable returns true if err is recoverable, err is from making a
+// connection
+func recoverable(err error) bool {
+	switch e := err.(type) {
+	case *ErrorInfo:
```
tcard, line 149: Just like we use `errors.Is`, we should use `errors.As` for type-assertions.

ably/realtime_conn.go
tcard: @gernest This function tries to exhaustively check all errors that are recoverable, and defaults to false otherwise. I'd say it needs to do the opposite: exhaustively check the handful of errors we know are unrecoverable (that is, 4xx response status from Ably, except token errors in the [40140, 40150) range for which there is some further logic) and default to true (recoverable) otherwise.

There can be all kinds of transient error conditions that aren't timeouts. Those shouldn't kill our retry loop.

ably/realtime_conn_test.go
gernest: done 825723188b172781e15028a6bdfac40b9f26198a

ably/realtime_conn_spec_test.go
gernest: done 80ccec5874864feea4babdcc1db112cf30fdee3e

ably/realtime_conn_spec_test.go
gernest: done 80ccec5874864feea4babdcc1db112cf30fdee3e

ably/realtime_conn_spec_test.go
gernest: done 80ccec5874864feea4babdcc1db112cf30fdee3e

ably/realtime_conn_spec_test.go
gernest: done 80ccec5874864feea4babdcc1db112cf30fdee3e

ably/realtime_conn_spec_test.go
gernest: done 80ccec5874864feea4babdcc1db112cf30fdee3e

ably/realtime_conn.go
gernest: done 80f6c2a0ef728a62a079f1614dffa99401469987 also addressed https://github.com/ably/ably-go/pull/172#discussion_r512090389 in the same commit

lmars:
@gernest I just tried out the changes in this PR by running the following program:

```go
package main

import (
	"log"
	"os"
	"os/signal"
	"syscall"
	"time"

	"github.com/ably/ably-go/ably"
)

func main() {
	log.Println("Connecting to 192.168.0.200")
	opts := ably.ClientOptions{}.
		Key("xxx:xxx").
		RESTHost("192.168.0.200").
		RealtimeHost("192.168.0.200").
		HTTPOpenTimeout(1 * time.Second).
		AutoConnect(false)

	client, err := ably.NewRealtime(opts)
	if err != nil {
		log.Fatal(err)
	}
	defer client.Close()

	client.Connection.OnAll(func(state ably.ConnectionStateChange) {
		log.Println("state change:", state.Event, state.Reason, state.RetryIn)
	})

	client.Connect()

	ch := make(chan os.Signal, 1)
	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
	<-ch
	log.Println("Received signal, exiting...")
}
```

and got the following output:

```
2020/10/28 22:06:41 Connecting to 192.168.0.200
2020/10/28 22:06:41 state change: CONNECTING <nil> 0s
2020/10/28 22:06:42 state change: DISCONNECTED [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603922801994: dial tcp 192.168.0.200:443: i/o timeout code=80003 statusCode=0] See https://help.ably.io/error/80003 0s
2020/10/28 22:06:58 state change: CONNECTING <nil> 0s
2020/10/28 22:06:59 state change: DISCONNECTED [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603922818004: dial tcp 192.168.0.200:443: i/o timeout code=80003 statusCode=0] See https://help.ably.io/error/80003 0s
2020/10/28 22:07:14 state change: CONNECTING <nil> 0s
2020/10/28 22:07:15 state change: DISCONNECTED [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603922834010: dial tcp 192.168.0.200:443: i/o timeout code=80003 statusCode=0] See https://help.ably.io/error/80003 0s
2020/10/28 22:07:30 state change: CONNECTING <nil> 0s
2020/10/28 22:07:30 state change: DISCONNECTED [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603922850017: dial tcp 192.168.0.200:443: connect: host is down code=80003 statusCode=0] See https://help.ably.io/error/80003 0s
2020/10/28 22:07:45 state change: CONNECTING <nil> 0s
2020/10/28 22:07:46 state change: DISCONNECTED [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603922865020: dial tcp 192.168.0.200:443: i/o timeout code=80003 statusCode=0] See https://help.ably.io/error/80003 0s
2020/10/28 22:08:01 state change: CONNECTING <nil> 0s
2020/10/28 22:08:02 state change: DISCONNECTED [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603922881030: dial tcp 192.168.0.200:443: i/o timeout code=80003 statusCode=0] See https://help.ably.io/error/80003 0s
2020/10/28 22:08:17 state change: CONNECTING <nil> 0s
2020/10/28 22:08:18 state change: DISCONNECTED [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603922897039: dial tcp 192.168.0.200:443: i/o timeout code=80003 statusCode=0] See https://help.ably.io/error/80003 0s
2020/10/28 22:08:33 state change: CONNECTING <nil> 0s
2020/10/28 22:08:33 state change: DISCONNECTED [ErrorInfo :websocket.Dial wss://192.168.0.200:443?echo=true&format=msgpack&heartbeats=true&key=xxx%3Axxx&timestamp=1603922913046: dial tcp 192.168.0.200:443: connect: host is down code=80003 statusCode=0] See https://help.ably.io/error/80003 0s
2020/10/28 22:08:43 state change: SUSPENDED [ErrorInfo :context deadline exceeded code=80002 statusCode=0] See https://help.ably.io/error/80002 0s
2020/10/28 22:09:13 state change: CONNECTING <nil> 0s
```

Some issues to address:

- after going to `SUSPENDED`, I get a `CONNECTING` event but nothing more, is it supposed to keep retrying?

- `RetryIn` is always zero

- the error when going to `SUSPENDED` is `context deadline exceeded`, but should it not be the actual error that occurred (since there is no context involved here)?

I'm aware this PR has been open for a while, do let me know if you'd like some assistance in addressing these issues as it would be great to get this merged and included in the 1.2 preview, it's a pretty critical feature for us.

lmars:
P.S. I also got a panic on exit:

```
^C2020/10/28 22:29:32 Received signal, exiting...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x14142f1]

goroutine 1 [running]:
github.com/ably/ably-go/ably.(*Connection).close(0xc0000a6240)
        /Users/lewis/src/github.com/ably/ably-go/ably/realtime_conn.go:393 +0x1c1
github.com/ably/ably-go/ably.(*Connection).Close(...)
        /Users/lewis/src/github.com/ably/ably-go/ably/realtime_conn.go:170
github.com/ably/ably-go/ably.(*Realtime).Close(0xc0000921e0)
        /Users/lewis/src/github.com/ably/ably-go/ably/realtime_client.go:50 +0x2f
main.main()
        /Users/lewis/src/github.com/ably/ably-go/tmp/_dial-error/main.go:38 +0x31c
exit status 2
```

paddybyers:
@gernest this PR is critical to us getting a usable preview release out for 1.2. When do you think you will be able to progress it?

lmars:
@gernest there is unfortunately a bigger problem, running the following program:

```go
package main

import (
	"log"
	"os"
	"os/signal"
	"syscall"

	"github.com/ably/ably-go/ably"
)

func main() {
	log.Println("Connecting to dev")
	opts := ably.ClientOptions{}.
		Key("xxx:xxx").
		Environment("dev")

	client, err := ably.NewRealtime(opts)
	if err != nil {
		log.Fatal(err)
	}
	defer client.Close()

	channel := client.Channels.Get("test")

	res, err := channel.Publish("test", "test")
	if err != nil {
		log.Fatal(err)
	}
	if err := res.Wait(); err != nil {
		log.Fatal(err)
	}
	log.Println("message published")

	ch := make(chan os.Signal, 1)
	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
	<-ch
	log.Println("Received signal, exiting...")
}
```

I get an error:

```
2020/10/29 02:00:26 Connecting to dev
2020/10/29 02:00:26 [ErrorInfo :attempted to attach channel to inactive connection code=80000 statusCode=0] See https://help.ably.io/error/80000
exit status 1
```

It looks like a publish whilst the connection is being made is throwing an error, but it should be queued and sent when the connection is established.

tcard:
> It looks like a publish whilst the connection is being made is throwing an error, but it should be queued and sent when the connection is established.

This isn't part of RTN14 but of [RTL6](https://docs.ably.io/client-lib-development-guide/features/#RTL6c2), which we haven't yet gone through.

gernest:
@lmars 

> after going to SUSPENDED, I get a CONNECTING event but nothing more, is it supposed to keep retrying?

The spec says 

> (RTN14f) The connection will remain in the SUSPENDED state indefinitely, whilst periodically attempting to reestablish a connection

I was even contemplating the having `CONNECTING` state, I wanted to ask about this and forgot. @paddybyers , @tcard  doesn this mean `SUSPENDED` will be the last state indefinitely until `CONNECTED`?

> RetryIn is always zero

Fixed this at e24bbca79d0f41df4c0c6841770da48e0676e198 , note that this is only set when emitting `DISCONNECTED` and `SUSPENDED` when we will make automatic connection opening attempt.


>(TA2) The ConnectionStateChange object contains the current state in attribute current, the previous state in attribute previous, and when the client is not connected and a connection attempt will be made automatically by the library, the amount of time in milliseconds until the next retry in the attribute retryIn

This is my understanding of 👆 

> the error when going to SUSPENDED is context deadline exceeded, but should it not be the actual error that occurred (since there is no context involved here)?

fixed at e24bbca79d0f41df4c0c6841770da48e0676e198  its now looking like `Exceeded connectionStateTtl=2s while in DISCONNECTED` I'm open to improvement on the error message

> P.S. I also got a panic on exit:

fixed at 298ac13e2e7b8f8126a95d62b1398fdb0d9a2bb5 and made a better change in e24bbca79d0f41df4c0c6841770da48e0676e198 , solution is to do nothing when we aren't in `CONNECTED` state because we don't have the connection yet.

>It looks like a publish whilst the connection is being made is throwing an error, but it should be queued and sent when the connection is established.

Also to add on what @tcard said about this , I tried the example and these are the logs

```
2020/10/29 11:29:12 [DEBUG] Dial protocol="application/x-msgpack" url "wss://dev-realtime.ably.io:443?echo=true&format=msgpack&key=xxx%3Axxx&timestamp=1603960152319" 
2020/10/29 11:29:13 [DEBUG] Dial success in 1.604306302s
2020/10/29 11:29:13 [VERBOSE] Realtime Connection: received (action="error", error=&proto.ErrorInfo{StatusCode: 400, Code: 40005, HRef:"https://help.ably.io/error/40005", Message:"Invalid key in request: xxx:xxx. (See https://help.ably.io/error/40005 for help.)", Server:""})
2020/10/29 11:29:13 [VERBOSE] Realtime Connection: closed
```

I thought we can't recover from `Invalid key in request` is it possible to  publish with invalid key?

Also you should be using `RealtimeRequestTimeout(1 * time.Second).` instead of `HTTPOpenTimeout(1 * time.Second).`

as per 
>(RTN14c) A new connection attempt will fail if not connected within the default realtime request timeout

paddybyers:
> This isn't part of RTN14 but of [RTL6](https://docs.ably.io/client-lib-development-guide/features/#RTL6c2), which we haven't yet gone through

@tcard That's now going to be critical then, or we have a breaking change for most users. Have you looked into what needs doing?

tcard:
@gernest

> The spec says
> 
> > (RTN14f) The connection will remain in the SUSPENDED state indefinitely, whilst periodically attempting to reestablish a connection
> 
> I was even contemplating the having `CONNECTING` state, I wanted to ask about this and forgot. @paddybyers , @tcard doesn this mean `SUSPENDED` will be the last state indefinitely until `CONNECTED`?

This isn't the best wording, but no, the retry cycle should go SUSPENDED -> CONNECTING -> SUSPENDED -> ...

What definitely shouldn't be happening is that the connection remains CONNECTING for a long period of time. It should only be CONNECTING while it's actively trying to, well, connect.

> > the error when going to SUSPENDED is context deadline exceeded, but should it not be the actual error that occurred (since there is no context involved here)?
> 
> fixed at [e24bbca](https://github.com/ably/ably-go/commit/e24bbca79d0f41df4c0c6841770da48e0676e198) its now looking like `Exceeded connectionStateTtl=2s while in DISCONNECTED` I'm open to improvement on the error message

The problem here is that we're going from DISCONNECTED directly to SUSPENDED because of the timeout. That shouldn't happen; when such timeout expires, we should first attempt to connect. Thus, it will go DISCONNECTED -> (timeout expires) -> CONNECTING -> (fails with `err`) -> SUSPENDED; and the emitted Reason must be that `err`.

--

@paddybyers 

> @tcard That's now going to be critical then, or we have a breaking change for most users. Have you looked into what needs doing?

Not really, but it doesn't _sound_ hard. I'll get to it ASAP.

gernest:
@paddybyers 

> @gernest this PR is critical to us getting a usable preview release out for 1.2. When do you think you will be able to progress it?

As soon as possible. The functionality is pretty much done(except for minor details). The challenge is, we change how we establish connections and there are lots of tests that relied on the old behaviour, so we broke lots of tests, I'm trying to hunt them down, understand what they do and refactor them to reflect the behaviour change.

lmars:
@gernest 

> I tried the example and these are the logs ... I thought we can't recover from Invalid key in request is it possible to publish with invalid key?

Sorry, I was running it locally with a real key but replaced it with `xxx:xxx` in my example to not expose my API key. Even with an invalid key though, I get the same publish error that exits the program before the library establishes the connection. We've also seen these publish errors in load testing when trying out this branch and had to workaround it, it's definitely an issue.

> Also you should be using `RealtimeRequestTimeout(1 * time.Second)`. instead of `HTTPOpenTimeout(1 * time.Second)`.

Why? My intention was to put a timeout on the time it takes to open a connection, isn't that what `httpOpenTimeout` is for?

gernest:
@lmars 
>Sorry, I was running it locally with a real key but replaced it with xxx:xxx in my example to not expose my API key. Even with an invalid key though, I get the same publish error that exits the program before the library establishes the connection. We've also seen these publish errors in load testing when trying out this branch and had to workaround it, it's definitely an issue

I see, I was mistaken about it. 

>Why? My intention was to put a timeout on the time it takes to open a connection, isn't that what httpOpenTimeout is for?


We are just following spec [RTN14c](https://docs.ably.io/client-lib-development-guide/features/#RTN14c)

Is it okay to use `192.168.0.20` in tests? I failed to mock dial `i/o` timeouts for tests, however using your example it was possible. Not so sure about that IP though.

lmars:
> We are just following spec RTN14c

Ok, but there is a more specific `httpOpenTimeout` (i.e. [`TO3l3`](https://docs.ably.io/client-lib-development-guide/features/#TO3l3)) that can be used specifically for the connection open timeout versus a timeout for actually establishing a connection (which includes TLS handshake + HTTP response headers + WebSocket handshake, which I didn't intend to use a 1s timeout for).

> Is it okay to use 192.168.0.20 in tests? I failed to mock dial i/o timeouts for tests, however using your example it was possible. Not so sure about that IP though.

I wouldn't suggest using that IP no, since it's not guaranteed to timeout everywhere the tests run (it times out for me because I know there is nothing on my home network using that IP so my operating system will timeout trying to connect to it, but that's not guaranteed to be the case everywhere).

I haven't looked at the tests, but I'd suggest implementing a mock dialer that returns an error after the configured timeout, and perhaps use a mocked clock to make it run quickly.

ably/realtime_conn_spec_test.go
```diff
@@ -1787,26 +1787,24 @@ func TestRealtimeConn_RTN23(t *testing.T) {
 
 func TestRealtimeConn_RTN14c(t *testing.T) {
 	t.Parallel()
-	app := ablytest.MustSandbox(nil)
```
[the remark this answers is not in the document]
gernest: done 5400dd0. and a5f9349

gernest:
@lmars @tcard PTAL . I have addressed all the reviews

CI is broken for`TestReadmeExamples`  test , related to `RTL6c` .

ably/realtime_conn.go
gernest: done at 80f6c2a and 8058306

lmars (approved):

tcard (approved):

PR #174, opened 2020-07-24
proto: Decode ms durations as a time.Duration wrapper.

tcard:
This way, we don't have to worry about conversions, units, etc. when
using those values.

### DIFF
--- a/ably/proto/types.go
+++ b/ably/proto/types.go
@@ -43,7 +43,7 @@ var _ interface {
 } = (*DurationFromMsecs)(nil)
 
 func (t DurationFromMsecs) asMsecs() int64 {
-	return int64(time.Duration(t) / time.Millisecond)
+	return time.Duration(t).Milliseconds()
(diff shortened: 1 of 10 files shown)


ably/proto/types.go
```diff
@@ -24,3 +31,45 @@ type Data struct {
...
+type DurationFromMsecs time.Duration
+
+var _ interface {
+	json.Marshaler
+	json.Unmarshaler
+	codec.Selfer
+} = (*DurationFromMsecs)(nil)
+
+func (t DurationFromMsecs) asMsecs() int64 {
+	return int64(time.Duration(t) / time.Millisecond)
```
gernest, line 46: lets use Duration method provided by stdlib

```suggestion
	return time.Duration(t).Milliseconds()
```

tcard:
Merging even with a failing build due to an unrelated bug fixed at #177.

PR #175, opened 2020-07-24
Skip nil options

gernest:
Options can be `nil` which may result in panics. This ensures we only apply non nil options.


tcard:
Well, it seems we stepped on each other's shoes a bit.

There's no reason you should pass a nil option. We can't be adding around nil checks randomly.

The fact that we're doing it now is just an incomplete API migration. See #173.

gernest:
🤦 sorry about this. I think github was having issues. Your patch resolves the issue.

PR #176, opened 2020-07-31
Have a single Travis build as originally intended.

tcard:
We're running two parallel builds, one with GOMAXPROCS=4, the
other with GO111MODULE=on. What we want is a single build with
those two env vars set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ go:
 before_install:
   - go get -t -v ./ably/...
 env:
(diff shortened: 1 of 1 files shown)


tcard:
I'm taking the liberty of merging this without an approval since it's so small, improves the development process, can be easily and painlessly reverted or adjusted and doesn't affect end users.

tcard:
(But @gernest please do take a look when you can!)

Issue #160, opened 2020-04-28
Add enough logs to troubleshoot problems


tcard:
In particular, every operation should have at least a "begin" and "end" log message, with parameters and outcome, plus a log message per relevant branch inside the operation.

As to the levels, if there isn't any other guideline elsewhere, I'd suggest:

* INFO for connection and channel lifetime events, and ENTER and LEAVE presence operations.
* WARNING for automatically recovered errors without state loss, such as token expirations or transient disconnection.
* ERROR for other errors, such as SUSPENDED or FAILED events, or failed requests.
* VERBOSE for fine-grained branch tracking inside operations, and incoming/outgoing protocol messages and requests with full parameters and results.

PR #179, opened 2020-07-31
RTN15h*: Handle incoming DISCONNECTED while CONNECTED

tcard:
Implements [RTN15h*](https://docs.ably.io/client-lib-development-guide/features/#RTN15h).

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -579,10 +579,14 @@ func (c *Connection) eventloop() {
 			c.state.Unlock()
 			c.queue.Flush()
 		case proto.ActionDisconnected:
-			c.state.Lock()
-			c.id = ""
-			c.setState(StateConnDisconnected, nil)
-			c.state.Unlock()
+			if !isTokenError(msg.Error) {
+				// The spec doesn't say what to do in this case, so do nothing.
+				// Ably is supposed to then close the transport, which will
+				// trigger a transition to DISCONNECTED.
+				continue
+			}
+
+			// TODO: RTN15h
 		case proto.ActionClosed:
 			c.state.Lock()
 			c.id, c.key = "", "" //(RTN16c)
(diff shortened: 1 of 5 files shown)


ably/realtime_conn.go
```diff
@@ -559,6 +566,7 @@ func (c *Connection) eventloop() {
 			if reconnecting {
 				// reset the mode
 				c.reconnecting = false
+				c.reauthorizing = true
```
gernest, line 569:
```suggestion
				c.reauthorizing = false
```

ably/realtime_conn_spec_test.go
```diff
@@ -971,6 +972,268 @@ func TestRealtimeConn_RTN15c4(t *testing.T) {
...
+		ConnectionID:      "new-connection-id",
+		ConnectionDetails: &proto.ConnectionDetails{},
+	}
+
+	// Expect a UPDATED event.
+
+	var change ably.ConnectionStateChange
+	ablytest.Instantly.Recv(t, &change, stateChanges, t.Fatalf)
+
+	if change.Event != ably.ConnectionEventUpdated {
```
gernest, line 1229: The test is failing with a different event here

```
--- FAIL: TestRealtimeConn_RTN15h2_Success (0.00s)
    realtime_conn_spec_test.go:1230: expected UPDATED event; got {DISCONNECTED DISCONNECTED CONNECTED [ErrorInfo :timeout code=80003 statusCode=0] See https://help.ably.io/error/80003}
```

tcard: This and similar intermittent errors should've been fixed by 324acf8.

ably/realtime_conn.go
```diff
@@ -519,86 +527,92 @@ func (c *Connection) eventloop() {
...
 			}
-			id := c.id
+			previousID := c.id
+			c.id = msg.ConnectionID
+			c.msgSerial = 0
+			if reconnecting && mode == recoveryMode {
+				// we are setting msgSerial as per (RTN16f)
+				msgSerial, err := strconv.ParseInt(strings.Split(c.opts.Recover, ":")[2], 10, 64)
+				if err != nil {
+					//TODO: how to handle this? Panic?
```
gernest, line 574: We will need to figure out our logging situation at some point. I think implementing the spec is more important at the moment, so its fine thisTODO stays.

tcard: I just moved this code around, the TODO was already there.

Anyway, I agree we should improve logging (see #160) once the codebase becomes less of a moving target.

In this particular case, though, I think a panic is in order. A malformed msgSerial just shouldn't ever happen, and it's already panicking if it doesn't have at least two parts separated by `:`.

gernest: I see, if that is a case you can just as well add a panic here as part of this PR. We don't need to keep that TODO anymore.

PR #183, opened 2020-08-13
Fix test for Heartbeat

gernest:
This rewrites test for heartbeat to use small maxIdleInterval and also
cover test for  RTN23b

### DIFF
--- a/ably/realtime_conn_spec_test.go
+++ b/ably/realtime_conn_spec_test.go
@@ -1384,7 +1384,7 @@ func TestRealtimeConn_RTN23(t *testing.T) {
 	// The connection should not disconnect as we received the heartbeat message
 	ablytest.Before(receiveTimeout).NoRecv(t, nil, disconnected, t.Fatalf)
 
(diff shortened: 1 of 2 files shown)


ably/realtime_conn_spec_test.go
```diff
@@ -1333,63 +1333,64 @@ func sameConnection(a, b string) bool {
...
-			if msg.Action != proto.ActionHeartbeat {
-				t.Errorf("expected action %v got %v", proto.ActionHeartbeat, msg.Action)
-			}
-		case <-time.After(receiveTimeout):
-			t.Error("no heartbeat was sent by ably")
-		}
+	// The connection should not disconnect as we received the heartbeat message
+	ablytest.Before(receiveTimeout).NoRecv(t, nil, disconnected, t.Fatalf)
+
+	// RTN23b The connection should be disconnected due to lack of activity past
```
tcard, line 1387:
```suggestion
	// RTN23a The connection should be disconnected due to lack of activity past
```

PR #185, opened 2020-08-16
RTL13: Handle DETACHED while not DETACHING.

tcard:
[RTL13](https://docs.ably.io/client-lib-development-guide/features/#RTL13)

Missing SUSPENDED, which isn't implemented.

### DIFF
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -198,10 +198,14 @@ func (c *RealtimeChannel) mayAttach(result, checkActive bool) (Result, error) {
 			return nopResult, nil
 		}
 	}
+	return c.lockAttach(result, nil)
+}
+
+func (c *RealtimeChannel) lockAttach(result bool, err error) (Result, error) {
 	if !c.client.Connection.lockIsActive() {
 		return nil, c.state.set(StateChanFailed, errAttach)
 	}
-	c.state.set(StateChanAttaching, nil)
+	c.state.set(StateChanAttaching, err)
 	var res Result
 	if result {
 		res = c.state.listenResult(attachResultStates...)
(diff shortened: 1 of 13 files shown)


ably/realtime_channel.go
```diff
@@ -411,6 +452,45 @@ func (c *RealtimeChannel) notify(msg *proto.ProtocolMessage) {
...
+		defer c.state.off(stateChange)
+
+		for !c.retryAttach(stateChange) {
+		}
+	}()
+}
+
+func (c *RealtimeChannel) retryAttach(stateChange chan State) (done bool) {
+	select {
+	case <-c.opts().After(c.opts().ChannelRetryTimeout):
```
gernest, line 474: `c.opts().After` is a callback. There is no default value for this set. It is passing in your test because you are explicitly setting this , this part will panic at runtime.

We should also avoid `time.After` according to docs https://pkg.go.dev/time?tab=doc#After  it might result into leaks in the case we receive  any concurrent changes that cancels retry.

ably/state.go
```diff
@@ -219,7 +219,7 @@ func (s *stateEmitter) set(state StateEnum, err error) error {
 	changed := s.current != state
 	s.current = state
 	s.err = stateError(state, err)
-	if changed {
+	if changed || err != nil {
```
gernest, line 222: Why do we need to check `err!=nil`?  aren't we interested on state change only? We are already passing err on `s.err`

gernest (changes requested):
Looks good, I left a few comments.

ably/realtime_channel_spec_test.go
```diff
@@ -0,0 +1,265 @@
...
+			Action:  proto.ActionDetached,
+			Channel: channel.Name,
+			Error:   &errInfo,
+		}
+
+		// Expect a state change with the error.
+
+		var change ably.State
+		ablytest.Instantly.Recv(t, &change, stateChanges, t.Fatalf)
+		if expected, got := ably.StateChanDetached, change.State; expected != got {
```
sacOO7, line 162: @tcard , this should be `ChannelStateSusended` as per [spec](https://docs.ably.io/client-lib-development-guide/features/#RTL13b)
I was looking into RTL13 because it's a bit related to channel attach

sacOO7: Just to add, if we change it to as per spec, we will lose one of the next state for transitioner, from
`ATTACHING` to `DETACHED`, so we will have to update transitioner, it's working for now

PR #186, opened 2020-08-21
RTN15f: Discard in-flight publishes on transport loss.

tcard:
[RTN15f](https://docs.ably.io/client-lib-development-guide/features/#RTN15f)

> Therefore, once a transport drops, the client library must either fail the publish attempt, or re-attempt by re-sending the messages on a new transport if the resume was successful

We go with the "fail the publish attempt" option.

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -500,7 +500,11 @@ func (c *Connection) setSerial(serial int64) {
 	c.serial = serial
 }
 
-func (c *Connection) eventloop() {
+func (c *Connection) eventloop() (err error) {
+	defer func() {
+		c.forciblyNackPendingMessages(err)
+	}()
+
 	var lastActivityAt time.Time
 	var connDetails *proto.ConnectionDetails
 	for c.lockCanReceiveMessages() {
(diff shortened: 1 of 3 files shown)


ably/realtime_conn.go
```diff
@@ -500,7 +500,11 @@ func (c *Connection) setSerial(serial int64) {
 	c.serial = serial
 }
 
-func (c *Connection) eventloop() {
+func (c *Connection) eventloop() (err error) {
+	defer func() {
+		c.forciblyNackPendingMessages(err)
```
gernest, line 505: [RTN19](https://docs.ably.io/client-lib-development-guide/features/#RTN19)
>Any ProtocolMessage that is awaiting an ACK/NACK on the old transport will not receive the ACK/NACK on the new transport. The client library must therefore resend any ProtocolMessage that is awaiting a ACK/NACK to Ably in order to receive the expected ACK/NACK for that message. The Ably service is responsible for keeping track of messages, ignoring duplicates and responding with suitable ACK/NACK messages

If we nack them now, how will we be able to re send them? I got stuck a bit implementing `RTN19` thinking of a better way to test it.

gernest (changes requested):
Left a comment, will take another look after it is addressed.

ably/realtime_conn.go
tcard: Hmm, then I'd say this part of RTN15f is in conflict with RTN19: "once a transport drops, the client library must either fail the publish attempt, or re-attempt by re-sending the messages on a new transport" RTN19 makes it non-optional to re-send the messages.

Also, this makes RTN15f redundant, so I'm closing the PR.

Will open a spec change with this.

tcard:
See https://github.com/ably/ably-go/pull/186#discussion_r478928292.

ably/realtime_conn.go
tcard: See https://github.com/ably/docs/pull/948.

PR #188, opened 2020-08-21
RTN15e: Check that Connection.Key changes on reconnections.

tcard:
[RTN15e](https://docs.ably.io/client-lib-development-guide/features/#RTN15e)


gernest (approved):
LGTM

PR #190, opened 2020-08-24
RTL2: EventEmitter for channel events; remove old State

tcard:
[RTL2](https://docs.ably.io/client-lib-development-guide/features/#RTL2). This also removes the old State machinery, including the temporary workarounds for bridging the two.

### DIFF
--- a/ably/ablytest/ablytest.go
+++ b/ably/ablytest/ablytest.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"reflect"
 	"strconv"
 	"time"
 
@@ -74,4 +75,20 @@ func protocol(opts ably.ClientOptions) string {
 		return "application/x-msgpack"
 	}
 	return "application/json"
+}
+
+func Contains(s, sub interface{}) bool {
+	return reflectContains(reflect.ValueOf(s), reflect.ValueOf(sub))
+}
+
+func reflectContains(s, sub reflect.Value) bool {
+	for i := 0; i+sub.Len() <= s.Len(); i++ {
+		if reflect.DeepEqual(
+			s.Slice(i, i+sub.Len()).Interface(),
+			sub.Interface(),
+		) {
+			return true
+		}
+	}
+	return false
 }
--- a/ably/ablytest/recorders.go
+++ b/ably/ablytest/recorders.go
@@ -4,13 +4,11 @@ import (
 	"bytes"
 	"context"
 	"errors"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
-	"reflect"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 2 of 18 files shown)


ably/realtime_channel.go
```diff
@@ -110,61 +110,67 @@ func (ch *Channels) All() []*RealtimeChannel {
...
 type RealtimeChannel struct {
+	mtx sync.Mutex
+
+	ChannelEventEmitter
 	Name     string            // name used to create the channel
 	Presence *RealtimePresence //
 
+	state           ChannelState
+	errorReason     *ErrorInfo
+	internalEmitter ChannelEventEmitter
```
gernest, line 143: Isn't this redundant? We are already embedding `ChannelEventEmitter`. Which we only use once  in `RealtimeChannel.lockSetState` in which we emit the same event twice. I think we can just use one emitter per `RealtimeChannel` to simplify things.

ably/realtime_channel.go
```diff
@@ -455,22 +474,22 @@ func (c *RealtimeChannel) notify(msg *proto.ProtocolMessage) {
...
 func (c *RealtimeChannel) lockStartRetryAttachLoop(err error) {
 	// TODO: Move to SUSPENDED; move it to DETACHED for now.
-	c.state.set(StateChanDetached, err)
-	c.state.Unlock()
+	c.lockSetState(ChannelStateDetached, err)
+	c.mtx.Unlock()
 
 	go func() {
 		// TODO: The listener should be set before unlocking the state.
 		// Will do once we have an EventEmitter.
```
gernest, line 482: Why do we still have this TODO when we already have event emitter?

ably/realtime_channel.go
```diff
@@ -308,23 +286,64 @@ func (c *RealtimeChannel) Unsubscribe(sub *Subscription, names ...string) {
...
+	return em.emitter.OnAll(func(change emitterData) {
+		handle(change.(ChannelStateChange))
+	})
+}
+
+// Once registers an one-off event handler for connection events of a specific kind.
+//
+// See package-level documentation on Event Emitter for details.
+func (em ChannelEventEmitter) Once(e ChannelEvent, handle func(ChannelStateChange)) (off func()) {
+	return em.emitter.On(e, func(change emitterData) {
```
gernest, line 321:
```suggestion
	return em.emitter.Once(e, func(change emitterData) {
```

ably/realtime_conn.go
```diff
@@ -320,86 +320,76 @@ func (c *Connection) RecoveryKey() string {
...
 	})
 }
 
 // Once registers an one-off event handler for connection events of a specific kind.
 //
 // See package-level documentation on Event Emitter for details.
-func (c *Connection) Once(e ConnectionEvent, handle func(ConnectionStateChange)) (off func()) {
-	return c.state.eventEmitter.On(e, func(change emitterData) {
+func (em ConnectionEventEmitter) Once(e ConnectionEvent, handle func(ConnectionStateChange)) (off func()) {
+	return em.emitter.On(e, func(change emitterData) {
```
gernest, line 367:
```suggestion
	return em.emitter.Once(e, func(change emitterData) {
```

ably/realtime_conn.go
```diff
@@ -31,14 +32,21 @@ const (
 // Connection represents a single connection Realtime instantiates for
 // communication with Ably servers.
 type Connection struct {
+	mtx sync.Mutex
+
+	ConnectionEventEmitter
+
+	state           ConnectionState
+	errorReason     *ErrorInfo
+	internalEmitter ConnectionEventEmitter
```
gernest, line 41: redundant? We should just use the embedded `ConnectionEventEmitter` so we can have a single emitter per `Connection`.

gernest (changes requested):
fantastic! I left a few comments, after they are addressed I will take another look.

ably/ablytest/recorders.go
```diff
@@ -125,168 +123,78 @@ func (rec *RoundTripRecorder) roundTrip(req *http.Request) (*http.Response, erro
...
-		select {
-		case <-rec.ch:
-		default:
-			return
-		}
-	}
+func (cs *ChanStatesRecorder) States() []ably.ChannelState {
+	cs.mtx.Lock()
+	defer cs.mtx.Unlock()
+	return cs.states
```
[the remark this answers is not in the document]
tcard: This isn't really necessary, because we're only appending to and reading from the slice concurrently.

An append will either write on the slice's underlying array past the previous last element (so the elements that can be read through the old slice won't be modified), or create a new underlying array and copy the old one (so the old slice will just be read for copying, not modified; and concurrent read is safe).

We only need a lock on the slice variable we're reassigning (in this case, the struct field), not on the slice itself.

ably/ablytest/recorders.go
```diff
@@ -125,168 +123,78 @@ func (rec *RoundTripRecorder) roundTrip(req *http.Request) (*http.Response, erro
...
 
-func (rec *StateRecorder) timeout() time.Duration {
-	if rec.Timeout != 0 {
-		return rec.Timeout
-	}
-	return 15 * time.Second
+func (ce *ConnErrorsRecorder) Errors() []*ably.ErrorInfo {
+	ce.mtx.Lock()
+	defer ce.mtx.Unlock()
+	return ce.errors
```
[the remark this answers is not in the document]
tcard: See https://github.com/ably/ably-go/pull/190#discussion_r478963317.

ably/realtime_channel.go
tcard: Fixed at e44ae34.

ably/realtime_channel.go
tcard: The thing is we want separate emitters for users and for the internals of the library. We don't want them to affect each other in any way. Specifically, we don't want [`off` called with no listener](https://docs.ably.io/client-lib-development-guide/features/#RTE5) (which deregisters all listeners) to affect internal listeners.

(I'm now realizing we're missing the `Off` and `OfAll` methods with no listeners. Will add in a new PR.)

tcard: This behavior is specified in [RTC10](https://docs.ably.io/client-lib-development-guide/features/#RTC10): "The client library should never register any listeners for internal use with the public EventEmitter interfaces"

ably/realtime_conn.go
tcard: See https://github.com/ably/ably-go/pull/190#discussion_r478996084.

gernest (changes requested):
Looks good, I left a comment. 

Also for some reason a test is consistently failing on my machine.

```
$ go test -run TestRealtime_RealtimeHost ./ably/
--- FAIL: TestRealtime_RealtimeHost (10.37s)
    realtime_client_test.go:80: want len(errors)=3; got 2
FAIL
FAIL    github.com/ably/ably-go/ably    10.389s
FAIL
```

ably/ablytest/ablytest.go
```diff
@@ -75,3 +76,19 @@ func protocol(opts ably.ClientOptions) string {
...
+	for i := 0; i+sub.Len() <= s.Len(); i++ {
+		if reflect.DeepEqual(
+			s.Slice(i, i+sub.Len()).Interface(),
+			sub.Interface(),
+		) {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
tcard: It's used [here](https://github.com/ably/ably-go/pull/190/files#diff-7dbcbdbc9b0996e1f6ab7060fa1e264dR87) and in a couple other places.

tcard:
> Also for some reason a test is consistently failing on my machine.
> 
> ```
> $ go test -run TestRealtime_RealtimeHost ./ably/
> --- FAIL: TestRealtime_RealtimeHost (10.37s)
>     realtime_client_test.go:80: want len(errors)=3; got 2
> FAIL
> FAIL    github.com/ably/ably-go/ably    10.389s
> FAIL
> ```

True, this was racy. Should be fixed at 4ab3b27.

gernest:
@tcard please try to check this test too, its failing 

```
$ go test -run TestAuth_ClientID ./ably/
--- FAIL: TestAuth_ClientID (14.38s)
    auth_test.go:708: wanted to have ConnectionStateFailed emitted; it wasn't
FAIL
FAIL    github.com/ably/ably-go/ably    14.397s
FAIL
```

I wonder why ci is green

tcard:
@gernest Should be fixed by f75839f.

> I wonder why ci is green

It also passes consistently in my machine. When a test is racy, who knows which conditions trigger which states...

gernest (approved):
LGTM

PR #191, opened 2020-08-24
Ad-hoc fix for enqueuing attach attempt when DISCONNECTED.

### LINKED ISSUE #189: Channel attach fails in the DISCONNECTED state
Channel attach operations when in a `DISCONNECTED` state fail with an error: `attempted to attach channel to inactive connection`.

tcard:
To be converted to a proper spec-oriented test when ported to the 1.2 branch.

Fixes #189.

### DIFF
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -170,14 +170,14 @@ func (c *RealtimeChannel) mayAttach(result, checkActive bool) (Result, error) {
 
 	switch c.client.Connection.State() {
 	// RTL4b
-	case StateConnClosed,
+	case StateConnInitialized,
+		StateConnClosed,
 		StateConnClosing,
 		StateConnFailed:
 		return nil, newError(80000, errAttach)
 
 	// RTL4i
-	case StateConnInitialized,
-		StateConnConnecting,
+	case StateConnConnecting,
 		StateConnDisconnected:
 
 		return goWaiter(func() error {
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -191,7 +191,10 @@ func (c *RealtimeChannel) mayAttach(result, checkActive bool) (Result, error) {
 				return err
 			}
 
-			return res.Wait()
+			if result {
+				return res.Wait()
+			}
+			return nil
 		}), nil
 	}
 
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -182,9 +182,12 @@ func (c *RealtimeChannel) mayAttach(result, checkActive bool) (Result, error) {
 
 		return goWaiter(func() error {
 			connected := make(chan State, 1)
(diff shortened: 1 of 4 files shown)


ably/realtime_channel.go
```diff
@@ -167,9 +167,37 @@ func (c *RealtimeChannel) mayAttach(result, checkActive bool) (Result, error) {
...
+
+	// RTL4i
+	case StateConnInitialized,
+		StateConnConnecting,
+		StateConnDisconnected:
+
+		return goWaiter(func() error {
+			connected := make(chan State, 1)
+			c.client.Connection.On(connected, StateConnConnected)
+			<-connected
```
gernest, line 186: This will block forever when `StateConnConnected` isn't fired.

ably/realtime_channel.go
```diff
@@ -167,9 +167,37 @@ func (c *RealtimeChannel) mayAttach(result, checkActive bool) (Result, error) {
...
-		return nil, c.state.set(StateChanFailed, errAttach)
+
+	switch c.client.Connection.State() {
+	// RTL4b
+	case StateConnClosed,
+		StateConnClosing,
+		StateConnFailed:
+		return nil, newError(80000, errAttach)
+
+	// RTL4i
```
gernest, line 178: can't find this on the spec.

gernest (changes requested):
Left a comment.

Also can you point me on the spec on expected behaviour for this scenario?

ably/realtime_channel.go
tcard: Good catch! Fixed at 71e958c.

ably/realtime_channel.go
tcard: [RTL4i](https://docs.ably.io/client-lib-development-guide/features/#RTL4i)

ably/realtime_channel.go
```diff
@@ -167,9 +167,40 @@ func (c *RealtimeChannel) mayAttach(result, checkActive bool) (Result, error) {
...
+	case StateConnInitialized,
```
gernest, line 179: >If the connection state is CONNECTING or DISCONNECTED, do the operation once the connection state is CONNECTED

I don't see spec mentioning  `StateConnInitialized` why is it here?

gernest (changes requested):
Left a comment. Is it possible to port the changes to `integration/1.2` branch as well?

ably/realtime_channel.go
tcard: True, [RTL4b](https://docs.ably.io/client-lib-development-guide/features/#RTL4b) says that would be an error. Fixed at 6fb73b8. (Tests to ensure compliance will follow in the 1.2 branch.)

tcard:
> Is it possible to port the changes to `integration/1.2` branch as well?

Yes, will do once merged. We should do this with every change on master.

gernest (approved):
LGTM

PR #195, opened 2020-09-05
Always include serial values in encoded protocol messages

lmars:
Both `msgSerial` and `connectionSerial` are zero-based, but realtime does not treat the absence of these values as the same as zero, which when done for a realtime message triggers the following error:

```
Fatal protocol error: message published without msgSerial
```

This commit removes the `omitempty` tag from the serial fields so that their values are always included when they are encoded.

This behaviour was witnessed in our load testing environment where the client was setting the `msgSerial` back to zero after establishing a connection [here](https://github.com/ably/ably-go/blob/main/ably/realtime_conn.go#L473), assigning a zero `msgSerial` to the next message that was sent, and receiving a publish error.

### DIFF
--- a/ably/proto/protocol_message.go
+++ b/ably/proto/protocol_message.go
@@ -85,8 +85,8 @@ type ProtocolMessage struct {
 	ChannelSerial     string             `json:"channelSerial,omitempty" codec:"channelSerial,omitempty"`
 	ConnectionDetails *ConnectionDetails `json:"connectionDetails,omitempty" codec:"connectionDetails,omitempty"`
 	Error             *ErrorInfo         `json:"error,omitempty" codec:"error,omitempty"`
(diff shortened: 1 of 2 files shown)


mattheworiordan (approved):

mattheworiordan:
Nice spot, perhaps would have been nice to have a test to prevent a regression in future though, given how hard this was to find.

lmars:
@mattheworiordan test added in e6b2c5a.

QuintinWillison:
Thanks, @lmars ... I'll add @gernest and @tcard as reviewers, given they're actively maintaining this repository right now.

tcard (approved):
LGTM

tcard:
Ported to `integration/1.2` as part of #196.

lmars:
@tcard it doesn't look like this change was included in #196, should it have been?

tcard:
@lmars Oops, true, had a race condition there. It's now on `integration/1.2` at b21a502.

PR #200, opened 2020-10-07
Explicitly annotate enum vars so that godoc places them with the type

tcard:
Turns out, if you left a top-level variable implicitly typed, it goes to the generic "Variables" section instead of to its type's section.

We go from this:

Screenshot 2020-10-07 at 12 02 35

To this:

Screenshot 2020-10-07 at 12 03 21

Screenshot 2020-10-07 at 12 03 07

### DIFF
--- a/ably/state.go
+++ b/ably/state.go
@@ -326,14 +326,14 @@ type ConnectionState struct {
 }
 
 var (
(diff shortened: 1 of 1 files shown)


tcard:
@gernest I'm merging this without review, since it seems uncontroversial and I'd rather not block for this. Please do add your comments if any and I'll address feedback in separate PRs.

PR #201, opened 2020-10-07
Remove things from public API that aren't in the spec

tcard:
Picking some low-hanging fruit for sticking to the spec.


gernest (approved):
Apologies for doing this late.

LGTM

Issue #197, opened 2020-09-11
v1.2 API design


tcard:
This issue tracks and discuss the changes to be made to the public API in order to make it [1.2 spec-compliant](https://docs.ably.io/client-lib-development-guide/features/#idl).

### Subscriptions: from channels to callbacks

For some features, we make a long-lived subscription to a series of events that the caller receives. The spec says those events are delivered by calling a callback. In Go, instead, we were sending them to channels, from which user code receive events at their own. For 1.2 we're taking callbacks. [See the relevant discussion](https://github.com/ably/ably-go/pull/144).

So we go from something like:

```go
ch := make(chan ably.ChannelStateChange)
channel.On(ably.ChannelStateChange, ably.ConnectionEventAttached)
for change := range ch {
    // ...
}
```

To:

```go
channel.On(ably.ConnectionEventAttached, func(change ably.ChannelStateChange) {
    // ...
})
```

The list of such features is:

- [x] Connection events (#144)
- [x] Channel events (#190)
- [ ] Messages
- [ ] Pressence messages

### Options

We've decided to implement the classes from the spec that define options [as functional options instead](https://github.com/ably/ably-go/pull/141#discussion_r401739264).

Each option is defined as a package-level function, with a `With` prefix. Additionally, except `ClientOptions`, each function has first the name of what they're options for as prefix (e. g. `RestChannel.publish(params?)` → `PublishBatchWithParams`). ([Discussion below.](https://github.com/ably/ably-go/issues/197#issuecomment-710844145))

- [x] ClientOptions and AuthOptions (#145)
  - [ ] Converted to package-level functions.
- [x] ChannelOptions (#146)
  - [ ] Converted to package-level functions.

### Optional arguments

Optional arguments are translated as functional options, as described above. ([Discussion below.](https://github.com/ably/ably-go/issues/197#issuecomment-705526229))

If an optional argument is itself an options aggregate per the spec, the options "collapse" into a single level.

### `io` actions: from `Result` to blocking calls

The spec marks with a `io` annotation values that result from making an IO operation (to the network, typically). In some environments (JavaScript, Cocoa) those are implemented by passing them to a callback or an equivalent Promise-style mechanism. In others, it's just a normal function call that blocks until the operation is complete and returns the value. If the user wants to do more than one thing at once, they use threads or some equivalent.

[...]

PR #202, opened 2020-10-14
v1.2-compliant message and presence publish and subscribe

tcard:
Missing tests, but can be reviewed.


ably/realtime_channel.go
```diff
@@ -383,32 +421,41 @@ func (em ChannelEventEmitter) OffAll() {
...
 // PublishAll publishes all given messages on the channel at once.
-// PublishAll does not block.
 //
-// This implicitly attaches the channel if it's not already attached.
-func (c *RealtimeChannel) PublishAll(messages []*proto.Message) (Result, error) {
+// This implicitly attaches the channel if it's not already attached. If the
+// context is canceled before the attach operation finishes, the call
+// returns with an error, but the operation carries on in the background and
+// the channel may eventually be attached and the message published anyway.
+func (c *RealtimeChannel) PublishAll(ctx context.Context, messages []*Message) error {
```
lmars, line 440: I think this should be called `PublishBatch` as it's publishing a batch of messages (and we use the term "batch" in the spec when referring to the publish of multiple messages).

ably/rest_channel.go
```diff
@@ -46,17 +49,26 @@ func newRestChannel(name string, client *REST) *RestChannel {
...
-// using the Rest API.
-func (c *RestChannel) PublishAll(messages []*proto.Message) error {
+// Message is what Ably channels send and receive.
+type Message = proto.Message
+
+// PublishAll publishes multiple messages in a batch.
+//
+// The params, if any, will be set as additional query parameters in the
+// resulting HTTP request to the REST API.
+func (c *RESTChannel) PublishAll(ctx context.Context, messages []*Message, params map[string]string) error {
```
lmars, line 66: Given `params` is optional, then as per https://github.com/ably/ably-go/issues/197#issuecomment-705526229 should there be a separate `PublishAllWithOptions` (or `PublishBatchWithOptions`), and a `WithPublishParam` option?

ably/realtime_channel.go
tcard: Sounds better. Done at 8f632a1.

ably/rest_channel.go
tcard: Sure, in fact this is what prompted the discussion but then I forgot to actually do it before submitting the PR. Done at d987390.

ably/rest_channel.go
```diff
@@ -46,17 +49,46 @@ func newRestChannel(name string, client *REST) *RestChannel {
...
-func (c *RestChannel) PublishAll(messages []*proto.Message) error {
+// Params adds query parameters to the resulting HTTP request to the REST API.
+func (o RESTChannelPublishBatchOptions) Params(params map[string]string) RESTChannelPublishBatchOptions {
+	return append(o, func(options *restChannelPublishBatchOptions) {
+		options.params = params
+	})
+}
+
+// PublishBatchWithOptions is PublishBatch with optional parameters.
+func (c *RESTChannel) PublishBatchWithOptions(ctx context.Context, messages []*Message, options RESTChannelPublishBatchOptions) error {
```
lmars, line 82: Why not use a variadic list of functional options rather that a struct?

With the current code, if I want to publish a batch with some params, I'd write:

```go
channel.PublishBatchWithOptions(
	ctx,
	msgs,
	ably.RESTChannelPublishBatchOptions{}.Params(map[string]string{"foo": "bar"}),
)
```

whereas with a variadic list it would be:

```go
channel.PublishBatchWithOptions(
	ctx,
	msgs,
	ably.PublishBatchParams(map[string]string{"foo": "bar"}),
)
```

(I also think there should be a singular form):

```go
channel.PublishBatchWithOptions(
	ctx,
	msgs,
	ably.PublishBatchParam("foo", "bar"),
)
```

Do you not find this to be a cleaner API? It's also inline with how this is often implemented in other popular API client libraries (e.g. [oauth2](https://godoc.org/golang.org/x/oauth2#Config.Exchange), [grpc](https://godoc.org/google.golang.org/grpc#ClientConn.Invoke) and [AWS SDK](https://docs.aws.amazon.com/sdk-for-go/api/service/sts/#STS.GetCallerIdentityWithContext)).

tcard: ClientOptions is already done like this, and I just followed the pattern here. ([We discussed this back then.](https://github.com/ably/ably-go/pull/141#discussion_r401739264))

I still think that's the right option for ClientOptions, given there's tons of them. But for optional parameters, of which there shouldn't be too many, maybe it's worth polluting the package-level namespace to avoid the `ably.FooOptions{}.` noise.

I agree that the `RESTChannelPublishBatch` prefix is probably too much and just `PublishBatch` would be preferahyble.

lmars: > ClientOptions is already done like this

My apologies, it looks like I made a similar comment [here](https://github.com/ably/ably-go/pull/145#discussion_r405653434) but failed to follow up, and I also made a similar comment more recently [here](https://github.com/ably/ably-go/issues/197#issuecomment-709661794), but I'll now make a more explicit comment in #197 so we can discuss this further and gather more feedback.

lmars: I've continued this discussion here: https://github.com/ably/ably-go/issues/197#issuecomment-710844145

ably/rest_channel.go
lmars: Great 👍

ably/rest_channel.go
tcard: Done at 7f31d1c.

tcard:
I wanted to add some spec tests here before merging, but given they would then conflict with #205, and we already have plenty of tests against the affected APIs, I think this is ready for merge and we can add the spec tests later, when both PRs are merged.

@lmars PTAL

lmars (approved):

ably/realtime_channel.go
tcard: This name is being superseded by https://github.com/ably/ably-go/pull/303. @lmars PTAL

PR #205, opened 2020-10-21
Implement options as package-level functions

tcard:
See https://github.com/ably/ably-go/issues/197#issuecomment-710844145


ably/options.go
```diff
@@ -397,265 +396,265 @@ type TokenString string
 func (TokenString) IsTokener() {}
 func (TokenString) isTokener() {}
 
-func (os AuthOptions) AuthCallback(authCallback func(context.Context, TokenParams) (Tokener, error)) AuthOptions {
-	return append(os, func(os *authOptions) {
+func AuthWithAuthCallback(authCallback func(context.Context, TokenParams) (Tokener, error)) AuthOption {
```
tcard, line 399: This `AuthWithAuth...` naming is what we get by applying #197 strictly, but it's ugly. Should we just go with `AuthWithCallback`? It seems very unlikely this would cause conflicts in the future.

gernest: @tcard I think `AuthWithCallback` is better suited here.

lmars: `ably.AuthWithCallback` 👍

lmars (approved):

ably/options.go
tcard: Done at 5cf7398cb3f3cfe2624ab2a39c0e1b442fee1f8a.

PR #206, opened 2020-10-28
Update README for 1.2 preview, with associated fixes

tcard:
Also, I was anxious about fixing README without actually checking that the fixes are correct, so I added script to generate a Go test from it.

[Rendered README](https://github.com/ably/ably-go/blob/feature/1.2-preview-readme/README.md)


README.md
```diff
@@ -255,18 +253,20 @@ if err != nil {
...
+-->
+
 ## Known limitations (work in progress)
 
 As the library is actively developed couple of features are not there yet:
 
-- Realtime connection recovery is not implemented
-- Realtime connection failure handling is not implemented
-- ChannelsOptions and CipherParams are not supported when creating a Channel
+- Realtime connection failure handling is partly implemented
```
tcard, line 264:
```suggestion
- Realtime connection failure handling is partially implemented
```

README.md
```diff
@@ -209,8 +214,8 @@ if err != nil {
 ### Presence on a channel
 
 ```go
-page, err := channel.Presence.Get(nil)
-for ; err == nil; page, err = page.Next() {
+page, err = channel.Presence.Get(nil)
+for ; err == nil && page != nil; page, err = page.Next() {
```
lmars, line 218: This is the first time I've seen how this API is used, so perhaps this isn't the correct place to discuss, but I'll put it here anyway since I've seen it.

The usage feels a little awkward, I wonder if something like the following would be better:

```go
params := &ably.PaginateParams{
	Limit: 10,
}
err := channel.Presence.Get(params, func(page *ably.PaginatedResult) bool {
	// iterate page.PresenceMessages()
	// return false to stop iterating
})
```

and similarly for the other paginated operations.

README.md
```diff
@@ -234,18 +239,11 @@ if err != nil {
 }
 ```
 
-### Generate Token and Token Request
-
-```go
-client.Auth.RequestToken()
-client.Auth.CreateTokenRequest()
```
lmars, line 241: Are these no longer relevant?

README.md
```diff
@@ -255,18 +253,20 @@ if err != nil {
...
+-->
+
 ## Known limitations (work in progress)
 
 As the library is actively developed couple of features are not there yet:
 
-- Realtime connection recovery is not implemented
-- Realtime connection failure handling is not implemented
-- ChannelsOptions and CipherParams are not supported when creating a Channel
+- Realtime connection failure handling is partially implemented
```
lmars, line 264: This is surprising, what is left to implement? I think this is one of the main features we're aiming to give to customers in this release.

README.md
```diff
@@ -255,18 +253,20 @@ if err != nil {
...
 - Realtime Ping function is not implemented
 
 ## Release process
 
-This library uses [semantic versioning](http://semver.org/). For each release, the following needs to be done:
+Starting with release 1.2, this library uses [semantic versioning](http://semver.org/). For each release, the following needs to be done:
```
lmars, line 269: The examples use `1.1.x`, but this says semver only started with `1.2.x`, so should the examples reference `1.2.x`, or was semver already in use for `1.1.x`?

README.md
```diff
@@ -14,7 +14,7 @@ This library implements the Ably REST and Realtime client APIs.
 
 ### REST API
 
-In respect of the Ably REST API, this library targets the Ably 1.1 client library specification,
+In respect of the Ably REST API, this library targets the Ably 1.2 client library specification,
```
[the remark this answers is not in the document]
tcard: > Should that also say 1.2?

True. Fixed at e9b274a.



> Also, why does the library need to be upgraded incrementally? Are we saying users have to upgrade to 1.1, then to 1.2?

As I understand it, this means that version 1.1 (now, 1.2) doesn't implement the whole 1.1 spec. While we haven't really decided this, I'd expect it will be the same case for 1.2. Version 1.2.0 can be released as soon as we have a partially spec-compliant version onto which 1.2 features can be added in a backwards-compatible way in further minor releases.

README.md
```diff
@@ -28,9 +28,6 @@ releases. If there are features that are currently missing that are a high prior
 
 ### Realtime API
 
-In respect of the Realtime API, this is an early experimental implementation that targets the (now superseded) 0.8
-library specification. This means that there are significant shortfalls in functionality; the principal issues are:
-
 - there is no channel `suspended` state; this means that the client will not automatically reattach to channels if a
```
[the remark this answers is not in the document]
tcard: Moved and reworded at 03b766c.

README.md
tcard: Note this API isn't covered by the 1.2 preview subset and is likely to change. I just fixed it here (previous version returned an error on end of iteration, so this example would always panic).

I'm moving this discussion at https://github.com/ably/ably-go/issues/197#issuecomment-718446264 if that's OK.

README.md
tcard: Those weren't really usage examples. They lack arguments and don't use the return values.

Also, they arguably aren't relevant enough for most Go users, who are likely to prefer key auth given Go runs (in the overwhelming majority of cases) on private servers. So they're beyond the scope of an introductory README.

README.md
tcard: Once #172 is merged this line can be removed, but the API preview release that will result from _this_ PR won't have that (as things stand now).

README.md
tcard: Using 1.2.0 now as an example at df19c8b.

README.md
```diff
@@ -41,98 +38,98 @@ automatically re-entered;
...
 
+<!-- GO IMPORT "context" -->
+
+<!-- GO EXAMPLE
+ctx := context.Background()
+-->
+
 ```go
-client, err := ably.NewRealtime(ably.NewClientOptions("xxx:xxx"))
+client, err := ably.NewRealtime(ably.WithKey("xxx:xxx"))
```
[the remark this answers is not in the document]
tcard: I'm not sure we should overwhelm the README with error paths. For a full reference, users can check the Ably docs and the Go package docs.

README.md
```diff
@@ -41,98 +38,98 @@ automatically re-entered;
...
 channel := client.Channels.Get("test")
 ```
 
-### Subscribing to a channel for all events
+### Subscribing to a channel for all messages
 
 ```go
-sub, err := channel.Subscribe()
+unsubscribe, err := channel.SubscribeAll(ctx, func(msg *ably.Message) {
+	fmt.Println("Received message:", msg)
```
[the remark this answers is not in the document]
tcard: Done at ad3bc49.

README.md
```diff
@@ -41,98 +38,98 @@ automatically re-entered;
...
 	panic(err)
 }
-
-// await confirmation
-if err = res.Wait(); err != nil {
-	panic(err)
-}
 ```
 
 ### Announcing presence on a channel
```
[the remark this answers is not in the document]
tcard: Added at ad3bc49.

README.md
```diff
@@ -186,7 +191,7 @@ channel := client.Channels.Get("test")
 ### Publishing a message to a channel
```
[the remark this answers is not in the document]
tcard: Added at ad3bc49.

README.md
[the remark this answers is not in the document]
lmars: > Version 1.2.0 can be released as soon as we have a partially spec-compliant version onto which 1.2 features can be added in a backwards-compatible way in further minor releases.

Ok, I agree with that, but I'm still confused by the sentence in the README, since it seems to suggest that the user needs to do some form of incremental upgrade (e.g. upgrade to 1.2.1, then 1.2.2, then 1.2.3 etc.), but I don't believe that to be the case.

Perhaps the section would be clearer as something like:

> This library targets the Ably 1.2 client library specification, although as of release 1.2.0 the following are not implemented and will be covered in future 1.2.x releases:
>
> [list of unimplemented things]
>
> If any of these unimplemented features are a high priority for your use-case then please contact Ably customer support. Pull Requests are also welcomed.

README.md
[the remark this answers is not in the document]
lmars: > I'm not sure we should overwhelm the README with error paths

This is a pretty important error path for users to be aware of though, no?

> For a full reference, users can check the Ably docs and the Go package docs.

Ok, can we at least put a link here to where those exist for handling connection errors?

README.md
lmars: That's fine, thanks 👍

README.md
```diff
@@ -145,37 +124,45 @@ for _, client := range clients {
 ### Subscribing to all presence messages
 
 ```go
-sub, err := channel.Presence.Subscribe()
+unsubscribe, err = channel.Presence.SubscribeAll(ctx, func(msg *ably.PresenceMessage) {
+	fmt.Println("Presence event: action=%v data=%v", msg.Action, msg.Data)
```
lmars, line 128: This should use `fmt.Printf`

README.md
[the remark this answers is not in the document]
tcard: See 1e4ff57.

README.md
tcard: Done at f2b682779470cb2b98833d3f702026984efdfe27.

README.md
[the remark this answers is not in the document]
tcard: See cbdaa10.

Those examples are how things _should_ be, but they won't work at least until #172 is merged, and they probably depend on [RTL6 too](https://github.com/ably/ably-go/pull/172#issuecomment-718666302) (which I'm working on now). But they showcase how users are likely to encounter a connection error: by trying to use the connection for publishing, subscribing or some other operation.

I can revert that commit for now so that this PR is mergeable, or let the PR blocked until those issues are fixed and then rebase.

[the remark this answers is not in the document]
lmars: > See cbdaa10.

Great thanks 👍 

> I can revert that commit for now so that this PR is mergeable, or let the PR blocked until those issues are fixed and then rebase.

I think we can merge this and just focus on merging the other PR and related fixes before we share this updated README with people.

lmars (approved):

PR #208, opened 2020-11-06
RTL6c: Publish while not connected

tcard:
- [x] [RTL6c1](https://docs.ably.io/client-lib-development-guide/features/#RTL6c1)
- [x] [RTL6c2](https://docs.ably.io/client-lib-development-guide/features/#RTL6c2)
- [x] [RTL6c4](https://docs.ably.io/client-lib-development-guide/features/#RTL6c4)
- [x] [RTL6c5](https://docs.ably.io/client-lib-development-guide/features/#RTL6c5)

This PR also includes random fixes that came up during testing, including the culprit of #172's CI failure.

### DIFF
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -180,8 +180,7 @@ func (c *RealtimeChannel) Attach(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	// TODO: Don't ignore context.
-	return res.Wait()
+	return res.Wait(ctx)
 }
 
 func (c *RealtimeChannel) attach(result bool) (Result, error) {
@@ -207,7 +206,11 @@ func (c *RealtimeChannel) lockAttach(result bool, err error) (Result, error) {
 		ConnectionStateClosing,
 		ConnectionStateFailed:
 		return nil, newError(80000, errAttach)
+	}
 
+	c.lockSetState(ChannelStateAttaching, err)
+
+	switch c.client.Connection.State() {
 	// RTL4i
 	case ConnectionStateConnecting,
 		ConnectionStateDisconnected:
@@ -233,13 +236,11 @@ func (c *RealtimeChannel) lockAttach(result bool, err error) (Result, error) {
 			}
 
 			if result {
-				return res.Wait()
+				return res.Wait(context.Background())
 			}
 			return nil
 		}), nil
 	}
-
-	c.lockSetState(ChannelStateAttaching, err)
 
 	var res Result
 	if result {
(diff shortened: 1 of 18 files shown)


ably/realtime_channel.go
```diff
@@ -462,24 +428,65 @@ func (c *RealtimeChannel) History(params *PaginateParams) (*PaginatedResult, err
...
 
 func (c *RealtimeChannel) send(msg *proto.ProtocolMessage) (Result, error) {
-	if _, err := c.attach(false); err != nil {
-		return nil, err
-	}
-	res, listen := newErrResult()
-	switch c.State() {
-	case ChannelStateInitialized, ChannelStateAttaching:
-		c.queue.Enqueue(msg, listen)
+	if res, enqueued := c.maybeEnqueue(msg); enqueued {
```
lmars, line 431: Given `Publish` is now blocking but takes a context, should this queue not take into account the context deadline? For example, if a connection is stuck in a connecting / disconnecting loop and I publish with a 5s timeout, I'd probably expect that to return with an error after 5s?

tcard: The contexts (here and elsewhere) are handled at the outer layer. The pattern is:

* Public function takes a context and calls an internal function.
* Internal function starts some process and returns a Result to wait for it. (Here, the Result waits for the message to be dequeued, and either be failed or published and then acknowledged.)
* Public function waits for the Result, but will stop waiting if the context is called.

This implements [what we decided](https://github.com/ably/ably-go/issues/197#issuecomment-696979466) in terms of what a context means in the Realtime API. (In REST, contexts will be passed all the way to the HTTP library.)

ably/realtime_channel.go
```diff
@@ -462,24 +428,65 @@ func (c *RealtimeChannel) History(params *PaginateParams) (*PaginatedResult, err
...
-	default:
-		return nil, &ErrorInfo{Code: 90001}
 	}
+
+	if !c.canSend() {
+		return nil, newError(ErrChannelOperationFailedInvalidChannelState, nil)
+	}
+
+	res, listen := newErrResult()
 	if err := c.client.Connection.send(msg, listen); err != nil {
```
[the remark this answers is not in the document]
tcard: That's what [RTN19](https://docs.ably.io/client-lib-development-guide/features/#RTN19) is about (being implemented at #170).

ably/realtime_channel.go
lmars: Ah I'm sorry, I thought that `Publish` would not return when the context was cancelled, but I see I was incorrect 👍

ably/realtime_channel.go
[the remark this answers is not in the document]
lmars: Ah ok thanks

lmars (approved):

gernest (approved):

PR #210, opened 2020-11-11
Add RTN2g

gernest:
@SimonWoolf the library name is `ably-go` and not just `go` so the `lib` query param will contain something like  `ably-go-1.1.5` instead of `go-1.1.5`

### DIFF
--- a/ably/proto/http.go
+++ b/ably/proto/http.go
@@ -7,7 +7,7 @@ const (
 	AblyErrorCodeHeader    = "X-Ably-Errorcode"
 	AblyErrormessageHeader = "X-Ably-Errormessage"
 	LibraryVersion         = "1.1.5"
-	LibraryName            = "ably-go"
+	LibraryName            = "go"
 	LibraryString          = LibraryName + "-" + LibraryVersion
 	AblyVersion            = "1.0"
 	AblyClientIDHeader     = "X-Ably-ClientId"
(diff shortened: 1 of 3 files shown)


SimonWoolf:
> the library name is ably-go and not just go

I mean, `ably-js`, `ably-ruby`, `ably-python` etc. are all called that, but still identify themselves as `lib=js-web-xxx`, `lib=ruby-xxx`, `lib=python-xxx`, per the spec.

But whatever, that doesn't matter all that much, the most important thing is that it's identifying itself somehow

tcard:
> the library name is `ably-go` and not just `go`

This isn't right according to [RSC7b](https://docs.ably.io/client-lib-development-guide/features/#RSC7b), which says that for `ably-X`, `lib` is `X`. Can we make that change? This should also affect the `X-Ably-Lib` so it's just a matter of removing `ably-` from `proto.LibraryName`.

gernest:
@tcard @SimonWoolf I checked the spec you are right. I have made the change to use `go` instead of `ably-go` PTAL

tcard (approved):
@SimonWoolf Do you need this backported to the 1.1 branch (main) and released?

SimonWoolf:
> Do you need this backported to the 1.1 branch (main) and released?

Given we're releasing ably-go 1.2 soon, do we expect significant numbers people to be shunning that and upgrading only to the newest 1.1.x? If yes then yes, else no.

tcard:
Given 1.2 will come not just with an API upgrade but with major, critical features, I'd expect we'll be pushy about migrating users to 1.2.

PR #268, opened 2021-01-27
Generate env fallbacks

### LINKED ISSUE #198: Defaults: Generate environment fallbacks
Per `RSC15g`, introduced in https://github.com/ably/docs/pull/965.

See https://github.com/ably/ably-js/pull/682 and https://github.com/ably/ably-ruby/pull/196 for reference implementations.

sacOO7:
- Added go specific .gitignore file
- Refactored constants
- Added code to generate env. fallbacks
- Added relevant tests
- Fixes #198

### DIFF
--- a/ably/ably_test.go
+++ b/ably/ably_test.go
@@ -85,15 +85,17 @@ func assertFalse(t *testing.T, value bool) {
 	}
 }
 
-func assertNil(t *testing.T, value interface{}) {
-	if value != nil {
-		t.Errorf("%v is not not nil", value)
+func assertNil(t *testing.T, object interface{}) {
+	value := reflect.ValueOf(object)
+	if !value.IsNil() {
+		t.Errorf("%v is not nil", object)
 	}
 }
 
-func assertNotNil(t *testing.T, value interface{}) {
-	if value == nil {
-		t.Errorf("%v is nil", value)
+func assertNotNil(t *testing.T, object interface{}) {
+	value := reflect.ValueOf(object)
+	if value.IsNil() {
+		t.Errorf("%v is nil", object)
 	}
 }
 
--- a/ably/ably_test.go
+++ b/ably/ably_test.go
@@ -3,6 +3,7 @@ package ably_test
 import (
 	"fmt"
 	"io"
+	"reflect"
 	"testing"
 
 	"github.com/ably/ably-go/ably"
@@ -63,5 +64,49 @@ func checkError(code int, err error) error {
 		return fmt.Errorf("want e.Code=%d; got %d: %s", code, e.Code, err)
 	default:
 		return nil
+	}
+}
+
+func assertEquals(t *testing.T, expected interface{}, actual interface{}) {
+	if expected != actual {
+		t.Errorf("%v is not equal to %v", expected, actual)
+	}
+}
+
+func assertTrue(t *testing.T, value bool) {
+	if !value {
+		t.Errorf("%v is not true", value)
+	}
+}
+
+func assertFalse(t *testing.T, value bool) {
+	if value {
+		t.Errorf("%v is not false", value)
+	}
+}
+
+func assertNil(t *testing.T, value interface{}) {
+	if value != nil {
+		t.Errorf("%v is not not nil", value)
+	}
+}
+
+func assertNotNil(t *testing.T, value interface{}) {
+	if value == nil {
+		t.Errorf("%v is nil", value)
+	}
+}
+
+func assertDeepEquals(t *testing.T, expected interface{}, actual interface{}) {
+	areEqual := reflect.DeepEqual(expected, actual)
+	if !areEqual {
+		t.Errorf("%v is not equal to %v", expected, actual)
+	}
+}
+
+func assertEmpty(t *testing.T, value []string) {
+	assertNotNil(t, value)
+	if len(value) > 0 {
+		t.Errorf("%v is not empty", value)
 	}
 }
--- a/ably/ably_test.go
+++ b/ably/ably_test.go
@@ -92,23 +92,9 @@ func assertNil(t *testing.T, object interface{}) {
 	}
 }
 
-func assertNotNil(t *testing.T, object interface{}) {
-	value := reflect.ValueOf(object)
-	if value.IsNil() {
-		t.Errorf("%v is nil", object)
-	}
-}
-
 func assertDeepEquals(t *testing.T, expected interface{}, actual interface{}) {
 	areEqual := reflect.DeepEqual(expected, actual)
 	if !areEqual {
 		t.Errorf("%v is not equal to %v", expected, actual)
-	}
-}
-
-func assertEmpty(t *testing.T, value []string) {
-	assertNotNil(t, value)
-	if len(value) > 0 {
-		t.Errorf("%v is not empty", value)
 	}
 }
--- a/ably/options.go
+++ b/ably/options.go
@@ -276,14 +276,14 @@ func (opts *ClientOptions) isProductionEnvironment() bool {
 	return env == "" || strings.EqualFold(env, "production")
 }
 
-func (opts *ClientOptions) getPort() (port int, isDefaultPort bool) {
+func (opts *ClientOptions) activePort() (port int, isDefault bool) {
 	if opts.NoTLS {
 		port = opts.Port
 		if port == 0 {
 			port = defaultOptions.Port
 		}
 		if port == defaultOptions.Port {
-			isDefaultPort = true
+			isDefault = true
 		}
 		return
 	}
@@ -292,7 +292,7 @@ func (opts *ClientOptions) getPort() (port int, isDefaultPort bool) {
 		port = defaultOptions.TLSPort
 	}
 	if port == defaultOptions.TLSPort {
-		isDefaultPort = true
+		isDefault = true
 	}
 	return
 }
@@ -376,7 +376,7 @@ func empty(s string) bool {
 
 func (opts *ClientOptions) restURL() ( restUrl string) {
 	restHost := opts.getRestHost()
-	port, _ := opts.getPort()
+	port, _ := opts.activePort()
 	baseUrl := net.JoinHostPort(restHost, strconv.FormatInt(int64(port), 10))
 	if opts.NoTLS {
 		return "http://" + baseUrl
@@ -386,7 +386,7 @@ func (opts *ClientOptions) restURL() ( restUrl string) {
 
 func (opts *ClientOptions) realtimeURL() ( realtimeUrl string) {
 	realtimeHost := opts.getRealtimeHost()
-	port, _ := opts.getPort()
+	port, _ := opts.activePort()
 	baseUrl := net.JoinHostPort(realtimeHost, strconv.FormatInt(int64(port), 10))
 	if opts.NoTLS {
 		return "ws://" + baseUrl
@@ -396,7 +396,7 @@ func (opts *ClientOptions) realtimeURL() ( realtimeUrl string) {
 
 func (opts *ClientOptions) getFallbackHosts() ( [] string, error) {
 	logger := opts.Logger.Sugar()
-	_, isDefaultPort := opts.getPort()
+	_, isDefaultPort := opts.activePort()
 	if (opts.FallbackHostsUseDefault) {
 		if opts.FallbackHosts != nil {
 			return nil , fmt.Errorf("fallbackHosts and fallbackHostsUseDefault cannot both be set")
--- a/ably/options.go
+++ b/ably/options.go
@@ -408,10 +408,7 @@ func (opts *ClientOptions) getFallbackHosts() ([]string, error) {
 		}
 		return getEnvFallbackHosts(opts.Environment), nil
 	}
-	if opts.FallbackHosts != nil {
-		return opts.FallbackHosts, nil
-	}
-	return []string{}, nil
+	return opts.FallbackHosts, nil
 }
 
 func (opts *ClientOptions) httpclient() *http.Client {
--- a/ably/options.go
+++ b/ably/options.go
@@ -255,6 +255,28 @@ func NewClientOptions(key string) *ClientOptions {
 	}
 }
 
+func (opts *ClientOptions) isProductionEnvironment() bool {
+	env := opts.Environment
+	return env == "" || strings.EqualFold(env, "production")
+}
+
+func (opts *ClientOptions) getPort() (port int, isDefaultPort bool) {
+	if opts.NoTLS {
+		port = opts.Port
+		if port == 0 || port == defaultOptions.Port {
+			port = defaultOptions.Port
+			isDefaultPort = true
+		}
+	}
+	port = opts.TLSPort
+	if port == 0 || port == defaultOptions.TLSPort {
+		port = defaultOptions.TLSPort
+		isDefaultPort = true
+	}
+	return
+}
+
+
 func (opts *ClientOptions) timeoutConnect() time.Duration {
 	if opts.TimeoutConnect != 0 {
 		return opts.TimeoutConnect
--- a/ably/options.go
+++ b/ably/options.go
@@ -267,6 +267,7 @@ func (opts *ClientOptions) getPort() (port int, isDefaultPort bool) {
 			port = defaultOptions.Port
 			isDefaultPort = true
 		}
+		return
 	}
 	port = opts.TLSPort
 	if port == 0 || port == defaultOptions.TLSPort {
@@ -323,24 +324,14 @@ func (opts *ClientOptions) restURL() string {
 	host := opts.RestHost
 	if host == "" {
 		host = defaultOptions.RestHost
-		if opts.Environment == "production" {
-			opts.Environment = ""
-		}
-		if opts.Environment != "" {
+		if !opts.isProductionEnvironment() {
 			host = opts.Environment + "-" + host
 		}
 	}
+	port, _ := opts.getPort()
 	if opts.NoTLS {
-		port := opts.Port
-		if port == 0 {
-			port = 80
-		}
 		return "http://" + net.JoinHostPort(host, strconv.FormatInt(int64(port), 10))
 	} else {
-		port := opts.TLSPort
-		if port == 0 {
-			port = 443
-		}
 		return "https://" + net.JoinHostPort(host, strconv.FormatInt(int64(port), 10))
 	}
 }
@@ -349,24 +340,14 @@ func (opts *ClientOptions) realtimeURL() string {
 	host := opts.RealtimeHost
 	if host == "" {
 		host = defaultOptions.RealtimeHost
-		if opts.Environment == "production" {
-			opts.Environment = ""
-		}
-		if opts.Environment != "" {
+		if !opts.isProductionEnvironment() {
 			host = opts.Environment + "-" + host
 		}
 	}
+	port, _ := opts.getPort()
 	if opts.NoTLS {
-		port := opts.Port
-		if port == 0 {
-			port = 80
-		}
 		return "ws://" + net.JoinHostPort(host, strconv.FormatInt(int64(port), 10))
 	} else {
-		port := opts.TLSPort
-		if port == 0 {
-			port = 443
-		}
 		return "wss://" + net.JoinHostPort(host, strconv.FormatInt(int64(port), 10))
 	}
 }
--- a/ably/options.go
+++ b/ably/options.go
@@ -415,6 +415,9 @@ func (opts *ClientOptions) getFallbackHosts() ([]string, error) {
 		}
 		return getEnvFallbackHosts(opts.Environment), nil
 	}
+	if opts.FallbackHosts != nil {
+		return opts.FallbackHosts, nil
+	}
 	return []string{}, nil
 }
 
--- a/ably/options.go
+++ b/ably/options.go
@@ -186,10 +186,13 @@ func (opts *AuthOptions) KeySecret() string {
 	return ""
 }
 
+
 type ClientOptions struct {
 	AuthOptions
 
 	RestHost                string // optional; overwrite endpoint hostname for REST client
+
+	// Deprecated: FallbackHostsUseDefault will be removed in the future
 	FallbackHostsUseDefault bool
 
 	FallbackHosts   []string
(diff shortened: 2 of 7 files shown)


ably/options.go
```diff
@@ -186,10 +186,13 @@ func (opts *AuthOptions) KeySecret() string {
...
 }
 
+
 type ClientOptions struct {
 	AuthOptions
 
 	RestHost                string // optional; overwrite endpoint hostname for REST client
+
+	// Deprecated: FallbackHostsUseDefault will be removed in the future
 	FallbackHostsUseDefault bool
```
tcard, line 196: Doc comments in Go [should start with the name being documented](https://golang.org/doc/effective_go.html#commentary):

> 
>  Inside a package, any comment immediately preceding a top-level declaration serves as a doc comment for that declaration. Every exported (capitalized) name in a program should have a doc comment.
> 
> Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.
> 
> ```go
> // Compile parses a regular expression and returns, if successful,
> // a Regexp that can be used to match against text.
> func Compile(str string) (*Regexp, error) {
> ```

Also, I'd suggest pointing users to the non-deprecated alternative.

ably/options.go
```diff
@@ -337,45 +340,84 @@ func (opts *ClientOptions) disconnectedRetryTimeout() time.Duration {
...
+	return host
+}
+
+func (opts *ClientOptions) getRealtimeHost() (string) {
+	host := opts.RealtimeHost
+	defaultHost := defaultOptions.RealtimeHost
+	if empty(host) {
+		if !empty(opts.RestHost) {
+			logger := opts.Logger.Sugar()
+			logger.Warnf(`restHost is set to %s but realtimeHost is not set so setting realtimeHost to %s too.
```
tcard, line 361: I'm not sure we should be logging this, but even if we do, shouldn't we log this if it's the other way around as well? Namely, that `realtimeHost` is set but `restHost` isn't.

ably/options.go
```diff
@@ -337,45 +340,84 @@ func (opts *ClientOptions) disconnectedRetryTimeout() time.Duration {
...
+			return nil , fmt.Errorf("fallbackHosts and fallbackHostsUseDefault cannot both be set")
+		}
+		if !isDefaultPort {
+			return nil, fmt.Errorf("fallbackHostsUseDefault cannot be set when port or tlsPort are set")
+		}
+		if !empty(opts.Environment) {
+			logger.Warn(`Deprecated fallbackHostsUseDefault : There is no longer a need to set this when the environment option is also set 
+								since the library can generate the correct fallback hosts using the environment option.`)
+		}
+		logger.Warn("Deprecated fallbackHostsUseDefault : fallbackHosts: Ably.Defaults.FALLBACK_HOSTS")
```
tcard, line 411: Not sure what this message means.

ably/rest_client.go
```diff
@@ -357,9 +357,10 @@ func (c *RestClient) doWithHandle(r *Request, handle func(*http.Response, interf
 				(c.opts.FallbackHostsUseDefault ||
 					strings.HasPrefix(req.URL.Host, defaultOptions.RestHost) ||
 					c.opts.FallbackHosts != nil) {
-				fallback := defaultOptions.FallbackHosts
-				if c.opts.FallbackHosts != nil {
-					fallback = c.opts.FallbackHosts
+				fallback, error := c.opts.getFallbackHosts()
```
tcard, line 360: If the options are invalid, this should be detected earlier, ie. on `NewREST` or `NewRealtime`.

tcard (commented):
Left some, mostly minor, comments.

I haven't really reviewed the functionality in depth yet; will do when tests are spec-oriented so it's easier to match tests and spec.

ably/options.go
sacOO7: I believe that's how it's implemented in other libraries as per spec.

ably/rest_client.go
sacOO7: It's invalid only in the case we are using fallbackHosts ... there might be the case when we might not need fallbackHosts. But, I will check if it can be added at the top level.

ably/options.go
sacOO7: Actually, I referred to the https://github.com/ably/ably-go/blob/681c5a8240349e7b4fe6ea4e7318125383786326/ably/options.go#L233-L234 in clientOptions for deprecation comment

sacOO7: Also, there is no alternative specified to the FallbackHostsUseDefault as per spec. FallbackHostsUseDefault option will be removed in the future as the comment says.

ably/rest_client.go
sacOO7: As per [RSC15k](https://docs.ably.io/client-lib-development-guide/features/#RSC15k), there will be an instant error when retried on fallback hosts that do not qualify according to the configuration

ably/options.go
tcard: Ah true, this is the internal `clientOptions` struct, not the public-facing option.

So, how about putting the deprecation notice in the public `WithFallbackHostsUseDefault` function too, where users and tooling may see it?

> Also deprecation comments won't work if they start with property name, you can check
https://rakyll.org/deprecated/

That very article shows how you can do both:

```go
// Execer is an optional interface that may be implemented by a Conn.
//
// [...]
//
// Deprecated: Drivers should implement ExecerContext instead (or additionally).
type Execer interface {
```

> Also, there is no alternative specified to the FallbackHostsUseDefault as per spec

How about something like:

> Deprecated: The library will automatically use default fallback hosts when a custom REST host or custom fallback hosts aren't provided.

ably/rest_client.go
tcard: > (RSC15b) The fallback behavior described by this section, RSC15, only applies when either:
> (RSC15b1) ClientOptions#restHost has not been set to an explicit value (see RSC11a) and ClientOptions#port is not set and ClientOptions#tlsPort is not set
> (RSC15b2) An array of ClientOptions#fallbackHosts is provided
> (RSC15b3) the deprecated ClientOptions#fallbackHostsUseDefault option is set to true
> (RSC15k) When none of the statements in RSC15b are true then host fallback does not apply, and failing HTTP requests that would have qualified for a retry against a fallback host will instead result in an error immediately

This is a whole different thing than [TO3k7](https://docs.ably.io/client-lib-development-guide/features/#TO3k7), which says (emphasis mine):

> (TO3k7) fallbackHostsUseDefault boolean (deprecated) – optionally allows the default fallback hosts [a-e].ably-realtime.com to be used when environment is not production or a custom realtime or REST host endpoint is being used. **It is never valid to configure fallbackHost and set fallbackHostsUseDefault to true**. The fallbackHostsUseDefault option is deprecated and future library releases will ignore any supplied value

RSC15k is about the effects of _valid_ fallback hosts options, while TO3k7 is about what makes  invalid.

Now, it is left unspecificed what should happen when invalid options are provided, but given (per [RSC1b](https://docs.ably.io/client-lib-development-guide/features/#RSC1b)) that we already validate _some_ options at instantiation, I think it's worth validating _all_ options at that point.

It's definitely better dev experience to get early, single feedback on misconfiguration, which is a coding mistake that can just be fixed, than to carry on only to have fallback hosts feature don't work when you actually need it.

ably/options.go
tcard: Sorry, I missed that this actually overrides the default realtime host.

ably/options_test.go
```diff
@@ -2,11 +2,195 @@ package ably_test
...
+	t.Run("with env should return environment fallback hosts", func(ts *testing.T) {
+		expectedFallBackHosts := []string{
+			"sandbox-a-fallback.ably-realtime.com",
+			"sandbox-b-fallback.ably-realtime.com",
+			"sandbox-c-fallback.ably-realtime.com",
+			"sandbox-d-fallback.ably-realtime.com",
+			"sandbox-e-fallback.ably-realtime.com",
+		}
+		hosts := ably.GetEnvFallbackHosts("sandbox")
+		assert.True(ts, reflect.DeepEqual(expectedFallBackHosts, hosts), " %s should match %s", hosts, expectedFallBackHosts)
```
tcard, line 38: Suggestion: If we're using testify after all, we might as well make full use of it:

```suggestion
		assert.Equal(ts, expectedFallBackHosts, hosts)
```

This will emit an appropriate error message, with diff and all.

ably/options_test.go
```diff
@@ -2,11 +2,195 @@ package ably_test
...
+			"e.ably-realtime.com",
+		}
+		hosts := ably.DefaultFallbackHosts()
+		assert.True(ts, reflect.DeepEqual(expectedFallBackHosts, hosts), " %s should match %s", hosts, expectedFallBackHosts)
+	})
+}
+
+func TestEnvFallbackHosts_RSC15i(t *testing.T) {
+	t.Parallel()
+	t.Run("with env should return environment fallback hosts", func(ts *testing.T) {
```
tcard, line 29: Naming the inner `*testing.T` also `t` is common practice and ensures that we don't use the external `t` accidentally.

ably/options_test.go
tcard, line 38: Suggestion: If we're using testify after all, we might as well make full use of it:

```suggestion
		assert.Equal(ts, expectedFallBackHosts, hosts)
```

This will emit an appropriate error message, with diff and all.

ably/options_test.go
```diff
@@ -2,11 +2,195 @@ package ably_test
...
+			"sandbox-d-fallback.ably-realtime.com",
+			"sandbox-e-fallback.ably-realtime.com",
+		}
+		hosts := ably.GetEnvFallbackHosts("sandbox")
+		assert.True(ts, reflect.DeepEqual(expectedFallBackHosts, hosts), " %s should match %s", hosts, expectedFallBackHosts)
+	})
+}
+
+func TestFallbackHosts_RSC15b(t *testing.T) {
+	t.Parallel()
```
tcard, line 43: This applies to the `*testing.T` it's called on, so the inner `t.Run`s won't run in parallel unless you do `ts.Parallel()` on each of them.

tcard (approved):
LGTM. However:

* I believe we should fail at NewRealtime/NewREST if invalid ClientOptions are given.
* We're discussing whether we should depend on testify.

I'll open issues about those separately if they aren't resolved before this PR is merged.

ably/rest_client.go
sacOO7: @lmars what do you think about this one?

ably/options.go
```diff
@@ -46,6 +51,19 @@ func DefaultFallbackHosts() []string {
 	}
 }
 
+func getEnvFallbackHosts(env string) []string {
+	formatHost := func(env string, host string) string {
+		return fmt.Sprintf("%s-%s", env, host)
+	}
```
lmars, line 57: The function feels unnecessary, I'd probably just concat the strings directly:

```go
return []string{
	env + "-a-fallback.ably-realtime.com",
	env + "-b-fallback.ably-realtime.com",
	env + "-c-fallback.ably-realtime.com",
	env + "-d-fallback.ably-realtime.com",
	env + "-e-fallback.ably-realtime.com",
}
```

ably/options.go
```diff
@@ -292,56 +338,87 @@ func (opts *ClientOptions) disconnectedRetryTimeout() time.Duration {
...
-	} else {
-		port := opts.TLSPort
-		if port == 0 {
-			port = 443
-		}
-		return "https://" + net.JoinHostPort(host, strconv.FormatInt(int64(port), 10))
+	if host == defaultHost && !opts.isProductionEnvironment() {
+		host = opts.Environment + "-" + host
 	}
+	return host
```
lmars, line 350: I feel like this would be clearer by dealing with the more specific cases first (rest host specified, environment specified) and then falling through to the more general case (default host), which also makes it easier to verify against [RSC11](https://www.ably.io/documentation/client-lib-development-guide/features/#RSC11):

```go
if opts.RestHost != "" {
	return opts.RestHost
}
if opts.Environment != "" && !opts.isProductionEnvironment() {
	return opts.Environment + "-" + defaultOptions.RestHost
}
return defaultOptions.RestHost
```

ably/options.go
```diff
@@ -292,56 +338,87 @@ func (opts *ClientOptions) disconnectedRetryTimeout() time.Duration {
...
-			opts.Environment = ""
-		}
-		if opts.Environment != "" {
-			host = opts.Environment + "-" + host
+	defaultHost := defaultOptions.RealtimeHost
+	if empty(host) {
+		if !empty(opts.RestHost) {
+			logger := opts.Logger.Sugar()
+			logger.Warnf(`restHost is set to %s but realtimeHost is not set so setting realtimeHost to %s too. 
+							If this is not what you want, please set realtimeHost explicitly.`, opts.RestHost, opts.RealtimeHost)
```
lmars, line 360: Since this string is enclosed in backticks, whitespace will be preserved and hence will appear in the resulting log message, so I think we should put this all on the line above.

ably/options_test.go
sacOO7: Got it

sacOO7: It seems we need to be careful while using parallel in certain cases, like this one 
https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

ably/options.go
```diff
@@ -292,56 +338,87 @@ func (opts *ClientOptions) disconnectedRetryTimeout() time.Duration {
...
+}
+
+func empty(s string) bool {
+	return len(strings.TrimSpace(s)) == 0
+}
+
+func (opts *ClientOptions) restURL() (restUrl string) {
+	restHost := opts.getRestHost()
+	port, _ := opts.activePort()
+	baseUrl := net.JoinHostPort(restHost, strconv.FormatInt(int64(port), 10))
```
lmars, line 378: Could use [`Itoa`](https://golang.org/pkg/strconv/#Itoa) rather than `FormatInt`

ably/options.go
```diff
@@ -292,56 +338,87 @@ func (opts *ClientOptions) disconnectedRetryTimeout() time.Duration {
...
+	}
+	return "wss://" + baseUrl
+}
+
+func (opts *ClientOptions) getFallbackHosts() ([]string, error) {
+	logger := opts.Logger.Sugar()
+	_, isDefaultPort := opts.activePort()
+	if opts.FallbackHostsUseDefault {
+		if opts.FallbackHosts != nil {
+			return nil, fmt.Errorf("fallbackHosts and fallbackHostsUseDefault cannot both be set")
```
lmars, line 400: Use [`errors.New`](https://golang.org/pkg/errors/#New) rather than `fmt.Errorf` (since nothing is being formatted)

ably/options.go
```diff
@@ -292,56 +338,87 @@ func (opts *ClientOptions) disconnectedRetryTimeout() time.Duration {
...
 		}
-		return "ws://" + net.JoinHostPort(host, strconv.FormatInt(int64(port), 10))
-	} else {
-		port := opts.TLSPort
-		if port == 0 {
-			port = 443
+		if !isDefaultPort {
+			return nil, fmt.Errorf("fallbackHostsUseDefault cannot be set when port or tlsPort are set")
```
lmars, line 403: Use [`errors.New`](https://golang.org/pkg/errors/#New) rather than `fmt.Errorf` (since nothing is being formatted)

ably/options.go
```diff
@@ -292,56 +338,87 @@ func (opts *ClientOptions) disconnectedRetryTimeout() time.Duration {
...
+	if opts.FallbackHosts == nil && empty(opts.RestHost) && empty(opts.RealtimeHost) && isDefaultPort {
+		if opts.isProductionEnvironment() {
+			return defaultOptions.FallbackHosts, nil
+		}
+		return getEnvFallbackHosts(opts.Environment), nil
+	}
+	if opts.FallbackHosts != nil {
+		return opts.FallbackHosts, nil
 	}
+	return []string{}, nil
```
lmars, line 421: These three lines could just be `return opts.FallbackHosts, nil`

ably/rest_client.go
lmars: I think the check should be in the constructor yes, I believe that's what we do in other client libs too.

ably/options.go
sacOO7: Won't it become too long to read? what about concatenation by + sign

ably/options.go
sacOO7: Good catch. I was looking for a replacement for this one too.

ably/options.go
lmars: My screen is quite wide, and my text editor (Vim) is advanced enough to wrap lines if I tell it 🙂

sacOO7: Vim 🙏

ably/realtime_client.go
```diff
@@ -23,6 +23,12 @@ func NewRealtimeClient(opts *ClientOptions) (*RealtimeClient, error) {
 	if opts == nil {
 		panic("called NewRealtimeClient with nil ClientOptions")
 	}
+	_, error := opts.getFallbackHosts()
```
lmars, line 26: I think we should consider this a general validation call on the options and hide the fact that validating includes checking fallbacks to be within that call, e.g.:

```go
if err := opts.validate(); err != nil {
	return nil, err
}
```

Also, call the error variable `err` to avoid shadowing the `error` type.

ably/rest_client.go
```diff
@@ -353,17 +359,11 @@ func (c *RestClient) doWithHandle(r *Request, handle func(*http.Response, interf
...
-				if c.opts.FallbackHosts != nil {
-					fallback = c.opts.FallbackHosts
-				}
-				log.Info("RestClient: trying to fallback with hosts=%v", fallback)
-				if len(fallback) > 0 {
-					left := fallback
+			if canFallBack(e.StatusCode) {
+				fallbacks, _ := c.opts.getFallbackHosts()
+				log.Info("RestClient: trying to fallback with hosts=%v", fallbacks)
+				if fallbacks != nil && len(fallbacks) > 0 {
```
lmars, line 365: This `nil` check on `fallbacks` is redundant, since the length of a nil slice is zero so is covered by `len(fallbacks) > 0`

sacOO7: Okay, the whitespace issue is resolved by formatter.  I didn't intend to introduce it.

ably/ably_test.go
```diff
@@ -65,3 +66,49 @@ func checkError(code int, err error) error {
...
+	value := reflect.ValueOf(object)
+	if !value.IsNil() {
+		t.Errorf("%v is not nil", object)
+	}
+}
+
+func assertNotNil(t *testing.T, object interface{}) {
+	value := reflect.ValueOf(object)
+	if value.IsNil() {
+		t.Errorf("%v is nil", object)
```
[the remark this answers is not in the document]
sacOO7: I didn't intend to generalize it. Feel free to update it based on the requirement. In my case, it shows 
[ ] is nil. The same goes for the above example.

ably/ably_test.go
```diff
@@ -65,3 +66,49 @@ func checkError(code int, err error) error {
...
+
+func assertDeepEquals(t *testing.T, expected interface{}, actual interface{}) {
+	areEqual := reflect.DeepEqual(expected, actual)
+	if !areEqual {
+		t.Errorf("%v is not equal to %v", expected, actual)
+	}
+}
+
+func assertEmpty(t *testing.T, value []string) {
+	assertNotNil(t, value)
```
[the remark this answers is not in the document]
sacOO7: I wasn't aware of nil-slices. I will update the same.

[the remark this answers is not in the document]
sacOO7: I think we can still extract it out. I didn't really consider the edge case scenarios. I just wrote abstractions that would work for me.

ably/realtime_client.go
tcard: If we do that, maybe it's worth moving the validation that happens on `newAuth` there as well.

ably/ably_test.go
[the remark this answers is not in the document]
tcard: > Also, an error message of `nil is nil` isn't very useful.

This is alleviated by calling `t.Helper()` from helper functions. Then the error feedback points to the line this function was called.

Not that I disagree that this function is not worth the trouble anyway. I think `assertEqual` is enough. Error checks can be just `if err !=nil { t.Fatal(err) }`, len checks can be just `assertEqual(y, len(x))`, etc.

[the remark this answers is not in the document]
lmars: > In my case, it shows [ ] is nil.

I mean, that's even worse, no? That doesn't seem to convey anything useful to the caller, I'd much prefer to see the name of the variable rather than what it's value actually is (e.g. `foo is nil`, not `nil is nil` or `[] is nil`)

> Feel free to update it based on the requirement

I'll leave it with you to decide whether this needs to change, but whatever you go for I think should be documented in a comment on the function (i.e. if it stays as is, a comment that the error is non fatal and a further nil check should be made on the value if it's subsequently used).

ably/ably_test.go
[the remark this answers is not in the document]
sacOO7: It's not being used, I will remove it for the time being

ably/ably_test.go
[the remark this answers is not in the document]
sacOO7: It's not being used, I will remove it for the time being

ably/rest_client.go
sacOO7: Fixed the issue

ably/realtime_client.go
sacOO7: @tcard , I was looking into validations that can be moved to validate method from newAuth, but it seems newAuth contains its own conditional clauses embedded inside validation to determine host, token details, method, and clientID. Even if we move validations here, there will be a replication of logic in newAuth. I would suggest keeping validation simple for now.

ably/options.go
```diff
@@ -250,6 +268,42 @@ func NewClientOptions(key string) *ClientOptions {
 	}
 }
 
+func (opts *ClientOptions) validate() error {
+	_, error := opts.getFallbackHosts()
```
lmars, line 272: Please use `err` rather than `error` as the variable name

ably/options.go
lmars: @sacOO7 this has still not been resolved, if you intend to leave it as is please say, but I really think the output will look very odd in a log

lmars (approved):

lmars:
@sacOO7 nice, thanks 🙂

PR #273, opened 2021-02-01
Replace Travis with GitHub workflow

QuintinWillison:
We're working towards conforming client libraries to use GitHub Actions.


QuintinWillison:
Bit confused that the JSON tests take around about 4 minutes but the MessagePack tests take 1 second!

Apart from that, I think this looks like a good starting point.

.github/workflows/check.yml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  check:
+
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    strategy:
+      matrix:
+        go-version: [1.13, 1.14, 1.15]
+
```
sacOO7, line 18: Not sure if we want this to be cross platform, but we can simply add here 
```
os: [macos-latest, ubuntu-latest, windows-latest]
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,46 @@
...
+on:
+  pull_request:
+  push:
+    branches: [ $default-branch ]
+
+jobs:
+  check:
+
+    runs-on: ubuntu-latest
```
sacOO7, line 12: And ```runs-on: ${{ matrix.os }}```

.github/workflows/check.yml
QuintinWillison: It's a valid question and equally applies to other repositories as we adopt workflows everywhere.

For the moment, however, I think focussing on the "most likely" host OS is acceptable. I'm going to keep this thought in the back of my mind, however, as I can imagine certain edge cases where perhaps Windows build and check might smoke out some issues around paths and the like.

.github/workflows/check.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Set up Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Download Packages
+        run: go get -t -v ./ably/...
+
+      - name: Vet
+        run: go vet ./ably ./scripts
```
sacOO7, line 33: Maybe we can reuse commands from makefile. Possibly need to update some of the makefile commands to keep both in sync
```make vet``` 
should work in this case

Comment (QuintinWillison) [reply .github/workflows/check.yml:33]: I've deleted the `Makefile`. It was unnecessary.

**Comment by QuintinWillison** [comment]
We were seeing [this](https://github.com/ably/ably-php/actions/runs/528889413) in the PHP repository workflow, hence https://github.com/ably/ably-go/pull/273/commits/27142383b74d6dd398c2474b206b77951cf10947. e.g. preventing failure of all workflows with something like:

> Process completed with exit code 1.
> The job was canceled because "_7_1" failed.

**Review by tcard — APPROVED** [review]

File: .github/workflows/check.yml (hunk 3 above)
Comment (sacOO7) [reply .github/workflows/check.yml:34]: If it's not needed anymore, then good

**Review by sacOO7 — APPROVED** [review]

File: .github/workflows/check.yml (hunk 3 above)
Comment (lmars) [reply .github/workflows/check.yml:34]: I'm not sure I agree with replacing `make test` with `for both JSON and MessagePack protocols (see [workflow](.github/workflows/check.yml) for the test commands executed in the CI environment)` as the way to run the tests, why make developers jump through that hoop, especially casual contributors?

File: .github/workflows/check.yml [hunk 4]
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Build
+        run: go build -v ./...
```
lmars, line 37: What exactly is being built here? It feels a bit redundant given ably-go is a library not an executable

tcard:
> Bit confused that the JSON tests take around about 4 minutes but the MessagePack tests take 1 second!

Go caches test results, and invalidation happens when the code being tests changes. It seems that passing different env vars doesn't invalidate the cache.

I suggest adding `-count 1` to the `go test` commands, which forces a real re-run.

tcard:
Alternatively, and probably better yet, we could run both in parallel. If that's too many parallel tasks, we can just test with the lowest Go version (1.13) since Go compatibility guarantees follows semver and ensures code continues working on later Go versions.

.github/workflows/check.yml
QuintinWillison: It was in the original Travis file, for starters. 😁 

However, if it's adding no benefit then I am happy to remove it. My assumption was that it a valid thing to run on a check workflow, if nothing else as a smoke test to find out asking Go to `build` blows up in smoke.

tcard: `go test` effectively includes `go build`, so yes, we can remove it.

`go build` on its own is useful when you want to check the library compiles, but don't want to run the tests. It shouldn't have been in Travis in the first place.

tcard:
@QuintinWillison Seeing those failures, how about basing this on `integration/1.2`? `main` is pretty stale and `integration/1.2` will become `main` soon enough.

QuintinWillison:
Thanks for the suggestion, @tcard - however, I've now removed the Travis webhook and have enabled Actions for this repository so my preference is to land this on `main` now, despite the failures. We need to get this moved forward and there's no benefit in hiding from the shortcomings of the current default branch state.

We should merge from `main` to the integration branch once this has landed so that you can benefit from the workflows in that arena.

PR #276, opened 2021-02-02
Propagate contexts everywhere

### LINKED ISSUE #275: Add missing context.Context arguments
Those methods are OK, except they're missing a context.Context.

tcard:
Now HTTP requests require contexts. Since lots of things end up making
HTTP requests, contexts need to be propagated everywhere. Huge change,
but not one that can be easily or usefully split.

For where we don't have a context passed as a parameter because it's the
library itself who initiates the operation, a base context can be set
with a new ClientOption.

Fixes #275.

### DIFF
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -338,11 +338,7 @@ func (a *Auth) setDefaults(opts *authOptions, req *TokenRequest) error {
 		req.ClientID = a.opts().ClientID
 	}
 	if req.Timestamp == 0 {
-		var ctx context.Context
-		if opts.UseQueryTime {
-			ctx = context.Background()
-		}
-		ts, err := a.timestamp(ctx)
+		ts, err := a.timestamp(context.Background(), opts.UseQueryTime)
 		if err != nil {
 			return err
 		}
@@ -352,9 +348,9 @@ func (a *Auth) setDefaults(opts *authOptions, req *TokenRequest) error {
 }
 
 //Timestamp returns the timestamp to be used in authorization request.
-func (a *Auth) timestamp(queryCtx context.Context) (time.Time, error) {
+func (a *Auth) timestamp(ctx context.Context, query bool) (time.Time, error) {
 	now := a.client.opts.Now()
-	if queryCtx == nil {
+	if !query {
 		return now, nil
 	}
 	if a.serverTimeOffset != 0 {
@@ -372,7 +368,7 @@ func (a *Auth) timestamp(queryCtx context.Context) (time.Time, error) {
 		}
 		serverTime = t
 	} else {
-		t, err := a.client.Time(queryCtx)
+		t, err := a.client.Time(ctx)
 		if err != nil {
 			return time.Time{}, newError(ErrUnauthorized, err)
 		}
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -168,18 +168,18 @@ func (a *Auth) createTokenRequest(params *TokenParams, opts *authOptions) (*Toke
 }
 
 // RequestToken
-func (a *Auth) RequestToken(params *TokenParams, opts ...AuthOption) (*TokenDetails, error) {
+func (a *Auth) RequestToken(ctx context.Context, params *TokenParams, opts ...AuthOption) (*TokenDetails, error) {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
 	var o *authOptions
 	if opts != nil {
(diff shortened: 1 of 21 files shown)


ably/auth.go
```diff
@@ -360,9 +352,9 @@ func (a *Auth) setDefaults(opts *authOptions, req *TokenRequest) error {
 }
 
 //Timestamp returns the timestamp to be used in authorization request.
-func (a *Auth) timestamp(query bool) (time.Time, error) {
+func (a *Auth) timestamp(queryCtx context.Context) (time.Time, error) {
 	now := a.client.opts.Now()
-	if !query {
+	if queryCtx == nil {
```
lmars, line 357: I'm not sure I agree with replacing a boolean with a nil check on the context (since it's bad practice to pass a `nil` context), why not continue to use the boolean argument?

ably/options.go
```diff
@@ -699,6 +702,15 @@ func WithDial(dial func(protocol string, u *url.URL, timeout time.Duration) (pro
 	}
 }
 
+// WithBaseContext establishes a context to use for operations initiated by the
+// the library, such as establishing realtime connections and requesting
+// implicit authorizations.
```
lmars, line 707: It's bad practice to store a context as a struct field, is passing a context directly to function calls not sufficient?

ably/realtime_conn.go
```diff
@@ -116,6 +116,7 @@ func (c *Connection) dial(proto string, u *url.URL) (conn proto.Conn, err error)
 	query.Add("heartbeats", "true")
 	u.RawQuery = query.Encode()
 	timeout := c.opts.realtimeRequestTimeout()
+	// TODO: Use c.opts.baseCtx to dial.
```
lmars, line 119: I disagree with this, if we need a context here, it should be passed through the call stack, or derived from the options (e.g. `context.WithTimeout(context.Background(), c.opts.DialTimeout)`)

ably/rest_client.go
```diff
@@ -157,17 +157,18 @@ func (c *REST) Time() (time.Time, error) {
...
 }
 
 // request this contains fields necessary to compose http request that will be
 // sent ably endpoints.
 type request struct {
-	Method string
-	Path   string
-	In     interface{} // value to be encoded and sent with request body
-	Out    interface{} // value to store decoded response body
+	Context context.Context
```
lmars, line 167: This context should be passed via the call stack, not stored as a struct field

ably/auth.go
tcard: I was on the fence for this one because an extra boolean flag that can just be a nil check on the context is redundant. But you're right it feels weird. Done at cf66a78.

ably/options.go
tcard: The library makes requests on its own, from the goroutine that manages the connection, for reauth, reconnection, etc. It can't take the context from the caller's stack because the caller is the library itself.

Now that I think of it, I don't think having a custom base context is useful at all; users can provide `Dial` and `http.Client` options that ultimately are passed and can wrap the context, so we might as well just pass `context.Background()` to them. WDYT?

ably/rest_client.go
tcard: Not sure why I did that, it wasn't even used. Fixed at c8a85d68fba44cd6f45c225702154cc15ecd0f51.

ably/realtime_conn.go
tcard: Comment removed at 3b04db76d797942ccfb586ab52e1b3fdcc9b37bb.

ably/options.go
tcard: Removed the option at 3b04db76d797942ccfb586ab52e1b3fdcc9b37bb.

ably/auth.go
lmars: 👍

ably/options.go
lmars: Yep makes sense 👍

lmars (approved):

tcard:
All CI errors are like this:

```
--- FAIL: TestFixConnLeak_ISSUE89 (6.87s)
    rest_client_test.go:574: expected conn 10.1.0.4:47296 to be closed
    rest_client_test.go:574: expected conn 10.1.0.4:47308 to be closed
    rest_client_test.go:574: expected conn 10.1.0.4:47312 to be closed
```

This is not caused by this PR, and is being fixed by #277. So, merging.

PR #277, opened 2021-02-04
TestFixConnLeak_ISSUE89: Change the way it detects closed conns

tcard:
In the new CI, we often see this:

```
--- FAIL: TestFixConnLeak_ISSUE89 (6.87s)
    rest_client_test.go:574: expected conn 10.1.0.4:47296 to be closed
    rest_client_test.go:574: expected conn 10.1.0.4:47308 to be closed
    rest_client_test.go:574: expected conn 10.1.0.4:47312 to be closed
```

It's probably just a race condition, so, fix that by waiting a bit.

Also, the way it was checking that conns were closed was by trying to
reuse them, seeing if that failed with a particular error message.
That's too implementation-dependent. Now it checks that the Close()
method is called on the connections.

### DIFF
--- a/ably/rest_client_test.go
+++ b/ably/rest_client_test.go
@@ -549,7 +549,7 @@ func TestFixConnLeak_ISSUE89(t *testing.T) {
 	}
 	defer app.Close()
 
-	var conns []*connTrackingClose
+	var conns []*connCloseTracker
 
 	httpClient := ablytest.NewHTTPClientNoKeepAlive()
 	transport := httpClient.Transport.(*http.Transport)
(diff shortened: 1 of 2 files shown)


sacOO7:
If the issue is fixed, CI shouldn't fail, right?

tcard:
> If the issue is fixed, CI shouldn't fail, right?

Right. Now it doesn't :)

ably/rest_client_test.go
```diff
@@ -567,16 +577,21 @@ func TestFixConnLeak_ISSUE89(t *testing.T) {
...
+			return atomic.LoadUintptr(&c.closed) != 0
+		}) {
+			t.Errorf("conn to %v wasn't closed", c.RemoteAddr())
 		}
 	}
 }
 
-func connIsClosed(err error) bool {
-	return strings.Contains(err.Error(), "use of closed network connection")
+type connTrackingClose struct {
```
sacOO7, line 589: Maybe we can rename it to `connCloseTracking` / `connCloseTracker`

tcard: Done at 506ab43.

PR #285, opened 2021-02-23
Added H2 with Resources

ramiro-nd:
Added resources with links to our Go content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,11 @@ This library uses [semantic versioning](http://semver.org/). For each release, t
 * Once the PR is approved, merge it into `main`
 * Add a tag to the new `main` head commit and push to origin such as `git tag v1.1.6 && git push origin v1.1.6`
 
+## Resources
+
+* [Building realtime apps with Go and WebSockets: client-side considerations](https://www.ably.io/topic/websockets-golang)
+* [Guide to Pub/Sub in Golang](https://www.ably.io/blog/pubsub-golang)
+
 ## Support and feedback
 
 Please visit http://support.ably.io/ for access to our knowledgebase and to ask for any assistance.
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-## [Ably Go](https://www.ably.io)
+## [Ably Go](https://ably.com/)
 
-A Go client library for [www.ably.io](https://ably.io), the realtime messaging service.
+A Go client library for [https://ably.com/](https://ably.com/), the realtime messaging service.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


mattheworiordan (approved):
On balance I think we need to be very careful about polluting our documentation with marketing content, but I think this content is relevant to the audience.

However, I think "Resources" is maybe not the right title as it implies (to me) that these are resources relating to the SDK itself.

BTW. I would put this above https://github.com/ably/ably-go#release-process, given the release process is aimed at devs building SDKs, whereas this is for consumers of the SDK.  You could move Support & feedback up too given that is the case.

QuintinWillison:
Sorry, I've only just passed by and seen Matt's comment here.

I'll add @tbedford as a reviewer as he has told me that he is planning to work on conforming read me documents.

In @ramiro-nd's defence, the "Resources" title does match what we currently have [in Flutter](https://github.com/ably/ably-flutter#resources). 😁 

Also, one of the tasks in the SDK Team's backlog is to move client library developer focussed read me content to a discrete `CONTRIBUTING` file - as we have already done [in ably-js](https://github.com/ably/ably-js#contributing).

tbedford (commented):
A few comments:

* Instead of Resources you could go with "Further information". It depends a bit on what you're linking to as to the best option. I'd probably go with the latter, but not a blocker.
* You could also add a line of text before the bullet points describing the nature of the resources and why someone should be interested in them.
* Shouldn't we be using `ably.com`? I'm confused about this - we changed a lot of links in docs to point at `ably.com`.

ramiro-nd:
@tbedford I actually created the PR just before the migration to .com, but you are right. I updated those and all the links to ably.com. Tedious thing to do but we should do it in every repo.

Let me know if it's good to go now.

README.md
```diff
@@ -43,7 +43,7 @@ connection.
 
 As with the REST API, it is intended that this library is upgraded incrementally and brought into line with the 1.1
 specification. If there are features that are currently missing that are a high priority for your use-case then please
-[contact Ably customer support](https://support.ably.io). Pull Requests are also welcomed.
+[contact Ably customer support](http://knowledge.ably.com/). Pull Requests are also welcomed.
```
QuintinWillison, line 46: Should be using `https`.

README.md
```diff
@@ -281,9 +281,16 @@ This library uses [semantic versioning](http://semver.org/). For each release, t
...
+
+Broaden your knowledge of realtime in Go with these useful materials:
+
+* [Building realtime apps with Go and WebSockets: client-side considerations](https://ably.com/topic/websockets-golang)
+* [Guide to Pub/Sub in Golang](https://ably.com/blog/pubsub-golang)
+
 ## Support and feedback
 
-Please visit http://support.ably.io/ for access to our knowledgebase and to ask for any assistance.
+Please visit http://knowledge.ably.com/ for access to our knowledgebase and to ask for any assistance.
```
QuintinWillison, line 293: Should be using `https`.

QuintinWillison (changes requested):
Thanks for making the unrelated changes to update `ably.io` references to `ably.com`. We're going to get that done in other repositories very soon.

tbedford (approved):

PR #286, opened 2021-02-25
Updated readme, fixed variable assignment and new pagination API's

sacOO7:
Updated README, fixed pagination API's


tcard:
@sacOO7 Can you please run `go generate ./...` so that readme_examples_test.go is updated?

(#287 should ensure we don't forget this again going forwards.)

PR #288, opened 2021-03-02
Channel resume

### LINKED ISSUE #237: RTL12: Random incoming ATTACHED message
Write spec tests for [RTL12](https://docs.ably.io/client-lib-development-guide/features/#RTL12), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

### LINKED ISSUE #243: RTL2f: ChannelEvent RESUMED flag
Write spec tests for [RTL2f](https://docs.ably.io/client-lib-development-guide/features/#RTL2f), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

sacOO7:
- Added channel resume flag implementation
- Fixes #243 
- Fixes #237


ably/readme_examples_test.go
```diff
@@ -1,13 +1,18 @@
 // Generated by test_readme_examples. DO NOT EDIT
 package ably_test
 
-import "testing"
-import "github.com/ably/ably-go/ably"
-import "github.com/ably/ably-go/ably/ablytest"
+import (
```
tcard, line 4: Please run `go generate ./...` and leave this leave unaltered afterwards. CI is failing because of this.

ably/proto/protocol_message.go
```diff
@@ -6,9 +6,17 @@ import (
 	"time"
 )
 
+// TR3
 const (
-	FlagPresence Flag = iota + 1
-	FlagBacklog
+	FlagHasPresence       Flag = 0
```
tcard, line 11: [TR3 says](https://docs.ably.io/client-lib-development-guide/features/#TR3):

> ProtocolMessage Flag enum has the following values, where a flag with value n is defined to be set if the bitwise AND of the flags field with 2ⁿ is nonzero

So those numbers are the _bit indexes_, starting at 0, whose corresponding bit being 1 means the flag is activated. They're not the _value of the bits_ themselves. For example, HAS_PRESENCE being 0 means that a Flag whose first bit is 1 will have the HAS_PRESENCE flag activated, but a Flag with value 0 doesn't.

Those Flag constants here aren't supposed to be bit those indexes, but bit _masks_, so that the `Has` method below works. So, they should be like this:

```go
FlagHasPresence Flag = 1 << 0
FlagHasBacklog Flag = 1 << 1
FlagResumed Flag = 1 << 2
FlagTransient Flag = 1 << 4
// etc.
```

I think naming doesn't help here. `Flag` type should be `Flags` or `FlagsField`; each individual bit in a `Flags`/`FlagsField` is a flag.

ably/realtime_channel_spec_test.go
```diff
@@ -469,6 +469,97 @@ func TestRealtimeChannel_RTL6c5_NoImplicitAttach(t *testing.T) {
...
+		channel.OnAll(stateChanges.Receive)
+
+		return
+	}
+
+	flags := make(map[proto.Flag]string)
+	flags[0] = "flag has presence is provided"
+	flags[1] = "flag has_backlog is provided"
+	flags[2] = "flag resumed is provided"
+	flags[4] = "flag transient is provided"
```
tcard, line 526: I don't really see the point of testing all possible flags fields in which the RESUMED flag isn't set.

ably/realtime_channel_spec_test.go
```diff
@@ -469,6 +469,97 @@ func TestRealtimeChannel_RTL6c5_NoImplicitAttach(t *testing.T) {
...
+
+		err := ablytest.Wait(ablytest.ConnWaiter(c, c.Connect, ably.ConnectionEventConnected), nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		channel = c.Channels.Get("test")
+
+		ctx, cancel := context.WithCancel(context.Background())
+		go channel.Attach(ctx)
```
tcard, line 511: Please don't use plain `go` in tests. It's easy to miss goroutine leaks this way. Use `ablytest.ResultGroup` with its `GoAdd` method instead, then call `Wait` once the ATTACHED message is received.

Alternatively, since you're not interested in the result of `Attach` anyway, you can call it with a cancelled context. For realtime operations, [a context cancel just signals the operation shouldn't block](https://github.com/ably/ably-go/issues/197#issuecomment-696979466).

tcard (changes requested):

ably/proto/protocol_message.go
sacOO7: You mean those numbers represent one or a group of bit indexes that should be present in the incoming number value with bit indexes at the same position.

tcard: > You mean those numbers represent one or a group of bit indexes that should be present in the incoming number value with bit indexes at the same position.

Yes, exactly. It's like indexes in an array of booleans. Except the array of booleans is packed in a single integer of 64 bits (ie. 64 booleans).

> I feel I have come across this type of encoding where one number can represent multiple flags at the same time.

It's a common data structure called [bit field](https://en.wikipedia.org/wiki/Bit_field).

ably/realtime_channel_spec_test.go
sacOO7: Okay, I will remove some of the flags, and will just keep a few ( to check negative test cases)

ably/proto/protocol_message.go
sacOO7: Okay, so I feel the current solution should be OK. I can see that's how it's implemented in other clients too.

tcard (approved):

sacOO7:
@tcard, I have updated PR to fix #237, both of them are strongly related. You can review the changes again.

tcard (approved):
RTL12 says:

> If **and only if** the resumed flag is false, this should result in the channel emitting an UPDATE event with a ChannelStateChange object.

Can we add a test in which re-attach carries the RESUMED flag, and no event is emitted?

sacOO7:
@tcard as per [RTL2g](https://docs.ably.io/client-lib-development-guide/features/#RTL2g), update event will always be emitted for current state equal to prev. state. I don't think it has anything to do with a resume true or false. I have seen similar behaviour in other clients. ( Currently, I am referring to standard dotnet implementation)

tcard:
The spec clearly could be less ambiguous here. I've opened https://github.com/ably/docs/pull/1061.

Anyway, this doesn't seem so important so as to block on it.

tcard (approved):

lmars (approved):

sacOO7:
Merging the PRs since it's already approved : )

PR #290, opened 2021-03-05
Run single Go version in CI

tcard:
Many PRs now look like this:

Screenshot 2021-03-05 at 08 41 37

We can extract the following information from that:

1. The test suite _can_ pass, more often than not.
2. Since it passes in Go 1.13, and Go versions are guaranteed to be backwards-compatible, failures in Go 1.14 and 1.15 aren't due to incompatibilities.
3. Since, by design, most of our tests target real Ably connections, they're inherently flaky; chances are the PR being tested didn't cause the flakiness that's making tests fail, and, if it's bad enough, can be dealt with separately.

Following this, we can conclude that this PR is probably good to merge.

Exactly the same information can be extracted from just testing with Go 1.13, and at the end we'll have a nice green checkmark aligned with the conclusion that the PR is good. So I propose doing that instead.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.13, 1.14, 1.15]
+        go-version: [1.13]
 
     steps:
       - uses: actions/checkout@v2


tcard:
In particular, most current opened PR with failing CI are due to this error:

```
panic: Post https://sandbox-rest.ably.io/apps: net/http: request canceled (Client.Timeout exceeded while awaiting headers) [recovered]
```

This is sandbox failing to create an app in time, and that's something libraries can do nothing about.

paddybyers:
A flaky test that causes 1 in 3 versions to fail will still cause 1 in 3 CI jobs to fail if we only test a single version. So we have to try to eliminate all causes of flaky tests - in this case perhaps increasing the timeout.

I do agree that it's overkill to test against lots of go versions. At any time there are only two supported minor versions anyway.

tcard:
@paddybyers The timeout is 1 minute, which I think should be plenty. I don't know what's causing the timeouts, but I suspect we can't do anything about it besides reducing paralellism, which may be affecting things. Testing just one Go version does just that for no cost, and may avoid us the cost of having to investigate those timeouts.

cc @QuintinWillison

QuintinWillison:
Thanks, @tbedford. But, on reflection, I'm not happy that this pull request addresses the problem that's actually being faced. Opinions on matrix testing are a separate concern.

I would prefer to see work done to work out why operations appear to be hanging. This could be something to do with the way the tests are constructed, or I do wonder whether it could be a sign of something deeper going on inside the library implementation. Either way, I don't want to be hiding from this. If anything, we should be _adding_ Go 1.16 to the matrix!

PR #295, opened 2021-03-11
RTN24, Handle Random Connected Message

### LINKED ISSUE #223: RTN24: Handle random CONNECTED message
Write spec tests for [RTN24](https://docs.ably.io/client-lib-development-guide/features/#RTN24), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

sacOO7:
- Added test for a random connected message with new connection details
- Fixes #223

### DIFF
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -127,10 +127,8 @@ func (a *Auth) clientIDForCheck() string {
 func (a *Auth) updateClientID(clientID string) {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
-	if clientID != "" {
-		//Spec RSA7b3, RSA7b4, RSA12a,RSA12b, RSA7b2,
-		a.clientID = clientID
-	}
+	//Spec RSA7b3, RSA7b4, RSA12a,RSA12b, RSA7b2,
+	a.clientID = clientID
 }
 
 // CreateTokenRequest
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -127,7 +127,7 @@ func (a *Auth) clientIDForCheck() string {
 func (a *Auth) updateClientID(clientID string) {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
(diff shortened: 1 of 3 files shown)


ably/auth.go
```diff
@@ -127,7 +127,7 @@ func (a *Auth) clientIDForCheck() string {
 func (a *Auth) updateClientID(clientID string) {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
-	if a.clientID == "" {
+	if clientID != "" {
```
tcard, line 130: Could you explain this change? In particular, this seems to preclude the possibility of the library being identified and then becoming unidentified. But [RSA12a] seems to require that this may happen.

(Not saying that the condition before is right. It seems wrong too, per RSA7b.)

ably/realtime_conn_spec_test.go
```diff
@@ -2358,3 +2358,97 @@ func TestRealtimeConn_RTN19a(t *testing.T) {
...
+		ClientID:           "id2",
+		ConnectionKey:      "bar",
+		MaxFrameSize:       13,
+		MaxInboundRate:     15,
+		MaxMessageSize:     70,
+		ConnectionStateTTL: proto.DurationFromMsecs(time.Minute * 2),
+		MaxIdleInterval:    proto.DurationFromMsecs(time.Second),
+	}
+
+	// Re-attach the channel
```
tcard, line 2405:
```suggestion
	// Send new connection details
```

But I would move this comment to line 2416 or 2395, or just remove it.

ably/auth.go
sacOO7: I think we can remove the check. We don't need to it in order to conform to both cases. The check shouldn't have been there from the start, since we need to set whatever value we received as clientId

tcard (approved):

lmars (approved):

PR #297, opened 2021-03-12
Connection Close

### LINKED ISSUE #262: RTN12: Connection.close
Write spec tests for [RTN12](https://docs.ably.io/client-lib-development-guide/features/#RTN12), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

sacOO7:
- Fixes #262

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -709,7 +709,7 @@ func (c *Connection) eventloop() {
 		if err != nil {
 			c.mtx.Lock()
 			if c.state == ConnectionStateClosing {
-				// RTN12b
+				// RTN12b, RTN12c
 				c.lockSetState(ConnectionStateClosed, err, 0)
 				c.mtx.Unlock()
 				return
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -133,15 +133,15 @@ func (c *Connection) connectAfterSuspension(arg connArgs) (result, error) {
 	retryIn := c.opts.suspendedRetryTimeout()
 	lg := c.logger().sugar()
 	lg.Debugf("Attemting to periodically establish connection after suspension with timeout %v", retryIn)
-	parentCtx, cancel := c.ctxCancelOnStateTransition()
-	defer cancel()
+	parentCtx, parentCtxCancel := c.ctxCancelOnStateTransition()
+	defer parentCtxCancel()
 	tick := ablyutil.NewTicker(c.opts.After)(parentCtx, retryIn)
 	for {
 		_, ok := <-tick
 		if !ok {
 			return nil, parentCtx.Err()
 		}
-		res, err := c.connectWith(arg)
+		res, err := c.connectWith(arg, parentCtxCancel)
 		if err != nil {
 			if recoverable(err) {
 				c.setState(ConnectionStateSuspended, err, retryIn)
@@ -289,7 +289,7 @@ const connectionStateTTLErrFmt = "Exceeded connectionStateTtl=%v while in DISCON
 
 func (c *Connection) connectWithRetryLoop(arg connArgs) (result, error) {
 	lg := c.logger().sugar()
-	res, err := c.connectWith(arg)
+	res, err := c.connectWith(arg, nil)
 	if err == nil {
 		return res, nil
 	}
@@ -301,8 +301,8 @@ func (c *Connection) connectWithRetryLoop(arg connArgs) (result, error) {
 	retryIn := c.opts.disconnectedRetryTimeout()
 	c.setState(ConnectionStateDisconnected, err, retryIn)
 
-	parentCtx, cancel := c.ctxCancelOnStateTransition()
-	defer cancel()
+	parentCtx, parentCtxCancel := c.ctxCancelOnStateTransition()
+	defer parentCtxCancel()
 
 	// The initial DISCONNECTED event has been fired. If we reach stateTTL without
 	// any state changes, we transition to SUSPENDED state
@@ -342,7 +342,7 @@ loop:
 			}
 
 			lg.Debug("Attemting to open connection")
-			res, err := c.connectWith(arg)
+			res, err := c.connectWith(arg, parentCtxCancel)
 			if err == nil {
 				return res, nil
 			}
@@ -368,7 +368,13 @@ loop:
 	return c.connectAfterSuspension(arg)
 }
 
-func (c *Connection) connectWith(arg connArgs) (result, error) {
+func (c *Connection) connectWith(arg connArgs, ctxCancel context.CancelFunc) (result, error) {
+	c.mtx.Lock()
+	// set ably connection state to connecting, connecting state exists regardless of whether raw connection is successful or not
+	if !c.isActive() { // check if already in connecting state
+		c.lockSetState(ConnectionStateConnecting, nil, 0)
+	}
+	c.mtx.Unlock()
 	u, err := url.Parse(c.opts.realtimeURL())
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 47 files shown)


ably/realtime_conn_spec_test.go
```diff
@@ -183,6 +183,245 @@ func connectAndWait(t *testing.T, realtime *ably.Realtime) {
...
+			if expected, got := ably.ConnectionStateClosed, change.Current; expected != got {
+				t.Fatalf("expected %v; got %v (event: %+v)", expected, got, change.Current)
+			}
+			ablytest.Instantly.NoRecv(t, nil, stateChange, t.Fatalf)
+		})
+	}
+
+	t.Skip("RTN12f: transition to closed when close is called intermittently")
+	// todo - Need to fix the test and check for possible race conditions
+	t.Run("RTN12f: transition to closed when close is called intermittently", func(t *testing.T) {
```
sacOO7, line 386: @tcard you can check the test, I don't think I am missing anything here. Not sure why still it's not working : (

ably/realtime_conn.go
```diff
@@ -958,8 +958,10 @@ func (c *Connection) lockSetState(state ConnectionState, err error, retryIn time
 func (c *Connection) ctxCancelOnStateTransition() (context.Context, context.CancelFunc) {
 	ctx, cancel := context.WithCancel(context.Background())
 
-	off := c.internalEmitter.OnceAll(func(ConnectionStateChange) {
-		cancel()
+	off := c.internalEmitter.OnceAll(func(change ConnectionStateChange) {
+		if change.Current == ConnectionStateConnected || change.Current == ConnectionStateClosed {
```
tcard, line 962: The purpose of this function is to stop the timer that moves the connection state from DISCONNECTED or SUSPENDED to CONNECTING. This should happen when:

* `connect` is called ([RTN11c](https://docs.ably.io/client-lib-development-guide/features/#RTN11c)).
* `close` is called ([RTN12d](https://docs.ably.io/client-lib-development-guide/features/#RTN12d)).

This isn't captured by this change; `connect` causes a transition to CONNECTING, and `close` causes a transition to CLOSING. Maybe those were the states 

But there's a deeper issue here: the context is meant to be "live" for as along as we're _waiting_ for the timer, but it actually it can live longer than that. [After the timer fires](https://github.com/ably/ably-go/blob/43053b50d6f8e504c289df4ace37f46d195d2395/ably/realtime_conn.go#L344), this context should be dismissed, but it isn't; it will still be used on the next iteration of the loop if reconnection fails, and it will be _cancelled_ due to the transition to CONNECTING that happened while reconnecting, as you pointed out in the chat.

So my suggestion would be to move the body of the `for` loop in `connectWithRetryLoop` and  `connectAfterSuspension` to separate functions, and call `c.ctxCancelOnStateTransition()` there, so that the context is created afresh on each attempt to connect. (It would also make things simpler if `next` and `nextCtx` are moved inside this new function.)

sacOO7: Okay, maybe there is confusion here. the purpose of the function is to stop reconnection when the state goes into the desired state. The primary state would be `CONNECTED` and not `CONNECTING`. The same goes for `CLOSED` and not `CLOSING`. There's no point in closing the timer when reconnection doesn't fulfill its purpose, that is to connect to the server

sacOO7: @tcard, I have updated the reconnection logic, I think the only `connectionStateConnecting` wasn't getting updated at the right place. Everything else looks fine to me (including the creation of parent context)

lmars (approved):

ably/realtime_conn.go
```diff
@@ -393,18 +393,32 @@ func (c *Connection) connectWith(arg connArgs) (result, error) {
...
+		return nopResult, nil
+	}
+
+	// if err is nil, raw connection with server is successful
 	conn, err := c.dial(proto, u)
 	if err != nil {
 		return nil, err
+	} else {
+		if ctxCancel != nil {
+			ctxCancel() // break reconnection loop, eventloop will be responsible for next reconnection iterations
```
tcard, line 407: Maybe I'm missing something but I don't see what is this accomplishing.

`ctxCancel()` is called here when `err == nil`, to break the reconnection loop. But the reconnection loop is broken anyway, because when `connectWith`  returns a nil error, we `return` and stop the loop, in both cases:

```go
res, err := c.connectWith(arg, parentCtxCancel)
if err == nil {
    return res, nil
}
```

```go
res, err := c.connectWith(arg, parentCtxCancel)
if err != nil {
	if recoverable(err) {
		c.setState(ConnectionStateSuspended, err, retryIn)
		continue
	}
	return nil, c.setState(ConnectionStateFailed, err, 0)
}
return res, nil
```

ably/realtime_conn.go
```diff
@@ -950,10 +969,15 @@ func (c *Connection) lockSetState(state ConnectionState, err error, retryIn time
...
 
 func (c *Connection) ctxCancelOnStateTransition() (context.Context, context.CancelFunc) {
+	// Returns parent context for retry process
 	ctx, cancel := context.WithCancel(context.Background())
 
-	off := c.internalEmitter.OnceAll(func(ConnectionStateChange) {
-		cancel()
+	off := c.internalEmitter.OnceAll(func(change ConnectionStateChange) {
+		// RTN12d - cancel the retry when connectionStateClosed
+		if change.Current == ConnectionStateClosed {
```
tcard, line 977: I'll copy the relevant part from my previous comment:

> The purpose of this function is to stop the timer that moves the connection state from DISCONNECTED or SUSPENDED to CONNECTING. This should happen when:
> 
> * `connect` is called ([RTN11c](https://docs.ably.io/client-lib-development-guide/features/#RTN11c)).
> * `close` is called ([RTN12d](https://docs.ably.io/client-lib-development-guide/features/#RTN12d)).
> 
> This isn't captured by this change; `connect` causes a transition to CONNECTING, and `close` causes a transition to CLOSING. Maybe those were the states you meant to put here?

tcard (changes requested):
See my previous review.

ably/realtime_conn.go
sacOO7: Okay, it seems it stops the loop either way, maybe I missed on that one.

ably/realtime_conn.go
tcard: > Currently, we are not using `retry` value anywhere, it's just a placeholder.

Not true, it's used in the `RetryIn` property of the `ConnectionEvent`: https://github.com/ably/ably-go/blob/083f8e334bc0cb55791f88223ed1c4e99a39b5ec/ably/realtime_conn.go#L912

> I find this solution more feasible because in other cases, things just makes code more messy

Sorry, I didn't make my suggestions clear enough because they actually simplify all this logic quite a bit. For instance:

> That way, it's easy to detect from where `connecting` is being called.

This just isn't necessary. The only reason it's necessary as things stand now is because of poorly factored code.

So, given I didn't manage to explain myself with words, here's my proposal in code: https://github.com/ably/ably-go/compare/fix/262-connection-close...fix/262-connection-close-alt As I hope you'll agree with, the overall logic is way simpler now, with less things interacting in non-obvious ways, with more explanatory naming, and more comments.

I'd really strongly prefer that to be the way forward. If you don't agree, I'll need to ask for a third opinion before reaching a decision.

> and we might need to go away from spec

This may be so, depending on the interpretation on what "aborts the retry process" means. I believe https://github.com/ably/docs/pull/1090 is a better way of defining what RTN11c is _actually_ trying to accomplish.

sacOO7: No. it's just writing the value to `RetryIn`, which is literally not being used anywhere. So, technically, it means, it's not being used. I checked at all places, we just write the value and not read it for doing any processing. In the new code, we will use 
it for cancelling the loop

sacOO7: I need to properly review your code, if it works, I am fine with it.

tcard: > No. it's just writing the value to `RetryIn`, which is literally not being used anywhere. So, technically, it means, it's not being used. I checked at all places, we just write the value and not read it for doing any processing.

It's part of the public API. We're not the ones using it, clients are.

sacOO7: Also, in the alt code, I think we might need to tweak it a bit, since it doesnt cover cancelling the loop when closed is called. There is explicit check to return nil error, but before that we set state to `connecting`, so it doesn't really cover the case. I am waiting for someone to accept the spec, if new spec changes are considered, I am happy to merge your code. Although, its not straightforward and combines disconnected & suspended spec in one loop. I would generally consider `WET ( write everything twice)` principal in the case, because it makes code easy to understand.

tcard: > Also, in the alt code, I think we might need to tweak it a bit, since it doesnt cover cancelling the loop when closed is called. There is explicit check to return nil error, but before that we set state to `connecting`, so it doesn't really cover the case.

Oops, forgot about that. Addressed at bd8c500.



> I am waiting for someone to accept the spec, if new spec changes are considered, I am happy to merge your code.

@ably/client-library-sdk-admins Maybe we should merge https://github.com/ably/docs/pull/1090 it already? It has several approvals.



> Although, its not straightforward and combines disconnected & suspended spec in one loop. I would generally consider `WET ( write everything twice)` principal in the case, because it makes code easy to understand.

I decided to do it in one loop because the only change between DISCONNECTED and SUSPENDED is the waiting time between reconnection attempts, and that can be just a variable. In contrast, the logic around timer cancellation is a bit more complex and is the same, so I'd rather not write it twice.

tcard: > No it's not being used clients. I checked, clients just write value to retryIn and not read from it

I meant end-users. They can access the field when they get a ConnectionEvent.


> > I am waiting for someone to accept the spec, if new spec changes are considered, I am happy to merge your code.
> 
> @ably/client-library-sdk-admins Maybe we should merge [ably/docs#1090](https://github.com/ably/docs/pull/1090) it already? It has several approvals.

I just realized it's been merged for a few hours now.

tcard (approved):

sacOO7:
Merging the PR, since approved by both parties : )

PR #298, opened 2021-03-16
Channel Detach

### LINKED ISSUE #212: RTL5: Channel.detach
Write spec tests for [RTL5](https://docs.ably.io/client-lib-development-guide/features/#RTL5), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

sacOO7:
- Fixes #212


sacOO7:
Most of the spec items are covered as a part of tests, till pending/skipped spec test item test is written, you can take a look @tcard

ably/realtime_channel.go
```diff
@@ -279,6 +305,23 @@ func (c *RealtimeChannel) detachUnsafe() (result, error) {
 	return res, nil
 }
 
+func (c *RealtimeChannel) detachUnsafe() (result, error) {
+	if c.state == ChannelStateAttaching { //RTL5i
```
tcard, line 309: Why isn't this part of the `detach` function? All other checks are there and `detachUnsafe` is always called from there, except possibly on reconnection, in which case the channel will be always DETACHING.

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +155,662 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		channel.OnAll(stateChanges.Receive)
+
+		// Initialized state
+		channel.Detach(ctx)
+		ablytest.Instantly.NoRecv(t, nil, out, t.Fatalf)          // Should not send detach message
+		ablytest.Instantly.NoRecv(t, nil, stateChanges, t.Fatalf) // Should not make any change to the state
+
+		channel.SetState(ably.ChannelStateDetached, nil)
```
tcard, line 212: Same point as here https://github.com/ably/ably-go/pull/299#discussion_r599334068

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +155,662 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		channel = c.Channels.Get("test")
+		stateChanges = make(ably.ChannelStateChanges, 10)
+		return
+	}
+
+	t.Run("RTL5a", func(t *testing.T) {
```
tcard, line 201: It would be nice to have a little summary of the spec item being tested, such as: "RTL5a: no-op when initialized or detached".

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +155,662 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+
+		ablytest.Instantly.Recv(t, &change, stateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateAttached, change.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, change)
+		}
+
+		var outMsg *proto.ProtocolMessage
+
+		go func() {
+			time.Sleep(time.Second / 2)
```
tcard, line 259: What's the point of waiting? This only makes the test flaky, while you control all events anyway since the connection transport is mocked. You can just do this below the DETACHING check synchronously.

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +155,662 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
```
tcard, line 313: Same as above.

ably/realtime_channel.go
```diff
@@ -240,19 +246,46 @@ func (c *RealtimeChannel) lockAttach(err error) (result, error) {
...
+
+	detachTimeout := c.client.Connection.opts.realtimeRequestTimeout()
+	select {
+	case err := <-resultErr: // RTL5e
+		return err
+	case <-time.After(detachTimeout): // RTL5f
+		err := newError(ErrTimeoutError, errors.New("timed out before detaching channel"))
+		c.setState(prevChannelState, err)
+		return err
+	}
```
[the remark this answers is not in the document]
tcard: Now that I think of it, we shouldn't use `context.WithTimeout` directly, because it bypasses ClientOptions.After and thus prevents us from controlling timers from tests.

We need a custom one, something like this:

```go
func (o *clientOptions) withTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
	ctx, cancel := context.WithCancel(ctx)
	go func() {
		<-o.After(ctx, timeout)
		cancel()
	}()
	return ctx, cancel
}
```

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +155,662 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		var change ably.ChannelStateChange
+
+		ablytest.Soon.Recv(t, &change, stateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateAttached, change.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, change)
+		}
+
+		c.Close()
+		// set connection state to closing
+		c.Connection.SetState(ably.ConnectionStateClosing, nil, time.Minute)
```
tcard, line 426: Same point as here https://github.com/ably/ably-go/pull/299#discussion_r599334068

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +155,662 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		ablytest.Instantly.Recv(t, &channelChange, channelStateChange, t.Fatalf) // Consume attached message
+
+		if expected, got := ably.ChannelStateAttached, channelChange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelChange)
+		}
+
+		var change ably.ConnectionStateChange
+
+		// set connection state to ConnectionStateInitialized
+		c.Connection.SetState(ably.ConnectionStateInitialized, nil, time.Minute)
```
tcard, line 475: Same point as here https://github.com/ably/ably-go/pull/299#discussion_r599334068

This is especially trivial to avoid, since INITIALIZED is the connection's initial state when autoConnect=false.

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +155,662 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+
+		ablytest.Instantly.Recv(t, &channelChange, channelStateChange, t.Fatalf) // Should timeout and reset to attached since detach message is queued
+
+		if expected, got := ably.ChannelStateAttached, channelChange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelChange)
+		}
+
+		ablytest.Instantly.NoRecv(t, &channelChange, channelStateChange, t.Fatalf) // Shouldn't receive detached state
+
+		c.Connection.SetState(ably.ConnectionStateConnecting, nil, time.Minute)
```
tcard, line 498: Same point as here https://github.com/ably/ably-go/pull/299#discussion_r599334068

ably/realtime_channel.go
sacOO7: Okay, if that's the case where for reconnection channel state is always detaching, I will move it to detach function

ably/realtime_channel.go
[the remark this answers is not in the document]
sacOO7: Okay, since the second solution already contains goroutine, I can directly use that function in the current implementation.
Replace `time.After(detachTimeout)` with `c.opts().After(context.Background(), detachTimeout)`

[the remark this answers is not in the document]
sacOO7: It seems, once I change it, current transitioner tests are breaking, also as far as detach is concerned, it seems to be working without the need of injecting an extra `option.After` function from outside.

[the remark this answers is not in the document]
sacOO7: So, I will keep it the way it is for the time being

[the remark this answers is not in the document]
tcard: > It seems, once I change it, current transitioner tests are breaking, also as far as detach is concerned, it seems to be working without the need of injecting an extra option.After function from outside.

As discussed, I've addressed this in f769f68779b7c85e35164df24b67bcc94ed82232 and c6d3c3d3d786b866b8d520e480fb0dd5a691e372.

ably/internal/ablyutil/time.go
```diff
@@ -75,21 +72,19 @@ func ContextWithTimeout(ctx context.Context, after TimerFunc, timeout time.Durat
 
 type contextWithCustomErr struct {
 	context.Context
-	err <-chan error
```
tcard: This was a buffered channel for a reason: reading and writing the error happens on different goroutines, so it needs to be synchronized.

At the very least, if you really don't like the channel, add a sync.Mutex.

ably/internal/ablyutil/time.go
```diff
@@ -57,16 +57,13 @@ func ContextWithTimeout(ctx context.Context, after TimerFunc, timeout time.Durat
...
+	ctx, setErr := contextWithCustomError(ctx)
 
 	go func() {
 		_, timerFired := <-after(ctx, timeout)
 		if timerFired {
 			setErr(context.DeadlineExceeded)
-		} else {
-			// after returned because the parent context was canceled, not
-			// because the timer fired.
-			setErr(ctx.Err())
```
tcard: This was also necessary: if the timer didn't fire, then `ctx.Err()` is the underlying error that caused `after` to return, and it needs to be propagated.

tcard (changes requested):

tcard:
@sacOO7 Unfortunately this PR now includes multiple commits that seem to have been rebased from `integration/1.2`. Can you please push with just the commits that are actually new?

ably/internal/ablyutil/time.go
sacOO7: Yes, that's being taken care of here 
```func (ctx *contextWithCustomErr) Err() error {
	if ctx.err != nil {
		return ctx.err
	}
	return ctx.Context.Err()```

File: ably/internal/ablyutil/time.go (hunk 10 above)
Comment (sacOO7) [reply ably/internal/ablyutil/time.go]: No, it won't cause, simultaneous read and write. When the timer is fired, context is not canceled,
so `ctx.Done()` is not yet called, once we set the error, then we cancel the context.

File: ably/internal/ablyutil/time.go (hunk 11 above)
Comment (tcard) [reply ably/internal/ablyutil/time.go]: Discussed offline.

File: ably/internal/ablyutil/time.go (hunk 10 above)
Comment (tcard) [reply ably/internal/ablyutil/time.go]: Discussed offline; decided OK to leave as-is for the purposes of this PR.

File: ably/realtime_channel.go (hunk 6 above)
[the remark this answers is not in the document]
Comment (sacOO7) [reply ably/realtime_channel.go]: I think we can close this one, since it's already addressed

File: ably/realtime_channel_spec_test.go (hunk 8 above)
Comment (sacOO7) [reply ably/realtime_channel_spec_test.go:617]: I am not sure how to test this, because it's difficult to reproduce the connection state to Initialized once the connection
is established with the server. In order to run the test for detach, it needs to be attached first, hence connected in the first place. You cannot go back to initialized once connected. Maybe, we should go back to the point we raised about attach sometime back.

Comment (sacOO7) [reply ably/realtime_channel_spec_test.go:617]: https://ably-real-time.slack.com/archives/C0114KK7S6R/p1617842973002400

Comment (sacOO7) [reply ably/realtime_channel_spec_test.go:617]: We should remove initialized condition from the spec I feel, there's no point in having it there ... since practically it's impossible to reach the state at least for detach operation

File: ably/realtime_channel_spec_test.go [hunk 12]
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		}
+
+		ablytest.Soon.Recv(t, &channelStatechange, channelStateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateDetached, channelStatechange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+	})
+
+	// Suspend not implemented, so directly setting the state
+	t.Run("RTL5j: if channel state is SUSPENDED, immediately transition to DETACHED state", func(t *testing.T) {
```
sacOO7, line 880: @tcard this one isn't practically producible, because channel suspended is not implemented

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		var change ably.ChannelStateChange
+		ablytest.Instantly.Recv(t, &change, stateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateDetached, change.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, change)
+		}
+		ablytest.Instantly.NoRecv(t, &change, stateChanges, t.Fatalf)
+	})
+
+	// Hard to reproduce, because channel ideally should be in attaching when attach is received, and channel detach waits for attaching state
+	t.Run("RTL5k: When receive ATTACH in detaching state, send new DETACH message", func(t *testing.T) {
```
sacOO7, line 902: @tcard This one as per the spec. I wasn't able to find any reference implementation from the Dotnet client. Same as suspend, this one can't be produced via transitioner or custom dials.

ably/realtime_channel_spec_test.go
tcard: > In order to run the test for detach, it needs to be attached first

Not in this case. This spec item covers what happens if you call `detach` without having attached first.

That said, as discussed in private, I agree it would make more sense if the operation failed immediately.

Anyway, for the purposes of this test, this state can be reached just with:

```go
r, _ := ably.NewRealtime(... WithAutoConnect(false) ...)
c := r.Channels.Get("foo")
c.Detach(ctx)
```

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		if expected, got := ably.ChannelStateDetached, channelStatechange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		err = channel.Detach(ctx)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		ablytest.Instantly.NoRecv(t, nil, channelStateChanges, t.Fatalf) // Should not make any change to the channel state
```
tcard, line 270: I suggest adding a check that we don't send a DETACH message, which is the other effect that could take place.

This applies to other similar tests too.

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		ablytest.Instantly.Recv(t, &channelStatechange, channelStateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateDetaching, channelStatechange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		ablytest.Instantly.Recv(t, &channelStatechange, channelStateChanges, t.Fatalf)
```
tcard, line 384:
```suggestion
		ablytest.Soon.Recv(t, &channelStatechange, channelStateChanges, t.Fatalf)
```

`Instantly` really means that it should happen synchronously, with no wait except for CPU. This actually waits for Ably servers.

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+
+		ablytest.Soon.Recv(t, &connectionChange, connectionStateChanges, t.Fatalf)
+		if expected, got := ably.ConnectionStateConnected, connectionChange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, connectionChange)
+		}
+
+		channel := client.Channels.Get("test")
+		channelStateChanges := make(ably.ChannelStateChanges, 10)
+
+		ctx, cancel := context.WithCancel(context.Background())
```
tcard, line 351: This doesn't seem useful. We're canceling at the end of the test, when the operations we have cancelled necessarily have returned already.

This applies to several other instances too.

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		err = channel.Detach(ctx)
+		if expected, got := "cannot Detach channel because connection is in CLOSING state", err.Error(); !strings.Contains(got, expected) {
+			t.Fatalf("expected error %+v; got %v", expected, got)
+		}
+
+		close = c.To(
+			failed,
+		)
+
+		defer safeclose(t, close)
```
tcard, line 576: This is the only `close` function from the transitioner we should be calling. The rest are superseded by subsequent calls to `c.To`.

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		ablytest.Instantly.Recv(t, &channelStatechange, channelStateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateAttached, channelStatechange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		close = c.To(
+			disconnected,
```
tcard, line 715: This is not _wrong_, but it seems unnecessary. Why don't call `detach` while the connection is first CONNECTING?

It seems you're assuming all these tests should apply when the channel has been attached at some point, but that's not really relevant in some cases, such as this one.

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		cancel()
+
+		channel.Detach(ctx)
+
+		ablytest.Instantly.Recv(t, &channelStatechange, channelStateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateDetaching, channelStatechange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		ablytest.Instantly.NoRecv(t, nil, channelStateChanges, t.Fatalf) // Shouldn't receive detach, detaching message queued
```
tcard, line 729: But you wouldn't get the DETACHED event `Instantly` in any case.

I suggest testing here that no DETACH is sent to the server.

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
```
tcard, line 799: Same as above:



> But you wouldn't get the DETACHED event Instantly in any case.
> 
> I suggest testing here that no DETACH is sent to the server.

ably/realtime_channel_spec_test.go
```diff
@@ -155,7 +155,804 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		defer rg.Wait()
+		rg.GoAdd(func(ctx context.Context) error {
+			err := channel.Detach(ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
+			return err
+		})
+
+		ablytest.Instantly.NoRecv(t, nil, channelStateChanges, t.Fatalf) // Shouldn't send detach waiting to get attached
```
tcard, line 859: As above, I suggest testing here that no DETACH is sent to the server.

ably/realtime_channel_spec_test.go
tcard: Then maybe just call `t.Skip`? We do that in some other cases already.

ably/realtime_channel_spec_test.go
tcard: Why not? The test controls the incoming messages with the `in` channel. Why not just move to DETACHING normally, and at that point send an ATTACHED, see that a new DETACH is sent on `out`?

ably/export_test.go
```diff
@@ -149,3 +149,11 @@ func (c *RESTChannels) Len() int {
 	defer c.mu.RUnlock()
 	return len(c.cache)
 }
+
+func (c *RealtimeChannel) SetState(state ChannelState, err error) error {
```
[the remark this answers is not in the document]
tcard: Correction: it is used, but it shouldn't be used by the time this PR is ready.

ably/realtime_channel_spec_test.go
sacOO7: when the context is cancelled right away, attach will return an error right away. I don't want it to happen. Attach should not return an error if it's a successful attach. I have added an error nil check for the same.

ably/realtime_channel_spec_test.go
sacOO7: Well, detach operation is already successful and it has already pushed the states in the buffered channel. You can check I have added nil check for err returned by detach which means detach will wait till channel is detached.

ably/realtime_channel_spec_test.go
sacOO7: Yeah, makes sense

ably/realtime_channel_spec_test.go
sacOO7: Okay, makes sense : ), thanks for the explaination

ably/realtime_channel_spec_test.go
sacOO7: Well, from the spec point of view, this is a positive detach case, In case, where the channel is not attached, and the connection state is connecting, the request will fail anyway, I feel that's unlikely, people will call detach before attach, anyways this point covers, it should queue detach and send after connecting and doesn't care what happens after that, Just makes it easy to test when channel transition to detached after connecting to connected

ably/realtime_channel_spec_test.go
sacOO7: Okay, actually I had a thought of having such a check at certain places. Can you guide me, how can we have such a check-in place using transitioner ?

ably/realtime_channel_spec_test.go
sacOO7: Okay, I will mark it as skip and todo in the future as a part of suspend functionality

ably/realtime_channel_spec_test.go
sacOO7: Although we can reproduce it locally, it seems highly unlikely in real case. For this to happen, ably needs to send attached message, it can only send it when local state is in attaching state and attach message is sent ... even if times out and state transitions to suspended, for detaching to be set by detach .. it has to be in attached state, for suspended it will just return detached directly. I cant test it, but his is a very hypothetical case I feel ... there are good chances of this to happen when ably servers start tripping and send random attach messages,..

ably/transitioner_test.go
```diff
@@ -312,6 +373,36 @@ func (c ChanTransitioner) To(path ...ably.ChannelState) (*ably.RealtimeChannel,
 	})
 }
 
+// preserves state context for channel
+func (c *ChanTransitioner) ToSpecifiedState(path ...ably.ChannelState) (*ably.RealtimeChannel, io.Closer) {
```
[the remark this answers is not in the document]
sacOO7: I think earlier version of `to` is already being used at places ..,. we need both for the time being, otherwise we need to refactor the code where earlier version of to is being used. I can rename it to something else for the time being

ably/realtime_channel_spec_test.go
tcard: It is a corner case, but a well-defined (and well-behaved) one that needs to be supported. The server may send an ATTACHED at any point ([RTL12](https://docs.ably.io/client-lib-development-guide/features/#RTL12), no need to be trippin' :) ), and it just may happen that it crosses paths with a DETACH from the client.

In any case, it's in the spec, so it needs to be supported.

ably/realtime_channel_spec_test.go
tcard: My point is that you can just pass `context.Background()` in these cases. It has exactly the same behavior, since we're not canceling those contexts until the test is done.

Anyway, it's a minor point, just to make the test a bit simpler.

ably/realtime_channel_spec_test.go
tcard: You can get a recorder with `ablytest.NewMessageRecorder`, and use its `Dial` method as DialFunc client option. Then you can check the messages that were exchanged with the `Sent` and `Received` methods.

ably/realtime_channel_spec_test.go
tcard: Ah, that's true. Thanks.

ably/realtime_channel_spec_test.go
sacOO7: Yeah, seems I missed it out. So, yes, this is a valid use case.

ably/realtime_channel_spec_test.go
sacOO7: Channel state would be initialized without calling attach first, if that's the case, spec ([RTL5a](https://docs.ably.io/client-lib-development-guide/features/#RTL5a)) already covers it and it's not about queuing the message but doing nothing. I believe we already have a test for the same.

sacOO7: Again coming back to the same point, maybe it should be removed from the spec same as `attach`

ably/transitioner_test.go
```diff
@@ -169,10 +177,35 @@ func (c ConnTransitioner) finishConnecting(err error) connTransitionFunc {
 		// Should be CONNECTED.
 		return connNextStates{
 			disconnected: c.disconnect,
+			closing:      c.close,
 		}, nil
 	}
 }
 
+func (c ConnTransitioner) failOnIntercept(msg <-chan *proto.ProtocolMessage, cancelIntercept func()) connTransitionFunc {
```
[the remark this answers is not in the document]
sacOO7: I believe, `finishConnecting` transitions to `disconnected` if error is passed as param

ably/realtime_channel_spec_test.go
tcard: That's true, unless I'm missing something there's no way for the connection to be INITIALIZED while the channel isn't INITIALIZED, so RTL5a covers this case already.

This part of the test can be removed then (CONNECTING and DISCONNECTED still apply, I think).

Maybe you want to raise this in ably/docs?

ably/transitioner_test.go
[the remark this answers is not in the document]
tcard: OK, not a blocking issue, can be refactored later.

ably/transitioner_test.go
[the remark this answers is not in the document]
tcard: True. Anyway, that's besides the point of the suggestion, really.

ably/realtime_channel_spec_test.go
```diff
@@ -926,15 +1014,35 @@ func TestRealtimeChannel_RTL5_Detach(t *testing.T) {
...
+		}
+
+		ablytest.Instantly.Recv(t, &change, stateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateAttached, change.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, change)
+		}
+
+		channel.Detach(cancelledContext)
+
+		// Cause a timeout.
```
tcard, line 1019: I'm probably missing something here. What's this timeout for?

ably/export_test.go
```diff
@@ -140,6 +140,7 @@ func (c *Connection) PendingItems() int {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 	return c.pending.Len()
+	return c.pending.Len()
```
tcard, line 143: Looks like this was an accidentally duplicated.

sacOO7: Fixed !

ably/realtime_channel_spec_test.go
sacOO7: Since we are calling detach to get into detaching state, the afterCall timeout needs to happen, otherwise test will be stuck there... anyways, it's just to get us to detaching state : )

ably/realtime_channel_spec_test.go
sacOO7: Yeah, I will raise the PR in the docs

ably/realtime_channel_spec_test.go
sacOO7: Fixed !

ably/realtime_channel_spec_test.go
sacOO7: Fixed !

lmars (approved):

tcard:
Besides a few suggestions, [this](https://github.com/ably/ably-go/pull/298#discussion_r617415998) is the only feedback that needs addressing.

ably/realtime_channel_spec_test.go
sacOO7: Okay, we don't need timeout there, we are already passing cancelled context

tcard (approved):

sacOO7:
Merging channel detach PR, since approved by both parties : )

Issue #184, opened 2020-08-16
Reduce race conditions by simplifying locking


tcard:
Right now, we have a lock for the connection and then a mutex per realtime channel. In the process of handling some event (protocol message, network condition, timer, or user call), we lock and unlock those any number of times. This means event handling isn't atomic: after unlocking and before locking again, another event may sneak in and maybe read or write partially consistent state.

We should carefully re-examine the overall thread-safety design. I think the simpler model that best fits how the library works (and arguably is implicitly assumed by the spec) is probably a JavaScript-like one: all events are dispatched to a single handler loop thread. (This is how iOS works too.)

PR #299, opened 2021-03-21
Channel Attach

### LINKED ISSUE #211: RTL4: Channel.attach
Write spec tests for [RTL4](https://docs.ably.io/client-lib-development-guide/features/#RTL4), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

sacOO7:
- Added tests for channel attach
- Fixes #211


ably/export_test.go
```diff
@@ -149,3 +149,11 @@ func (c *RESTChannels) Len() int {
 	defer c.mu.RUnlock()
 	return len(c.cache)
 }
+
+func (c *RealtimeChannel) SetState(state ChannelState, err error) error {
+	return c.setState(state, err)
+}
+
+func (ablyConn *Connection) SetState(state ConnectionState, err error, retryIn time.Duration) error {
```
tcard, line 157: The problem with setting just the state in tests is that it can easily lead to inconsistent overall states that can't happen in real conditions, so the tests can't be trusted. For instance, here you're resetting the state to INITIALIZED, which doesn't happen otherwise, and thus the several variables that realtime instances carry around could be in a state that's incompatible with such state (e. g. having a `connection.key` set), maybe causing unexpected behavior.

We should instead transition the state machine the via external events, as it's supposed to be the case. We have [ablytest.TransitionConn](https://github.com/ably/ably-go/blob/841b4e774670b72f648aa889c22b9f3e6b9cb261/ably/transitioner_test.go#L33) to make this explicit and deterministic.

sacOO7: This is really helpful 🙏. Seems I am getting some race conditions because of incompatible states across the client. I will go through the above recommended solution.

sacOO7:
Channel options are not implemented yet, we need to implement them as a part of RTL4k, RTL4l and RTL4m

lmars (approved):

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +157,1208 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		if channel.GetAttachResume() != false {
+			t.Fatalf("Channel attach resume should be false when channel state is FAILED")
+		}
+	})
+
+	t.Skip("RTL4j2")
+	t.Run("RTL4j2", func(t *testing.T) {
```
sacOO7, line 1148: @lmars this is the test, you can check

lmars: I suspect it's a problem with [these assertions](https://github.com/ably/ably-go/blob/fix/211-RTL4-channel-attach/ably/realtime_channel_spec_test.go#L1205-L1211) which are checking `messages1` for different lengths.

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +157,1258 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		if channel.GetAttachResume() != false {
+			t.Fatalf("Channel attach resume should be false when channel state is FAILED")
+		}
+	})
+
+	t.Skip("RTL4j2: Rewind flag should allow to receive historic messages")
+	t.Run("RTL4j2: Rewind flag should allow to receive historic messages", func(t *testing.T) {
```
sacOO7, line 1183: @tcard , this test is having a flaky behaviour, I went through the code multiple times, seems to be no issue with the logic. Still, not sure why we don't receive messages. It will be great if you can debug and check what might be the real issue here

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +157,1258 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+
+		if ably.UnwrapErrorCode(attachErr) != 50001 {
+			t.Fatalf("want code=50001; got %d", ably.UnwrapErrorCode(err))
+		}
+
+		ablytest.Instantly.NoRecv(t, nil, channelStateChanges, t.Fatalf) // Should not make any change to the channel state
+	})
+
+	// Transition to channelStateSuspended not implemented, so cannot implement test as of now
+	t.Skip("RTL4d : should return error on SUSPENDED while attaching channel")
```
tcard, line 596: This should be removed now.

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +157,1258 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		ablytest.Instantly.Recv(t, &channelStatechange, channelStateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateAttaching, channelStatechange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		ablytest.Instantly.Recv(t, &channelStatechange, channelStateChanges, t.Fatalf)
+		if expected, got := ably.ChannelStateAttached, channelStatechange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+		if channel.GetAttachResume() != true {
```
tcard, line 1155: RTL4j talks about the ATTACH_RESUME flag in the ATTACH message. We should check that, not the internal `attachResume` flag which isn't required (the spec just suggests it as an implementation).

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +157,1258 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+		if expected, got := ably.ChannelStateFailed, channelStatechange.Current; expected != got {
+			t.Fatalf("expected %v; got %v (event: %+v)", expected, got, channelStatechange)
+		}
+
+		if channel.GetAttachResume() != false {
+			t.Fatalf("Channel attach resume should be false when channel state is FAILED")
+		}
+	})
+
+	t.Skip("RTL4j2: Rewind flag should allow to receive historic messages")
```
tcard, line 1182: This should be removed.

ably/realtime_channel.go
```diff
@@ -63,9 +68,38 @@ func ChannelWithCipher(params CipherParams) ChannelOption {
...
+func ChannelWithParams(key string, value string) ChannelOption {
+	return func(o *channelOptions) {
+		if o.Params == nil {
+			o.Params = map[string]string{}
+		}
+		o.Params[key] = value
+	}
+}
+
+func ChannelWithRewind(numberOfMessages string) ChannelOption {
```
tcard, line 80: As the spec doesn't define this as part of the public API, this should be unexported or just removed.

ably/realtime_channel_spec_test.go
tcard: Try using `persisted:test` as channel name. There are no history guarantees for non-persisted channels. Maybe that's the issue.

I'd also add a check that the ATTACH message carries the ATTACH_RESUME flag.

Other than that, I really don't know. This looks to me as if it's following RTL4j2 correctly. (I'm not sure why we should be testing this anyway, given it's mainly server behavior.) Maybe ask the realtime team?

ably/realtime_channel.go
```diff
@@ -615,6 +703,34 @@ func (c *RealtimeChannel) isActive() bool {
...
+func (c *RealtimeChannel) Modes() []ChannelMode {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.modes
+}
+
+func (c *RealtimeChannel) Params() ChannelParams {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.params
```
tcard, line 731: This won't be read-only. This should return a copy of the original map.

ably/realtime_channel.go
```diff
@@ -615,6 +703,34 @@ func (c *RealtimeChannel) isActive() bool {
...
+func (c *RealtimeChannel) setModes(modes []ChannelMode) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	c.modes = modes
+}
+
+func (c *RealtimeChannel) Modes() []ChannelMode {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.modes
```
tcard, line 725: This won't be read-only. This should return a copy of the original slice.

ably/realtime_channel.go
sacOO7: Actually, it's being used in the dotnet client, and makes it easy for the developer to use it on the go

ably/realtime_channel_spec_test.go
sacOO7: Tried, but not working. Maybe, you can try running test on your machine locally.

sacOO7: It's a weird flaky behaviour at the same time, it's less flaky in debug mode ( like it passes most of the times in debug mode)

tcard (changes requested):
Change review status so that it doesn't appear as "approved".

ably/realtime_channel.go
```diff
@@ -151,9 +185,14 @@ type RealtimeChannel struct {
 	client         *Realtime
 	messageEmitter *eventEmitter
 	queue          *msgQueue
+	options        *channelOptions
+	params         ChannelParams
+	modes          []ChannelMode
+
+	attachResume bool
```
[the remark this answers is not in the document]
sacOO7: Yes, actually I am following it as per DotNet implementation, I don't think we will be able to find it in the spec, also it's not exposed via public API, so shouldn't matter IMO.

ably/realtime_channel.go
```diff
@@ -203,34 +252,68 @@ func (c *RealtimeChannel) mayAttach(checkActive bool) (result, error) {
...
+
 	switch c.client.Connection.State() {
 	// RTL4b
 	case ConnectionStateInitialized,
 		ConnectionStateClosed,
 		ConnectionStateClosing,
+		ConnectionStateSuspended,
 		ConnectionStateFailed:
-		return nil, newError(80000, errAttach)
+		return nil, newError(ErrConnectionFailed, errConnAttach(c.client.Connection.State()))
```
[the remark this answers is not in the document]
sacOO7: Haha, not sure, seems I have followed it from the originalCode, I vaguely remember the reason was, I wasn't able to 
find the error code for `errChannelAttachFailed`, we don't have error code for channel attach : (

ably/realtime_channel.go
```diff
@@ -203,34 +252,68 @@ func (c *RealtimeChannel) mayAttach(checkActive bool) (result, error) {
...
+		return res, nil
+	}
 
-	msg := &proto.ProtocolMessage{
-		Action:  proto.ActionAttach,
-		Channel: c.Name,
+	if c.state == ChannelStateDetaching { //// RTL4h
+		attachRes := c.internalEmitter.listenResult(ChannelStateDetached, ChannelStateFailed)
+		return resultFunc(func(ctx context.Context) error { // runs inside goroutine, need to check for locks again before accessing state
+			detachErr := attachRes.Wait(ctx)
```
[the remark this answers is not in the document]
sacOO7: Okay, I will check changes with respect to contexts

ably/realtime_conn.go
```diff
@@ -115,7 +115,7 @@ func (c *Connection) dial(proto string, u *url.URL) (conn proto.Conn, err error)
 	query := u.Query()
 	query.Add("heartbeats", "true")
 	u.RawQuery = query.Encode()
-	timeout := c.opts.httpOpenTimeout()
+	timeout := c.opts.realtimeRequestTimeout()
```
sacOO7, line 118: @tcard , I have reverted this change that belong to https://github.com/ably/ably-go/pull/312, because all of the channel
attach tests suddenly started breaking, because default `httpOpenTimeout` doesn't seem to be enough to get state from
`conneting` to `connected`. Most of the transitioner based tests just failed with error saying, went from `connecting` to `disconnected` cuz of short timeout

ably/proto/crypto_test.go
```diff
@@ -29,3 +30,66 @@ func TestGenerateRandomKey(t *testing.T) {
 		}
 	})
 }
+
+func TestRealtimeChannelModes_ToFlag(t *testing.T) {
```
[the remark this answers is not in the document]
sacOO7: Not sure, where to put this stuff, I might just refactor other stuff too that's part of channel options

ably/realtime_channel.go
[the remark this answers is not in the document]
sacOO7: I believe, it's because we are doing explicit check on connection states and since connections are in invalid states, we are specifying error as `ErrConnectionFailed`

ably/export_test.go
sacOO7: Resolved, we don't have `setState` anymore in the code

ably/realtime_channel_spec_test.go
sacOO7: I have marked test as skipped, you can take a look at it, if you get enough time

ably/ablytest/recorders.go
```diff
@@ -326,6 +326,30 @@ func (rec *MessageRecorder) Sent() []*proto.ProtocolMessage {
...
+				if counter == times {
+					return true
+				}
+			}
+		}
+		return false
+	}
+}
+
+func (rec *MessageRecorder) FindLatest(action proto.Action) *proto.ProtocolMessage {
```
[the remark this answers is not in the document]
sacOO7: renamed to `FindFirst`, resolved.

ably/realtime_channel.go
```diff
@@ -11,7 +11,12 @@ import (
 )
 
 var (
-	errAttach = errors.New("attempted to attach channel to inactive connection")
+	errConnAttach = func(connState ConnectionState) error {
+		return errors.New("cannot Attach channel because connection is in " + connState.String() + " state")
```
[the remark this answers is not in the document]
sacOO7: Resolved

ably/realtime_channel.go
```diff
@@ -203,34 +252,68 @@ func (c *RealtimeChannel) mayAttach(checkActive bool) (result, error) {
...
+	sendAttachMsg := func() (result, error) {
+		res := c.internalEmitter.listenResult(ChannelStateAttached, ChannelStateDetached, ChannelStateSuspended, ChannelStateFailed) //RTL4d
+		msg := &proto.ProtocolMessage{
+			Action:  proto.ActionAttach,
+			Channel: c.Name,
+		}
+		if c.channelOpts().Params != nil {
+			msg.Params = c.channelOpts().Params
+		}
+		if c.channelOpts().Modes != nil {
```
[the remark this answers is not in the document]
sacOO7: Resolved

ably/realtime_channel.go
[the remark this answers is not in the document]
sacOO7: Added small doc, resolved

ably/realtime_channel_spec_test.go
sacOO7: I have marked it as skipped, in the near future we surely have to implement it, so implementation is already there for the test, as soon as we have suspended state for channel, the test will work out of the box

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +157,1258 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+
+		var change ably.ChannelStateChange
+
+		var outMsg *proto.ProtocolMessage
+
+		errCh := asyncAttach(channel)
+
+		// Cause a timeout.
+		var afterCall ablytest.AfterCall
+		ablytest.Instantly.Recv(t, &afterCall, afterCalls, t.Fatalf)
```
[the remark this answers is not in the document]
sacOO7: resolved

ably/realtime_channel_spec_test.go
```diff
@@ -155,6 +157,1258 @@ func attachAndWait(t *testing.T, channel *ably.RealtimeChannel) {
...
+			t.Fatalf("expected %v; got %v", expected, got)
+		}
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		err = channel.Attach(ctx)
+
+		// Check that the attach message is sent
+		checkIfAttachSent := recorder.CheckIfSent(proto.ActionAttach, 1)
```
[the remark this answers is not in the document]
sacOO7: resolved

ably/realtime_channel_spec_test.go
sacOO7: It's skipped, as per spec we need to implement this test, so we can't remove it for the time being

ably/realtime_channel.go
sacOO7: resolved

ably/realtime_channel.go
sacOO7: resolved

ably/realtime_channel.go
```diff
@@ -63,9 +68,38 @@ func ChannelWithCipher(params CipherParams) ChannelOption {
...
+	for _, set := range os {
+		set(&to)
+	}
+	return &to
+}
+
 type CipherParams = proto.CipherParams
 type CipherAlgorith = proto.CipherAlgorithm
 type CipherMode = proto.CipherMode
+type ChannelParams = proto.ChannelParams
```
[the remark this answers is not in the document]
sacOO7: resolved

ably/proto/crypto_test.go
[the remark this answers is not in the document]
sacOO7: resolved, created new file `channel_options` and `channel_options_test` for channel options, lets not combine them with 
crypto, since we will have more channel options in the future

ably/realtime_channel.go
```diff
@@ -190,7 +230,16 @@ func (c *RealtimeChannel) Attach(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	return res.Wait(ctx)
+
+	attachTimeout := c.client.Connection.opts.realtimeRequestTimeout()
+	timeoutCtx, cancel := c.opts().contextWithTimeout(ctx, attachTimeout)
```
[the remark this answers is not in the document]
sacOO7: Need to check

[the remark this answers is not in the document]
sacOO7: We do transition to `suspended` only when error is `context.DeadlineExceeded`. Can you check this one again. 

Actually, you had suggested this custom change for `channel detach` in the past and I am reusing same implementation here. So, this shouldn't be a issue.

ably/realtime_channel.go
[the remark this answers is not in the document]
sacOO7: Yes, if you look carefully, it will unblock that `wait`, move to next line where it sends `attach` message and returns immediately. So, we are not blocking anywhere

ably/realtime_channel_spec_test.go
sacOO7: I think, I probably need to test another test for RTL4j

sacOO7: Added missing test, resolved

ably/realtime_channel_spec_test.go
tcard: I'd rather just leave it this way for now.

ably/realtime_channel.go
```diff
@@ -770,10 +770,10 @@ func (c *RealtimeChannel) Modes() []ChannelMode {
 	return modes
 }
 
-func (c *RealtimeChannel) Params() ChannelParams {
+func (c *RealtimeChannel) Params() channelParams {
```
tcard, line 773: This should return `map[string]string` to avoid returning an unexported type.

ably/realtime_channel.go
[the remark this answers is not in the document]
tcard: > We do transition to `suspended` only when error is `context.DeadlineExceeded`. Can you check this one again.

Right, so that's why we shouldn't propagate the caller-provided context to the inner code. If that context timeouts, a `context.DeadlineExceeded` will be emitted.

The fix is just this:

```go
timeoutCtx, cancel := c.opts().contextWithTimeout(context.Background(), attachTimeout)
```

> Actually, you had suggested this custom change for `channel detach` in the past and I am reusing same implementation here. So, this shouldn't be a issue.

I'm often wrong :)

ably/realtime_channel.go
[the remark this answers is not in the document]
tcard: I meant it to make it easier for us reviewers and soon to be maintainers.

ably/realtime_channel.go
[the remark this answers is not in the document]
tcard: But we _don't_ want the user to be able to unblock a wait in the inner parts of the library. We just want the user to be able to unblock its own call to `Detach`, but the underlying process started by such call such carry on in the background.

We made this explicit decision in https://github.com/ably/ably-go/issues/197#issuecomment-696979466.

ably/realtime_channel.go
tcard: We're trying to protect ourselves from incompatible spec changes down the line. If tomorrow the spec adds an explicit `rewind` ChannelOption which happens to be different to this one, then we'll have a problem. That's why we should stick to the public API as published in the spec.

ably/realtime_conn.go
tcard: Opened #325 to deal with this separately.

ably/realtime_channel.go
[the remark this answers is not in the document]
sacOO7: Then I am not sure, why we are passing `context` to `contextWithTImeout`, we don't need to, if it's not doing anything useful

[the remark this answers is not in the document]
tcard: Because the convention is that every function that produces a context takes a context, in case you want the new context to inherit from another one. If you don't, you pass `context.Background()`.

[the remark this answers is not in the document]
sacOO7: We can just create background context internally, rather than injecting it from outside, looking at current usecase

ably/realtime_channel.go
sacOO7: It doesn't give any errors, we are already using it in the tests. We can use it as long as we don't need to declare the type. In go, ofc we don't need to declare type at lot of places. Also, I want to be explicit about `channelParam` type, rather than 
returning random map

ably/realtime_channel.go
sacOO7: I will remove it for the time being, but I believe we should raise request for this public API in the docs, because `rewind` is just an extension of params and doesn't make sense for a user to pass hardcoded value for the same, rather it should be exposed via some public interface. We already have it in .net and works well for users ( based on the recent ticket raised by one of the customer)

sacOO7: Resolved

ably/realtime_channel.go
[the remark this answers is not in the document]
tcard: Yes, we could. But then that makes the function less flexible. Anyway, feel free to change it if you feel strongly about it.

ably/realtime_channel.go
tcard: > I believe we should raise request for this public API in the docs

Sure, that sounds like a good idea.

ably/realtime_channel.go
tcard: I know we _can_ do it, it's just pretty unfriendly to the user. The user doesn't even know what's a `channelParams` so they wouldn't know what to do with it. They also can't use it in declarations, so they can't pass it to a function of their own, or put it in a struct, or whatever else. In general it's a quirk and should be avoided.

> Also, I want to be explicit about `channelParam` type, rather than returning random map

What purpose does that serve to the consumers of the API?

ably/realtime_channel.go
```diff
@@ -237,14 +233,22 @@ func (c *RealtimeChannel) Attach(ctx context.Context) error {
...
+	go func() {
+		internalOpErr <- res.Wait(timeoutCtx)
+	}()
+	select {
+	case <-ctx.Done(): //Unblock channel wait, when external context is cancelled
+		return ctx.Err()
+	case err = <-internalOpErr:
+		if errors.Is(err, context.DeadlineExceeded) {
+			err = newError(ErrTimeoutError, errors.New("timed out before attaching channel"))
+			c.setState(ChannelStateSuspended, err, false)
```
tcard, line 248: This should happen even if the user cancels their context.

In general, we should make a strong distinction between the calling user goroutine and the goroutine that performs library operations, but that's for a further refactoring that includes #184.

sacOO7: I believe we agreed, user cancelling the context shouldn't affect internal library operations

ably/realtime_channel.go
sacOO7: Makes sense, explicit DS is easy to identify and easy to use. Calling params means it's returning params anyways.

ably/realtime_channel.go
sacOO7: I am not getting what are you trying to convey here?

sacOO7: resolved

ably/realtime_channel.go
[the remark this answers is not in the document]
sacOO7: resolved

tcard (approved):

PR #300, opened 2021-03-31
Remove Ping public Interface

sacOO7:
- Disabled Ping functionality
- References #263

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -470,9 +470,9 @@ func (c *Connection) Key() string {
 //
 // Ping returns non-nil error without any attempt of communication with Ably
 // if the connection state is ConnectionStateClosed or ConnectionStateFailed.
-func (c *Connection) Ping() (ping, pong time.Duration, err error) {
-	return 0, 0, errors.New("TODO")
-}
+//func (c *Connection) Ping() (ping, pong time.Duration, err error) {
+//	return 0, 0, errors.New("TODO")
+//}
 
 // ErrorReason gives last known error that caused connection transit to
 // ConnectionStateFailed state.


tcard (approved):

ably/realtime_conn.go
```diff
@@ -470,9 +470,9 @@ func (c *Connection) Key() string {
 //
 // Ping returns non-nil error without any attempt of communication with Ably
 // if the connection state is ConnectionStateClosed or ConnectionStateFailed.
-func (c *Connection) Ping() (ping, pong time.Duration, err error) {
-	return 0, 0, errors.New("TODO")
-}
+//func (c *Connection) Ping() (ping, pong time.Duration, err error) {
+//	return 0, 0, errors.New("TODO")
+//}
```
lmars, line 475: Is there some difficulty in actually implementing this? I think we should either implement it or remove the code rather than just commenting it out.

sachinsh76: Actually, we need to implement it, but it's not prioritized for this release. #263

sacOO7: Yes we are supposed to be implement it #263 , but it's of less priority as of now

lmars: Ok, so I don't see the benefit of commenting out this code rather than just leaving it as a TODO until it gets implemented?

sacOO7:
I am thinking, we might implement it right after we go beta.

ably/realtime_conn.go
sacOO7: Current TODO just returns error, we don't want it to be a part of public interface, at the same time, we don't want to 
lose current partial implementation/ doc-comments for the method. This is going to be implemented soon after beta release.

tcard:
@sacOO7 Is anything blocking the merge?

sacOO7:
Merging this PR, since there is no major change required for the time being.

PR #308, opened 2021-05-06
Fix channel iteration

### LINKED ISSUE #307: Fix channel iteration methods
They must comply with the spec:
```
{{ iterate() -> Iterator<ChannelType> // RSN2, RTS2 }}

```
Instead, we have these:

```
{{func (ch *RealtimeChannels) All() []*RealtimeChannelfunc (c *RESTChannels) Range(fn func(name string, channel *RESTChannel) bool) {}}

```

sacOO7:
- Added `Iterator()` to `RealtimeChannels` and `RESTChannels`
- Refactored code/tests, removed default channel name sort and range from restchannels
- Internal discussion [link](https://ably-real-time.slack.com/archives/C0114KK7S6R/p1620302233050500)
- Fixes #307


ably/export_test.go
```diff
@@ -147,5 +147,5 @@ type Result = result
 func (c *RESTChannels) Len() int {
```
tcard, line 147: Not necessarily as part of this PR, but can we unexport or remove this? It isn't part of the spec, and arguably it's not very useful.

ably/realtime_channel.go
```diff
@@ -85,26 +85,23 @@ func (ch *RealtimeChannels) Get(name string, options ...ChannelOption) *Realtime
...
 		chans = append(chans, c)
 	}
 	ch.mtx.Unlock()
-	chanSlice(chans).Sort()
 	return chans
 }
 
 // Exists returns true if the channel by the given name exists.
-func (c *RealtimeChannels) Exists(name string) bool {
+func (c *RealtimeChannels) Exists(name string) bool { // RSN2, RTS2
```
tcard, line 104: Same as above, this isn't part of the spec.

tcard (approved):
Looks good, but we could to make the whole Channels API spec-compliant by removing the extra methods.

ably/realtime_channel.go
tcard: > Same as above, this isn't part of the spec.

Well, this clearly isn't true. Sorry.

ably/export_test.go
sacOO7: Removed

tcard (approved):

PR #312, opened 2021-05-13
Allow setting connection and request timeouts separately

Jmgr:
See the discussion [there](https://github.com/ably/ably-boomer/pull/44).

This is a minimal set of changes required to set a different connection and request timeout for Realtime. This also adds `WithHTTPRequestTimeout` to set a timeout for REST requests without using `WithHTTPClient`.

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -115,7 +115,7 @@ func (c *Connection) dial(proto string, u *url.URL) (conn proto.Conn, err error)
 	query := u.Query()
 	query.Add("heartbeats", "true")
 	u.RawQuery = query.Encode()
-	timeout := c.opts.realtimeRequestTimeout()
+	timeout := c.opts.httpOpenTimeout()
 	if c.opts.Dial != nil {
 		conn, err = c.opts.Dial(proto, u, timeout)
 	} else {
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -91,12 +91,6 @@ func (c *Connection) MsgSerial() int64 {
 	return c.msgSerial
 }
 
-func WithRealtimeRequestTimeout(d time.Duration) ClientOption {
-	return func(os *clientOptions) {
-		os.RealtimeRequestTimeout = d
-	}
-}
-
 func WithTrace(trace *httptrace.ClientTrace) ClientOption {
 	return func(os *clientOptions) {
 		os.Trace = trace
--- a/ably/options.go
+++ b/ably/options.go
@@ -340,7 +340,7 @@ func (opts *clientOptions) httpclient() *http.Client {
 		return opts.HTTPClient
 	}
 	return &http.Client{
-		Timeout: defaultOptions.HTTPRequestTimeout,
+		Timeout: opts.HTTPRequestTimeout,
 	}
 }
 
@@ -609,6 +609,12 @@ func WithRESTHost(host string) ClientOption {
 	}
 }
 
+func WithHTTPRequestTimeout(timeout time.Duration) ClientOption {
+	return func(os *clientOptions) {
+		os.HTTPRequestTimeout = timeout
+	}
+}
+
 func WithRealtimeHost(host string) ClientOption {
 	return func(os *clientOptions) {
 		os.RealtimeHost = host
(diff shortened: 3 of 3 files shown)


ably/realtime_conn.go
```diff
@@ -115,7 +115,7 @@ func (c *Connection) dial(proto string, u *url.URL) (conn proto.Conn, err error)
 	query := u.Query()
 	query.Add("heartbeats", "true")
 	u.RawQuery = query.Encode()
-	timeout := c.opts.realtimeRequestTimeout()
+	timeout := c.opts.httpOpenTimeout()
```
tcard, line 118: [From the spec](https://docs.ably.io/client-lib-development-guide/features) (emphasis mine):

> (RTN14c) A new connection attempt will fail if not connected within the default **realtime request timeout**
> (DF1b) realtimeRequestTimeout – default 10s. When a realtime client library is **establishing a connection with Ably**, or sending a HEARTBEAT, CONNECT, ATTACH, DETACH or CLOSE ProtocolMessage to Ably, this is the amount of time that the client library will wait before considering that request as failed and triggering a suitable failure condition

But then:

> (TO3l3) httpOpenTimeout integer – default 4,000 (4s). Timeout for opening the connection, available in the client library if supported by the transport

This is a bit ambiguous but the `http` prefix there suggests to me that `httpOpenTimeout` is meant for connections over which HTTP requests are performed.

Given this change isn't critical for the main purpose of this pull request, maybe we should leave it out?

tcard: > This is a bit ambiguous but the `http` prefix there suggests to me that `httpOpenTimeout` is meant for connections over which HTTP requests are performed.

Doing a bit of blame archeology, this [indeed seems to be the case](https://github.com/ably/docs/pull/51#discussion_r41497849).

Jmgr: > This is a bit ambiguous but the `http` prefix there suggests to me that `httpOpenTimeout` is meant for connections over which HTTP requests are performed.

I agree, I had similar concerns there: https://github.com/ably/ably-boomer/pull/44#issuecomment-839981801

> Given this change isn't critical for the main purpose of this pull request, maybe we should leave it out?

The issue then is that we don't have a way to specify a different connection and request timeout for Realtime. We could use a custom `Dial` to override the timeout value, but the default websocket `Dial` is internal so we cannot use it in ably-boomer. That would mean either copy and pasting that default websocket dial or making it non-internal.

tcard: Actually what you discussed in https://github.com/ably/ably-boomer/pull/44 sounds to me like a more sensible interpretation of the spec than what I said above. `realtimeRequestTimeout` is for "establishing a connection with Ably", ie. an _Ably realtime_ connection. So it's the timeout for receiving a CONNECTED ProtocolMessage response.

`Dial` is about establishing the _transport_ WebSocket connection on which the Ably realtime connection will be based, so, while there isn't a specific timeout defined for this, `httpOpenTimeout` is close enough.

ably/realtime_conn_spec_test.go
```diff
@@ -1861,7 +1861,7 @@ func TestRealtimeConn_RTN14c(t *testing.T) {
 		ably.WithKey("xxx:xxx"),
 		ably.WithRealtimeHost(host),
 		ably.WithTLSPort(p),
-		ably.WithRealtimeRequestTimeout(reqTimeout),
+		ably.WithHTTPOpenTimeout(reqTimeout),
```
tcard, line 1864: This test for RTN14c should still be testing `realtimeRequestTimeout`.

ably/realtime_conn.go
tcard: I meant to say that I agree with this change now :) But I don't think the test for RTN14c should be changed. RTN14c is about `realtimeRequestTimeout`.

ably/realtime_conn_spec_test.go
Jmgr: I did that change because the test broke after replacing `RealtimeRequestTimeout` with `HTTPOpenTimeout` for the initial connection. I'm ok with reverting that though.

tcard (approved):

ably/realtime_conn_spec_test.go
tcard: Opened https://github.com/ably/ably-go/issues/315 to fix this.

PR #314, opened 2021-05-14
Fix a decode error when the server returns 204

Jmgr:
This fixes a `mime: no media type` error returned on calling `rest.Request` when the server returns 204 (no content). This occurred when trying to unsubscribe a push device from a channel.

I am not sure if there are other cases where the is no content to decode and how to catch them. Maybe we should not use the statusCode but the length of the data instead?

### DIFF
--- a/ably/paginated_result.go
+++ b/ably/paginated_result.go
@@ -267,11 +267,13 @@ func newPaginatedResult(ctx context.Context, opts *proto.ChannelOptions, req pag
 	}
 	p.path = builtPath
 	p.links = resp.Header["Link"]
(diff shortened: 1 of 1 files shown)


tcard:
paginated_result.go has changed quite a bit since the version you're basing your changes on. Could you rebase your changes on the current one? Maybe it's even accidentally fixed now.

Jmgr:
> paginated_result.go has changed quite a bit since the version you're basing your changes on. Could you rebase your changes on the current one? Maybe it's even accidentally fixed now.

This has indeed fixed the issue, thanks.

PR #316, opened 2021-05-17
REST + REALTIME Fallback

### LINKED ISSUE #225: RTN17: Host fallback
Write spec tests for [RTN17](https://docs.ably.io/client-lib-development-guide/features/#RTN17), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4327) by [Unito](https://www.unito.io)

sacOO7:
- Fixes #225


ably/ably_test.go
```diff
@@ -112,6 +113,85 @@ func assertNil(t *testing.T, object interface{}) {
 	}
 }
 
+// ObjectsAreEqual determines if two objects are considered equal.
+//
+// This function does no assertion of any kind.
+func objectsAreEqual(expected, actual interface{}) bool {
```
tcard, line 119: This function can be replaced by just `reflect.DeepEqual`. See https://play.golang.org/p/rQOdjzcqpaD

ably/ably_test.go
```diff
@@ -112,6 +113,85 @@ func assertNil(t *testing.T, object interface{}) {
...
+}
+
+// ElementsMatch asserts that the specified listA(array, slice...) is equal to specified
+// listB(array, slice...) ignoring the order of the elements. If there are duplicate elements,
+// the number of appearances of each of them in both lists should match.
+//
+// assert.ElementsMatch(t, [1, 3, 2, 3], [1, 3, 3, 2]))
+//
+// Returns whether the assertion was successful (true) or not (false).
+func assertElementsMatch(t *testing.T, listA, listB interface{}) (ok bool) {
```
tcard, line 144: It would be simpler to just do this:

```go
reflect.DeepEqual(sort.Strings(a), sort.Strings(b))
reflect.DeepEqual(sort.Ints(a), sort.Ints(b))
// etc.
```

ably/error.go
```diff
@@ -144,6 +145,31 @@ func errFromUnprocessableBody(resp *http.Response) error {
 	return &ErrorInfo{Code: ErrBadRequest, StatusCode: resp.StatusCode, err: err}
 }
 
+func isTimeoutOrDnsErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	switch err := err.(type) {
```
tcard, line 152: It's almost always wrong to unwrap errors this way. Errors are supposed to follow a "chain" convention. See https://pkg.go.dev/errors

This should also the special cases here unnecessary, because `*net.DNSError`, `*url.Error` and the timeout errors returned by the net/http package can all be unwrapped to a `net.Error`.

ably/export_test.go
```diff
@@ -77,12 +136,12 @@ func (a *Auth) SetServerTimeFunc(st func() (time.Time, error)) {
 	a.serverTimeHandler = st
 }
 
-func (c *REST) SetSuccessFallbackHost(duration time.Duration) {
-	c.successFallbackHost = &fallbackCache{duration: duration}
-}
+//func (c *REST) SetSuccessFallbackHost(duration time.Duration) {
```
tcard, line 139: Should this be removed?

ably/hosts.go
```diff
@@ -0,0 +1,233 @@
...
+func (restHosts *restHosts) getPrimaryHost() string {
+	return restHosts.opts.getPrimaryRestHost()
+}
+
+func (restHosts *restHosts) getFallbackHost() string {
+	restHosts.Lock()
+	defer restHosts.Unlock()
+	hosts, _ := restHosts.opts.getFallbackHosts()
+	shuffledFallbackHosts := ablyutil.Shuffle(hosts)
+	getNonVisitedHost := func() string {
```
tcard, line 38: Why is this a function given it's only called once, right away?

ably/hosts.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (restHosts *restHosts) getPrimaryHost() string {
+	return restHosts.opts.getPrimaryRestHost()
+}
+
+func (restHosts *restHosts) getFallbackHost() string {
+	restHosts.Lock()
+	defer restHosts.Unlock()
+	hosts, _ := restHosts.opts.getFallbackHosts()
+	shuffledFallbackHosts := ablyutil.Shuffle(hosts)
```
tcard, line 37: If we move this just above where it's used, we avoid shuffling unnecessarily if we return the primary host earlier.

ably/hosts.go
```diff
@@ -0,0 +1,233 @@
...
+
+	"github.com/ably/ably-go/ably/internal/ablyutil"
+)
+
+// RSC15
+type restHosts struct {
+	primaryFallbackHost string
+	opts                *clientOptions
+	cache               *hostCache
+	visitedHosts        []string
```
tcard, line 16: It's customary to use `map[T]struct{}` or `map[T]bool` as a `Set<T>` type. It also makes `ablyutil.Contains` unnecessary.

ably/internal/ablyutil/strings.go
```diff
@@ -0,0 +1,36 @@
+package ablyutil
+
+import (
+	"math/rand"
+	"strings"
+	"time"
+)
+
+func Shuffle(list []string) []string {
```
tcard, line 9: Simpler:

```go
r := rand.New(rand.NewSource(time.Now().UnixNano()))
r.Shuffle(len(list), func(i, j int) { list[i], list[j] = list[j], list[i] })
return list
```

ably/error.go
sacOO7: Referring to https://blog.golang.org/go1.13-errors. Seems new way of handling errors have been introduced, since we support minimum 1.13, I will use new ways of handling errors

ably/hosts.go
sacOO7: It's not about **re-usability**, rather **readability**. Also, that particular snippet exists in the context of returning right fallback host, so makes sense to keep it anonymous, while abstracting internal details

ably/hosts.go
sacOO7: Moved

ably/hosts.go
sacOO7: Updated with map based set type

ably/internal/ablyutil/strings.go
sacOO7: Done

ably/hosts.go
```diff
@@ -0,0 +1,241 @@
...
+	realtimeHosts.Lock()
+	defer realtimeHosts.Unlock()
+	host := realtimeHosts.getPrimaryHost() // primary host is always preferred host/ fallback host in realtime
+	realtimeHosts.visitedHosts.Add(host)
+	return host
+}
+
+// hostCache this caches a successful fallback host for 10 minutes.
+// Only used by REST client while making requests RSC15f
+type hostCache struct {
```
tcard, line 176: This seems way too complicated, dealing unnecessary with concurrency and timeouts. How about something like:

```go
type hostCache struct {
    duration time.Duration

    sync.RWMutex
    deadline time.Time
    host string
}

func (c *hostCache) put(host string) {
    c.Lock()
    defer c.Unlock()
    c.host = host
    c.deadline = time.Now().Add(c.duration)
}

func (c *hostCache) get() string {
    c.RLock()
    defer c.RUnlock()
    if c.host == "" || time.Until(c.deadline) <= 0 {
        return ""
    }
    return c.host
}
```

ably/options.go
```diff
@@ -22,10 +23,12 @@ const (
 	protocolMsgPack = "application/x-msgpack"
 
 	// restHost is the primary ably host .
-	restHost     = "rest.ably.io"
-	realtimeHost = "realtime.ably.io"
-	Port         = 80
-	TLSPort      = 443
+	restHost         = "rest.ably.io"
+	realtimeHost     = "realtime.ably.io"
+	Port             = 80
```
tcard, line 28: I just noticed we made those unnecessarily exported consts part of the stable 1.2 release. Oh well.

ably/options.go
```diff
@@ -443,6 +440,22 @@ func (opts *clientOptions) idempotentRESTPublishing() bool {
 	return opts.IdempotentRESTPublishing
 }
 
+func isActiveInternetConnection() bool {
+	res, err := http.Get(internetCheckUrl)
```
tcard, line 444: We should always use the HTTP client from the ClientOptions.

ably/realtime_client_test.go
```diff
@@ -47,6 +47,83 @@ func TestRealtime_RealtimeHost(t *testing.T) {
...
+		t.Run("does not apply when environment is overriden and fallback not specified", func(t *testing.T) {
+
+		})
+
+		t.Run("apply when environment is overriden and fallback specified, the fallback is used", func(t *testing.T) {
+
+		})
+	})
+
+	t.Run("RTN17c: Verifies internet connection is active in case of error necessitating use of an alternative host", func(t *testing.T) {
```
tcard, line 114: Could you mark empty tests with `t.Skip`? This way, if we forget to finish here, it will show up in test runs.

(I realize this is realtime and thus probably part of #359.)

ably/options.go
tcard: (Resolving because this is just an observation.)

ably/options.go
tcard: I see this is fixed in the next PR: https://github.com/ably/ably-go/pull/359/files#diff-721ab670921034c0c26f7999ad630004f73001cb85be84c768e7bf93c6abf14fR444

ably/hosts.go
sacOO7: Yeah, this looks way more easy, actually earlier one was implemented by someone else, so I didn't try to change it

ably/realtime_client_test.go
sacOO7: Yes, it's a part of realtime fallback and I have updated it

ably/hosts.go
sacOO7: Simplified the implementation !

ably/ably_test.go
```diff
@@ -132,6 +132,27 @@ func assertNotDeepEquals(t *testing.T, expected interface{}, actual interface{})
 	}
 }
 
+func assertSubset(t *testing.T, set []string, subset []string) {
+	t.Helper()
+	for _, item := range subset[1:] {
```
tcard, line 137: What's special about `subset[0]`? It seems to me that the `[1:]` should be in the caller, not here.

tcard (approved):

ably/ably_test.go
sacOO7: Thanks, fixed it !

ably/hosts_test.go
```diff
@@ -0,0 +1,284 @@
...
+		prefHost := restHosts.GetPreferredHost()
+		assertEquals(t, "rest.ably.io", prefHost)
+
+		actualHosts = append(actualHosts, prefHost)
+
+		// Get all fallback hosts in random order
+		actualHosts = append(actualHosts, restHosts.GetAllRemainingFallbackHosts()...)
+
+		assertNotDeepEquals(t, expectedHosts, actualHosts)
+		assertDeepEquals(t, ablyutil.Sort(expectedHosts), ablyutil.Sort(actualHosts))
```
lmars, line 37: Given this test is called "should get fallback hosts in random order", should it be asserting that the hosts are actually random and not in alphabetical order?

ably/internal/ablyutil/strings_test.go
```diff
@@ -0,0 +1,132 @@
...
+		areEqual = reflect.DeepEqual(strList, shuffledList)
+		if !areEqual {
+			t.Errorf("%v should be equal to %v", strList, shuffledList)
+		}
+
+		strList = []string{"a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l"}
+		shuffledList = ablyutil.Shuffle(strList)
+		areEqual = reflect.DeepEqual(strList, shuffledList)
+		if areEqual {
+			t.Errorf("%v is equal to %v", strList, shuffledList)
```
lmars, line 32: How can we assert that these are not equal?

ably/options.go
```diff
@@ -443,6 +440,22 @@ func (opts *clientOptions) idempotentRESTPublishing() bool {
...
+	res, err := opts.httpclient().Get(internetCheckUrl)
+	if err != nil {
+		return false
+	}
+	data, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		return false
+	}
+	if string(data) == internetCheckOk {
```
lmars, line 453: This could just be `return string(data) == internetCheckOk`

ably/realtime_conn.go
```diff
@@ -360,21 +362,37 @@ func (c *Connection) connectWith(arg connArgs) (result, error) {
...
-	u.RawQuery = query.Encode()
-	proto := c.opts.protocol()
-
-	if c.State() == ConnectionStateClosed { // RTN12d - if connection is closed by client, don't try to reconnect
-		return nopResult, nil
-	}
+	var conn conn
+	host := c.hosts.getPreferredHost()
+	for {
+		u, err := url.Parse(c.opts.realtimeURL(host))
```
lmars, line 368: It's incorrect to pass a fallback host into `opts.realtimeURL`, since it will reference `opts.Port` and `opts.TLSPort` which should only be used when connecting to the primary endpoint, not the fallbacks.

ably/rest_client.go
```diff
@@ -500,155 +497,49 @@ func (c *REST) do(ctx context.Context, r *request) (*http.Response, error) {
...
-	}
-}
-
-func (f *fallbackCache) put(host string) {
-	if f.get() != host {
-		if f.isRunning() {
-			f.stop()
+		if host != c.hosts.getPrimaryHost() { // set hostheader for fallback host
+			req.Host = req.URL.Host // RSC15j set host header, https://github.com/golang/go/issues/7682, since req.Host overrides req.URL.Host, use the same value
+			req.Header.Set(hostHeader, host)
```
lmars, line 518: Won't `newHTTPRequest` do this anyway?

ably/hosts.go
```diff
@@ -0,0 +1,192 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/ably/ably-go/ably/internal/ablyutil"
+)
+
+// RSC15
+type restHosts struct {
+	primaryFallbackHost string
```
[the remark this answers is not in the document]
sacOO7: `RTN17` does talk about using realtime fallback host as primary host for all rest requests

ably/hosts.go
```diff
@@ -0,0 +1,192 @@
...
+	restHosts.Lock()
+	defer restHosts.Unlock()
+
+	getNonVisitedHost := func() string {
+		visitedHosts := restHosts.visitedHosts
+		cachedHost := restHosts.cache.get()
+		if !ablyutil.Empty(cachedHost) && !visitedHosts.Has(cachedHost) {
+			return cachedHost
+		}
+		primaryHost := restHosts.getPrimaryHost()
```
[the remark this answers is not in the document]
sacOO7: It can return `primary` host, if `fallback` is cached and it fails next time. So, for all subsequent retries, `primary` host is being treated as `fallback` host.

ably/hosts.go
```diff
@@ -0,0 +1,192 @@
...
+	return len(hosts) + 1 - len(restHosts.visitedHosts)
+}
+
+func (restHosts *restHosts) setPrimaryFallbackHost(host string) {
+	restHosts.Lock()
+	defer restHosts.Unlock()
+	restHosts.primaryFallbackHost = host
+}
+
+func (restHosts *restHosts) getPreferredHost() string {
```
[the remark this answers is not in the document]
sacOO7: Most of those terms, I have reused from `java SDK`, so we will have a good consistency in terms of understanding. I will attach the reference code 
https://github.com/ably/ably-java/blob/4a785a4cd96149bf9b398c4a07ef0c87bc554529/lib/src/main/java/io/ably/lib/transport/Hosts.java#L108-L111

ably/hosts.go
```diff
@@ -0,0 +1,192 @@
...
+			host = restHosts.getPrimaryHost()
+		} else {
+			host = restHosts.primaryFallbackHost
+		}
+	}
+	restHosts.visitedHosts.Add(host)
+	return host
+}
+
+func (restHosts *restHosts) cacheHost(host string) {
```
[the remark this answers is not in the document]
sacOO7: It can cache any host that succeeds, it can be primary or a fallback host

ably/hosts_test.go
sacOO7: `assertNotDeepEquals` makes sure they are random

ably/realtime_conn.go
sacOO7: It will reference only standard `PORT/TLSPORT`. We already have checks in place to make sure not to return fallbacks for custom port/TLSport set
https://github.com/ably/ably-go/blob/1405d1fc946be7f8649471f1f7059b4d92a5a30f/ably/options.go#L346-L369

ably/rest_client.go
sacOO7: No it will only set `req.URL.Host` and not `req.Host`

ably/export_test.go
```diff
@@ -6,22 +6,104 @@ import (
...
 func NewClientOptions(os ...ClientOption) *clientOptions {
 	return applyOptionsWithDefaults(os...)
 }
 
+func NewRestHosts(opts *clientOptions) *restHosts {
+	return newRestHosts(opts)
+}
+
+func (restHosts *restHosts) GetPrimaryHost() string {
+	return restHosts.getPrimaryHost()
```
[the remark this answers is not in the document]
sacOO7: Need to check on this one

ably/hosts.go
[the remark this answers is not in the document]
lmars: [RTN17e](https://docs.ably.io/client-lib-development-guide/features/#RTN17e) mentions using the same host the realtime client is connected to for REST requests if that realtime client is connected to a fallback host:

> If the realtime client is connected to a fallback host endpoint, then for the duration that the transport is connected to that host, all HTTP requests, such as history or token requests, should be first attempted to the same datacenter the realtime connection is established with i.e. the same fallback host must be used as the default HTTP request host. If however the HTTP request against that fallback host fails, then the normal fallback host behavior should be followed attempting the request against another fallback host as described in RSC15

So, I would understand perhaps calling this "active realtime fallback host", but it really is unrelated to the "primary host", which continues to be `rest.ably.io` (or whatever was set in `restHost`).

ably/hosts.go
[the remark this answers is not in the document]
lmars: That's really not very clear. We have a primary endpoint, and a set of fallback endpoints, let's please not refer to the primary endpoint as a fallback endpoint, since it isn't a fallback endpoint.

ably/hosts.go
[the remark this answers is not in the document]
lmars: Ok, please add a comment, since this isn't very clear.

ably/hosts.go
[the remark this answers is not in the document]
lmars: Ok, so given we're using the term "preferred" in other places, perhaps call this `setPreferredHost`?

ably/hosts_test.go
lmars: I think a better approach is to use an explicit seed for the random generator that leads to a well known order, and then assert that order here, since not controlling randomness in tests typically leads to flaky tests.

ably/realtime_conn.go
lmars: Ok, again, this really isn't very clear, please add a comment to this effect.

ably/rest_client.go
lmars: From the docs on [`http.Request.Host`](https://pkg.go.dev/net/http#Request):

```
// For client requests, Host optionally overrides the Host
// header to send. If empty, the Request.Write method uses
// the value of URL.Host. Host may contain an international
// domain name.
Host string
```

So it claims to do exactly what we're doing here (i.e. set `Host` to the value of `URL.Host`), are you sure this is required?

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: Ok, calling something like that would probably make sense, it's more explicit. I would make the change for the same

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: What name would you recommend? In the context of retry loop, `nextFallback` makes perfect sense, since we will be trying out fallbacks, doesn't matter if it's a primary host or a fallback host. It's a part of fallback loop. This particular case happens because of cache, this will not happen in case of realtimeHosts, since preferred host is always primary host

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: Preferred host is cached host (if found) or primary host

ably/realtime_conn.go
sacOO7: There will never be a case of having fallback hosts with custom ports. If there are custom ports, as per spec, all fallbacks will be empty.

ably/rest_client.go
sacOO7: It's required to set host header explicitly, I think we discussed this before. Although, as per documentation things are not clear about it.

ably/internal/ablyutil/strings_test.go
sacOO7: I didn't get

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: Also, I am sure any other name would just be more confusing

ably/hosts.go
```diff
@@ -0,0 +1,192 @@
+package ably
+
+import (
+	"sync"
+	"time"
+
+	"github.com/ably/ably-go/ably/internal/ablyutil"
+)
+
+// RSC15
```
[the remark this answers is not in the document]
sacOO7: Yes, I will add the comment

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: I will add it as a comment on top of the method

ably/hosts_test.go
sacOO7: I will check, currently I never faced failing tests related to shuffled hosts

ably/options.go
sacOO7: I will update

ably/export_test.go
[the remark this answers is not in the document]
sacOO7: Done

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: If you read `https://docs.ably.io/client-lib-development-guide/features/#RSC15f` carefully, it states to use primaryFallbackHost if `cachedHost` expires. Implementation is as per spec.

ably/rest_client.go
sacOO7: I have updated it, but I strongly remember not able to set header when Host was not set : (

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: Done

ably/hosts_test.go
sacOO7: I have removed unnecessary assertions to check shuffle, since we already have assertions in place for shuffle method. Also, I do not want deterministic order for shuffle, since it destroys purpose of having shuffle. I do not see any flakiness nature of the tests as of now, for ~40-50 runs.  I will change the tests, if I found them flaky in the future (highly unlikely).

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: Done

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: Done

ably/hosts.go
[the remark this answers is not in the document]
sacOO7: Done

lmars:
@sacOO7 I'll be honest, I think this PR adds quite a lot of complexity (a cached host, a preferred host, an active realtime fallback host, a visited hosts set that needs to be reset, a "next fallback host" iterator that doesn't just return fallbacks, lots of mutex locking), and I'm not sure that extra complexity is really warranted.

Reading the spec, the logic is something like:

For realtime, try the primary host, then each fallback in random order, tracking which fallback is successful as the active fallback for 10 minutes.

For REST, try the active fallback if set, then the primary host, then each fallback in random order.

Could that logic not simply just be put in the for loop within the Realtime connection / REST request methods? For example:

```go
// realtime connect
func connect() (conn Connection, err error) {
	// try the primary host
	conn, err = c.dial(c.opts.RealtimeHost)
	if err == nil || !retryableConnectionErr(err) {
		return
	}

	// try each fallback in random order
	fallbackHosts := ablyutil.Shuffle(c.opts.fallbackHosts())
	for _, host := range fallbackHosts {
		conn, err = c.dial(host)
		if err == nil {
			c.setActiveFallback(host)
			return
		} else if !retryableConnectionErr(err) {
			return
		}
	}
	return
}

// REST request
func request(req Request) (res Response, err error) {
	// try the active realtime fallback if set
	if activeFallback := c.realtime.activeFallback(); activeFallback != "" {
		res, err = c.do(req, activeFallback)
		if err == nil || !retryableRESTErr(err) {
			return
		}
	}

	// try the primary host
	res, err = c.do(req, c.opts.RESTHost)
	if err == nil || !retryableRESTErr(err) {
		return
	}

	// try each fallback in random order
	fallbackHosts := ablyutil.Shuffle(c.opts.fallbackHosts())
	for _, host := range fallbackHosts {
		res, err = c.do(req, c.opts.RESTHost)
		if err == nil || !retryableRESTErr(err) {
			return
		}
	}
	return
}
```

This is obviously pseudo-code, but hopefully illustrates my point.

sacOO7:
> @sacOO7 I'll be honest, I think this PR adds quite a lot of complexity (a cached host, a preferred host, an active realtime fallback host, a visited hosts set that needs to be reset, a "next fallback host" iterator that doesn't just return fallbacks, lots of mutex locking), and I'm not sure that extra complexity is really warranted.
> 
> Reading the spec, the logic is something like:
> 
> For realtime, try the primary host, then each fallback in random order, tracking which fallback is successful as the active fallback for 10 minutes.
> 
> For REST, try the active fallback if set, then the primary host, then each fallback in random order.
> 
> Could that logic not simply just be put in the for loop within the Realtime connection / REST request methods? For example:
> 
> ```go
> // realtime connect
> func connect() (conn Connection, err error) {
> 	// try the primary host
> 	conn, err = c.dial(c.opts.RealtimeHost)
> 	if err == nil || !retryableConnectionErr(err) {
> 		return
> 	}
> 
> 	// try each fallback in random order
> 	fallbackHosts := ablyutil.Shuffle(c.opts.fallbackHosts())
> 	for _, host := range fallbackHosts {
> 		conn, err = c.dial(host)
> 		if err == nil {
> 			c.setActiveFallback(host)
> 			return
> 		} else if !retryableConnectionErr(err) {
> 			return
> 		}
> 	}
> 	return
> }
> 
> // REST request
> func request(req Request) (res Response, err error) {
> 	// try the active realtime fallback if set
> 	if activeFallback := c.realtime.activeFallback(); activeFallback != "" {
> 		res, err = c.do(req, activeFallback)
> 		if err == nil || !retryableRESTErr(err) {
> 			return
> 		}
> 	}
> 
> 	// try the primary host
> 	res, err = c.do(req, c.opts.RESTHost)
> 	if err == nil || !retryableRESTErr(err) {
> 		return
> 	}
> 
> 	// try each fallback in random order
> 	fallbackHosts := ablyutil.Shuffle(c.opts.fallbackHosts())
> 	for _, host := range fallbackHosts {
> 		res, err = c.do(req, c.opts.RESTHost)
> 		if err == nil || !retryableRESTErr(err) {
> 			return
> 		}
> 	}
> 	return
> }
> ```
> 
> This is obviously pseudo-code, but hopefully illustrates my point.

I do not agree with your point, also your understanding about the spec is wrong. Your suggestion about putting everything together would instead create more mess, that's how it was implemented before and it was hard to understand when multiple spec items were mixing up together. As far as locking goes, it's not a major issue. Keeping fallback spec related code separate makes it easy to read and unit test separately.

Updated spec logic goes like this -> 
1. For realtime client, always get primary host as pref. host and then use fallback hosts in iterative order. If one of the fallback host succeds, use it as preferred host for subsequent **REST** requests.
2. For rest client, get preferred host in the order 1. Cached host 2. Active realtime host 3. primary host, , for getting fallback host, check if active realtime host is present or primary host not visited yet, if neither true, then return fallback host.

If we need to put all of this logic in single file according to spec, it will create a much more complex mess if you ask me. In current case, we rather abstract out the logic, and return basic interfaces like getting next fallback host, get pref. host and check for loop termination condition.

sacOO7:
@lmars , also most of the code as a part of this PR comprises of tests, around 80 %, we are covering every other edge case. So, I don't think current implementation is complex, rather simplified. It is the way it is, according to spec.

Rosalita:
Closing this PR as it has been outstanding for too long and become stale. A fresh PR will be raised to pick this work up in the future.

PR #317, opened 2021-05-17
Rewrite RTN14c to test full connection establishment

### LINKED ISSUE #315: Fix RTN14c test to include full connection establishment
We were wrongly using `realtimeRequestTimeout` as a timeout for establishing a transport; it should be the timeout for establishing a realtime connection, which includes receiving the CONNECTED protocol message. For the transport, we now use `httpOpenTimeout` (see https://github.com/ably/ably-go/pull/312).

RTN14c needs to be adapted to this change.

tcard:
Also, mock time to make it deterministic.

Fixes #315.

### DIFF
--- a/ably/realtime_conn_spec_test.go
+++ b/ably/realtime_conn_spec_test.go
@@ -1,18 +1,12 @@
 package ably_test
 
 import (
-	"bytes"
 	"context"
-	"crypto/tls"
 	"errors"
 	"fmt"
 	"io"
-	"log"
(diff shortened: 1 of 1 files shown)


sacOO7 (approved):
Wow, test is so much simplified. Seems, we already had tools to simplify it, but somehow we missed out on this one

tcard:
Failing tests are unrelated and being addressed separately.

PR #318, opened 2021-05-17
Avoid current stats period interference with fixtures in tests

### LINKED ISSUE #313: Flaky test in 1.2 branch: TestStatsPagination_RSC6a_RSCb3

tcard:
Fixes #313.

### DIFF
--- a/ably/ablytest/pagination.go
+++ b/ably/ablytest/pagination.go
@@ -10,7 +10,7 @@ import (
 // request into the slice pointed to by dst, which must be a pointer to a slice
 // of the same type as the paginated response's items.
 func AllPages(dst, paginatedRequest interface{}) error {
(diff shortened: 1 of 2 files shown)


sacOO7 (approved):
Not sure what exact issue was, but I checkout this branch, and tests seem to be working for me ( for multiple runs)

tcard:
@sacOO7 Sorry, let me explain. We're posting stats fixtures, for a given period in the past. This allows us to know what to expect when we fetch the stats again.

But the very act of posting those fixtures causes activity that is captured in a new stats period. So, when we retrieve stats, sometimes we get that extra activity, sometimes we don't, making it impossible to know what to expect.

So we need to add a filter to skip that extra stats, to get stable results.

PR #321, opened 2021-05-20
Simplify and uniformize logging

tcard:
Cleans up logging-related code as a prerequisite for #160. Commit messages contain rationale.

### DIFF
--- a/ably/rest_client.go
+++ b/ably/rest_client.go
@@ -13,7 +13,6 @@ import (
 	"mime"
 	"net/http"
 	"net/http/httptrace"
-	"net/http/httputil"
 	"net/url"
 	"reflect"
 	"strconv"
@@ -118,6 +117,7 @@ type REST struct {
 	Channels            *RESTChannels
 	opts                *clientOptions
 	successFallbackHost *fallbackCache
+	log                 logger
 }
 
 // NewREST constructs a new REST.
@@ -125,6 +125,7 @@ func NewREST(options ...ClientOption) (*REST, error) {
 	c := &REST{
 		opts: applyOptionsWithDefaults(options...),
 	}
+	c.log = logger{l: c.opts.LogHandler}
 	auth, err := newAuth(c)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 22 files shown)


ably/rest_client.go
```diff
@@ -615,19 +597,19 @@ func (c *REST) doWithHandle(ctx context.Context, r *request, handle func(*http.R
 				if c.opts.FallbackHosts != nil {
 					fallback = c.opts.FallbackHosts
 				}
-				log.Info("RestClient: trying to fallback with hosts=%v", fallback)
+				c.log.Infof("RestClient: trying to fallback with hosts=%v", fallback)
```
sacOO7, line 600: Should be `Info` instead of `Infof`

ably/rest_client.go
```diff
@@ -118,13 +117,15 @@ type REST struct {
...
 	successFallbackHost *fallbackCache
+	log                 logger
 }
 
 // NewREST constructs a new REST.
 func NewREST(options ...ClientOption) (*REST, error) {
 	c := &REST{
 		opts: applyOptionsWithDefaults(options...),
 	}
+	c.log = logger{l: c.opts.LogHandler}
```
sacOO7, line 128: Just curious, are we passing any logger context here? as in this logger belongs to `RestClient`, so everytime it logs stuff, 
it will be prefixed with `RestClient`

sacOO7: And it should be applicable at every other place, where we are initializing domain specific logger

tcard: That's what #323 is about, this PR just cleans up what's already there.

ably/rest_client.go
tcard: Yes, leaving just the `f` versions is probably a good idea.

tcard: Just to reiterate, this PR just cleans up what's already there. We can do this separately.

sacOO7 (approved):

ably/rest_client.go
tcard: #323 is closed now but the task will be done in a further PR.

PR #323, opened 2021-05-20
Draft of scoped logging

tcard:
#160 says:

> every operation should have at least a "begin" and "end" log message, with parameters and outcome

This is a proposal on how to do that, by introducing log scopes.

A pet peeve of mine is when logs from concurrent threads interleave, making it hard to discern what causes what. Log scopes fix that by prepending a mini stack trace to each log message, with IDs. Scopes have a beginning and an end, and can nest.

A scope is created like this:

```go
ctx, log, end := a.log().Begin(ctx, LogVerbose, "requestToken", "params", params)
defer end(&err)
```

This way, the rest of the function will be inside the log scope, and any log message from it, or from functions called from it, will be included in the scope.

To get a taste of how this looks like, I've run `TestRealtimeConn_RTN14b` (reauthorization) with different log levels. The default one (WARNING):

```
2021/05/25 17:26:24 [WARN] requestToken:end (err="[ErrorInfo :bad token request code=40170 error from client token callback statusCode=401] See https://help.ably.io/error/40170") @ requestToken:000c←authorize:000a←reauth:0007
2021/05/25 17:26:24 [WARN] authorize:end (err="[ErrorInfo :bad token request code=40170 error from client token callback statusCode=401] See https://help.ably.io/error/40170") @ authorize:000a←reauth:0007
```

From this, we now some non-critical error happened and can see how: reauthorization called authorize called requestToken, and that failed.

INFO:

```
2021/05/25 17:25:55 [INFO] conn:begin @ conn:0001
2021/05/25 17:25:55 [INFO] Connection state change: INITIALIZED → CONNECTING (reason=<nil>) @ conn:0001
2021/05/25 17:25:55 [INFO] Auth: sending  token request @ authorize:0005
2021/05/25 17:25:55 [INFO] Auth: sending  token request @ authorize:000c←reauth:000b
2021/05/25 17:25:55 [WARN] requestToken:end (err="[ErrorInfo :bad token request code=40170 error from client token callback statusCode=401] See https://help.ably.io/error/40170") @ requestToken:000d←authorize:000c←reauth:000b
2021/05/25 17:25:55 [WARN] authorize:end (err="[ErrorInfo :bad token request code=40170 error from client token callback statusCode=401] See https://help.ably.io/error/40170") @ authorize:000c←reauth:000b
2021/05/25 17:25:55 [INFO] Connection state change: CONNECTING → DISCONNECTED (reason=[ErrorInfo :bad token request code=40170 error from client token callback statusCode=401] See https://help.ably.io/error/40170) @ conn:0001
```

This includes connection state changes, which makes it clear at what point reauthorization happened.

VERBOSE:

```
2021/05/25 17:25:02 [INFO] conn:begin @ conn:0001
2021/05/25 17:25:02 [INFO] Connection state change: INITIALIZED → CONNECTING (reason=<nil>) @ conn:0001
2021/05/25 17:25:02 [VBOS] authorize:begin (params="<nil>" force="false") @ authorize:0009
2021/05/25 17:25:02 [INFO] Auth: sending  token request @ authorize:0009
2021/05/25 17:25:02 [VBOS] requestToken:begin (params="<nil>") @ requestToken:000a←authorize:0009
2021/05/25 17:25:02 [VBOS] Auth: found AuthCallback in []AuthOption @ requestToken:000a←authorize:0009
2021/05/25 17:25:02 [VBOS] requestToken:end @ requestToken:000a←authorize:0009
2021/05/25 17:25:02 [VBOS] authorize:end @ authorize:0009
2021/05/25 17:25:02 [VBOS] Realtime Connection: received (action="error", error=&proto.ErrorInfo{StatusCode: 401, Code: 40140, HRef:"", Message:"", Server:""}) @ conn:0001
2021/05/25 17:25:02 [VBOS] Protocol message received: error @ conn:0001
2021/05/25 17:25:02 [VBOS] reauth:begin @ reauth:000b
2021/05/25 17:25:02 [VBOS] authorize:begin (params="<nil>" force="true") @ authorize:000c←reauth:000b
2021/05/25 17:25:02 [INFO] Auth: sending  token request @ authorize:000c←reauth:000b
2021/05/25 17:25:02 [VBOS] requestToken:begin (params="<nil>") @ requestToken:000d←authorize:000c←reauth:000b
2021/05/25 17:25:02 [VBOS] Auth: found AuthCallback in []AuthOption @ requestToken:000d←authorize:000c←reauth:000b
2021/05/25 17:25:02 [WARN] requestToken:end (err="[ErrorInfo :bad token request code=40170 error from client token callback statusCode=401] See https://help.ably.io/error/40170") @ requestToken:000d←authorize:000c←reauth:000b
2021/05/25 17:25:02 [WARN] authorize:end (err="[ErrorInfo :bad token request code=40170 error from client token callback statusCode=401] See https://help.ably.io/error/40170") @ authorize:000c←reauth:000b
2021/05/25 17:25:02 [VBOS] reauth:end @ reauth:000b
2021/05/25 17:25:02 [INFO] Connection state change: CONNECTING → DISCONNECTED (reason=[ErrorInfo :bad token request code=40170 error from client token callback statusCode=401] See https://help.ably.io/error/40170) @ conn:0001
```

This includes the beginning and end of each scope, plus fine-grained branching inside each scope.

Adding @QuintinWillison as a reviewer in case he wants to provide feedback on the output (or, well, anything else!).

### DIFF
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -178,12 +178,15 @@ func (a *Auth) RequestToken(ctx context.Context, params *TokenParams, opts ...Au
 }
 
 func (a *Auth) requestToken(ctx context.Context, params *TokenParams, opts *authOptions) (tok *TokenDetails, tokReqClientID string, err error) {
+	ctx, log, end := a.log().Begin(ctx, LogVerbose, "requestToken", "params", params)
+	defer end(&err)
+
 	switch {
 	case opts != nil && opts.Token != "":
-		a.log().Verbose("Auth: found token in []AuthOption")
+		log.Verbose("Auth: found token in []AuthOption")
 		return newTokenDetails(opts.Token), "", nil
 	case opts != nil && opts.TokenDetails != nil:
-		a.log().Verbose("Auth: found TokenDetails in []AuthOption")
+		log.Verbose("Auth: found TokenDetails in []AuthOption")
 		return opts.TokenDetails, "", nil
 	}
 	if params == nil {
@@ -193,10 +196,9 @@ func (a *Auth) requestToken(ctx context.Context, params *TokenParams, opts *auth
 	var tokReq *TokenRequest
 	switch {
 	case opts.AuthCallback != nil:
-		a.log().Verbose("Auth: found AuthCallback in []AuthOption")
+		log.Verbose("Auth: found AuthCallback in []AuthOption")
 		v, err := opts.AuthCallback(context.TODO(), *params)
 		if err != nil {
-			a.log().Error("Auth: failed calling opts.AuthCallback ", err)
 			return nil, "", newError(ErrErrorFromClientTokenCallback, err)
 		}
 
@@ -220,10 +222,9 @@ func (a *Auth) requestToken(ctx context.Context, params *TokenParams, opts *auth
 			panic(fmt.Errorf("unhandled TokenLike: %T", v))
 		}
 	case opts.AuthURL != "":
-		a.log().Verbose("Auth: found AuthURL in []AuthOption")
+		log.Verbose("Auth: found AuthURL in []AuthOption")
 		res, err := a.requestAuthURL(ctx, params, opts)
 		if err != nil {
-			a.log().Error("Auth: failed calling requesting token with AuthURL ", err)
 			return nil, "", err
 		}
 		switch res := res.(type) {
@@ -234,7 +235,7 @@ func (a *Auth) requestToken(ctx context.Context, params *TokenParams, opts *auth
 			tokReqClientID = tokReq.ClientID
 		}
 	default:
-		a.log().Verbose("Auth: using default token request")
+		log.Verbose("Auth: using default token request")
 
 		req, err := a.createTokenRequest(params, opts)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


sacOO7:
For verbose logs, I don't feel we need to prefix with stacktrace, we need stacktrace only for errors, which is helpful for debugging.
I think we should add comprehensive stacktrace while logging errors

tcard:
> For verbose logs, I don't feel we need to prefix with stacktrace, we need stacktrace only for errors, which is helpful for debugging.

Well, I think it's also helpful for verbose logs, for the reasons I tried to explain above. I agree though that as a prefix is too noisy. I've moved it to the end.

> I think we should add comprehensive stacktrace while logging errors (rather than mini stacktrace)

This might be useful when debugging, but I think it would be too noisy otherwise. The key thing is that when we get logs from clients, we know where errors are happening, and scopes plus good error messages should make that unambiguous.

lmars:
@tcard 

> every operation should have at least a "begin" and "end" log message, with parameters and outcome

I'm not sure I agree with this, if we feel it's important to know when a given operations starts and ends, then shouldn't that just be logged as normal? For example:

```go
func (f *Foo) handleRequest(req *Request) {
	log := f.log.New("req.id", req.ID)

	log.Info("handling request")

	// ... code to handle the request

	log.Info("finished handling request", ...<context like response code and duration>)
}
```

> Log scopes fix that by prepending a mini stack trace to each log message, with IDs

I don't see the benefit of a stack trace, since doesn't the stream of logs for the given ID effectively give you the context of how the given operation has been handled, based on what has been logged for it?

QuintinWillison:
I've discussed with @tcard and we've decided not to proceed with this approach for the moment - it's a deviation from the approach taken by other client libraries so needs wider discussion were it to be adopted more widely. Focus right now is on pure 1.2 functionality.

PR #326, opened 2021-05-25
Replace all `ts *testing.T` instances with `t`

tcard:
Only the innermost `*testing.T` must be used from a test, that's why it
should always be named `t`, so that the `*testing.T` from outer scopes
are shadowed and can't be used accidentally. In fact, there were quite
a few instances of such accidental uses.

Also, add some of the missing `t.Parallel()`s.


sacOO7 (approved):
LGTM

PR #322, opened 2021-05-20
Integration/1.2

QuintinWillison:
We're getting closer to releasing version 1.2 of this library, so creating this pull request representing the work that will land to `main` before we do that. 🚀 

This pull request will remain in Draft state with no reviewers assigned until all the work intended to form part of the 1.2.0 release has been landed into the `integration/1.2` branch.

PR #328, opened 2021-05-26
[FIX CONFLICTS] Merge env. fallbacks to 1.2

sacOO7:
- Previous [env. fallbacks PR](https://github.com/ably/ably-go/pull/268) was merged to main, so it's throwing conflicts
- Current PR aims to fix **MERGE CONFLICTS** from **[INTEGRATION 1.2 TO MAIN](https://github.com/ably/ably-go/pull/322)**


ably/options.go
```diff
@@ -713,6 +797,10 @@ func WithDial(dial func(protocol string, u *url.URL, timeout time.Duration) (pro
 
 func applyOptionsWithDefaults(os ...ClientOption) *clientOptions {
 	to := defaultOptions
+	// No need to set hosts by default
+	to.RESTHost = ""
+	to.RealtimeHost = ""
+	to.FallbackHosts = nil
```
sacOO7, line 803: @tcard I have updated this one, to support env. fallbacks functionality, we need a way to detect which hosts were set and which were not, so we can't set them by default, they should be calculated as a part of fallback logic.

tcard (approved):
LGTM

lmars (approved):

PR #331, opened 2021-06-03
Don't reuse IV between encryptions

### LINKED ISSUE #330: IV shouldn't be reused between messages
IV is supposed to be unique for each piece of cleartext (ie. for each message).

The most direct way of fixing this is by removing IV from CipherParams. IV as a parameter is only useful for our testing with ably-common fixtures, and we should do that some other way.

tcard:
Fixes #330.

### DIFF
--- a/ably/proto/crypto.go
+++ b/ably/proto/crypto.go
@@ -57,12 +57,9 @@ type CipherParams struct {
 	// This is the private key used to  encrypt/decrypt payloads.
 	Key []byte // Spec item (TZ2d)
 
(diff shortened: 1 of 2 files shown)


lmars (approved):

sacOO7:
@tcard , you can merge this since approved. I reran tests, seems one of the workflow passing all tests

tcard:
@sacOO7 I'm not authorized (by GitHub) to merge it. I guess @QuintinWillison can?

PR #332, opened 2021-06-03
Unexport package proto, ablycrypto, ablytest

### LINKED ISSUE #291: Unexport package proto, ablycrypto, ablytest
We have some types that are part of the public API (Messages, Stats...) implemented in the [proto](https://pkg.go.dev/github.com/ably/ably-go@v1.2.0-apipreview.4/ably/proto) package. The problem is, that package also exposes internals that aren't required to be part of the public API, and having them exposed means we should be guaranteeing semver compatibility for those parts too, which gains us nothing.

So we have two options:

1. Move proto to [an internal directory](https://golang.org/cmd/go/#hdr-Internal_Directories). The problem then becomes that some types from proto are supposed to be public _and_ are consumed by proto. And, since ably (the public package) imports proto, proto can't import ably. So we'd need to spend time decoupling with interfaces and whatnot.
2. Just merge proto into the ably package, unexport the internal parts.

I favor 2; I don't see much point in having a separate proto package at all.

---

For the same reasons, [ablycrypto](https://pkg.go.dev/github.com/ably/ably-go@v1.2.0-apipreview.4/ably/ablycrypto) and [ablytest](https://pkg.go.dev/github.com/ably/ably-go@v1.2.0-apipreview.4/ably/ablytest) should be unexported, but those can be straightforwardly put into an internal directory.

tcard:
Mostly massive, boring moving and unexporting. The potentially interesting bits are in separate commits, for easier review.

Fixes #291.


sacOO7 (approved):
Reran tests again, seems to be working.

sacOO7:
@tcard , if you get time, you can take a look at failing tests.

tcard:
> @tcard , if you get time, you can take a look at failing flaky tests.

Opened #333.

PR #334, opened 2021-06-04
Unexport CipherParams.IV, useful only for tests

tcard:
Taking IV as a parameter is not just useless for production code, it can
easily become a security vulnerability.

See #330.


sacOO7 (approved):

sacOO7:
@tcard Not sure what `IV` exactly means

ben-xD:
I was curious too: https://en.wikipedia.org/wiki/Initialization_vector

tcard:
@sacOO7 I'm also not an expert at all, sorry. Any result from Google will do a better work than me at explaining this stuff.

PR #335, opened 2021-06-14
RTN7b: Handle ACK/NACK

### LINKED ISSUE #215: RTN7: ACK and NACK
Write spec tests for [RTN7](https://docs.ably.io/client-lib-development-guide/features/#RTN7), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

tcard:
Handling of ACK/NACK was too complicated and apparently broken.

Fixes #215.


tcard:
I just noticed [this](https://docs.ably.io/client-lib-development-guide/protocol/#message-acknowledgement):

> It is a protocol error if the system sends an ACK or NACK that skips past one or more msgSerial without there having been either and ACK or NACK; but a client in this situation should treat this case as implicitly @NACK@ing the skipped messages.
>
> It is also a protocol error if the system sends an ACK or NACK that covers a msgSerial that was covered by an earlier ACK or NACK; in such cases the client library must silently ignore the response insofar as it relates to @msgSerial@s that were covered previously (whether the response is the same now or different).

So it wasn't so simple after all. I'll get to it next.

tcard:
Done at 26cb5f0.

SimonWoolf:
```
// It is a protocol error if the system sends an ACK or NACK that skips past
// one or more msgSerial without there having been either and ACK or NACK;
// but a client in this situation should treat this case as
// implicitly @NACK@ing the skipped messages.
```
@tcard do I gather from this that this is what you think was happening with the customer?

If that is the case, ie if there was a protocol violation by realtime, then that needs to be investigated and fixed serverside.

But as I said in slack I am not convinced that that is the case. I continue to strongly suspect that the problem is that ably-go [is incorrectly setting the msgSerial to 0 in the case of a successful resume](https://github.com/ably/ably-go/blob/645ac50e99c48eea75ca31fd97bfed862752ad75/ably/realtime_conn.go#L776) , treating [RTN15c2](https://docs.ably.io/client-lib-development-guide/features/#RTN15c2) like [RTN15c3](https://docs.ably.io/client-lib-development-guide/features/#RTN15c3), given that you can see in the log the customer provided that the error occurred just after a successful resume.

tcard:
> do I gather from this that this is what you think was happening with the customer?

@SimonWoolf No, the implementation was broken also for the protocol-abiding case, as attested by the first test at e90f5f5, which used to fail. I'd expect just fixing that fixes the customer's problem.

This PR includes handling of those edge, "broken" cases just to cover everything.

tcard:
> But as I said in slack I am not convinced that that is the case. I continue to strongly suspect that the problem is that ably-go [is incorrectly setting the msgSerial to 0 in the case of a successful resume](https://github.com/ably/ably-go/blob/645ac50e99c48eea75ca31fd97bfed862752ad75/ably/realtime_conn.go#L776) , treating [RTN15c2](https://docs.ably.io/client-lib-development-guide/features/#RTN15c2) like [RTN15c3](https://docs.ably.io/client-lib-development-guide/features/#RTN15c3), given that you can see in the log the customer provided that the error occurred just after a successful resume.

We already have tests in place for RTN15c2 and RTN15c3, but I'll double-check they cover everything.

tcard:
@SimonWoolf You were right, RTN15c2 was _also_ broken. Should be fixed by c7f339f, PTAL.

ably/state.go
```diff
@@ -120,85 +120,51 @@ type msgCh struct {
...
-		nack = i
-		ack = min(i+msg.Count, q.Len())
-	default:
-		nack = i + 1
-		ack = min(i+1+msg.Count, q.Len())
+	// The msgSerial from the server may not be the same we're waiting. If the
+	// server skipped some messages, they get implicitly NACKed. If the server
+	// ACKed some messages again, we ignore those. In both cases, we just need
+	// to correct the number of messages that get ACKed by that difference.
+	serialShift := int(msg.MsgSerial - q.queue[0].msg.MsgSerial)
```
tcard, line 147: I realized I can't just do this, because the msgSerial distance between consecutive messages isn't always 1. Non-MESSAGE (e. g. ATTACH, PRESENCE...) protocol messages also increase msgSerial.

Will fix.

tcard: See 9896fce.

tcard: @SimonWoolf clarified this isn't true; see https://github.com/ably/docs/pull/1115.

tcard:
@SimonWoolf PTAL

SimonWoolf (approved):

PR #336, opened 2021-06-24
RTC8a: client-requested reauthorization while CONNECTED

tcard:
Part of #228. Implements [RTC8a](https://docs.ably.io/client-lib-development-guide/features/#RTC8a) for explicit calls to `Authorize`.

This work is a prerequisite for RTN22 and the rest of RTC8.

### DIFF
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -68,6 +68,10 @@ type Auth struct {
 	host     string       // a host part of AuthURL
 	clientID string       // clientID of the authenticated user or wildcard "*"
 
+	// onClientAuthorize is the callback that Realtime sets to reauthorize with the
+	// server when Authorize is explicitly called.
+	onClientAuthorize func(context.Context, *TokenDetails)
+
 	serverTimeOffset time.Duration
 
 	// ServerTimeHandler when provided this will be used to query server time.
@@ -76,7 +80,8 @@ type Auth struct {
 
 func newAuth(client *REST) (*Auth, error) {
 	a := &Auth{
-		client: client,
+		client:            client,
+		onClientAuthorize: func(context.Context, *TokenDetails) {},
 	}
 	method, err := detectAuthMethod(a.opts())
 	if err != nil {
@@ -266,8 +271,13 @@ func (a *Auth) Authorize(ctx context.Context, params *TokenParams, setOpts ...Au
 		opts = applyAuthOptionsWithDefaults(setOpts...)
 	}
 	a.mtx.Lock()
-	defer a.mtx.Unlock()
-	return a.authorize(ctx, params, opts, true)
+	token, err := a.authorize(ctx, params, opts, true)
+	a.mtx.Unlock()
+	if err != nil {
+		return nil, err
+	}
+	a.onClientAuthorize(ctx, token)
+	return token, nil
 }
 
 func (a *Auth) authorize(ctx context.Context, params *TokenParams, opts *authOptions, force bool) (*TokenDetails, error) {
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -68,9 +68,9 @@ type Auth struct {
 	host     string       // a host part of AuthURL
 	clientID string       // clientID of the authenticated user or wildcard "*"
 
-	// onClientAuthorize is the callback that Realtime sets to reauthorize with the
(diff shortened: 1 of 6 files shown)


ably/realtime_conn.go
```diff
@@ -880,6 +880,33 @@ func (c *Connection) reauthorize(arg connArgs) {
 	c.reconnect(arg)
 }
 
+func (c *Connection) onClientAuthorize(ctx context.Context, token *TokenDetails) {
```
sacOO7, line 883: This callbacks feels a bit misleading, it feels it gets called when client is authorised. Can we not make it callback and just change method name to `authorizeClient`. I believe thats what it does after receiving token. 
PS. if there is a name for callback, it should be `onClientTokenReceived`

ably/auth.go
```diff
@@ -266,8 +271,13 @@ func (a *Auth) Authorize(ctx context.Context, params *TokenParams, setOpts ...Au
 		opts = applyAuthOptionsWithDefaults(setOpts...)
 	}
 	a.mtx.Lock()
-	defer a.mtx.Unlock()
-	return a.authorize(ctx, params, opts, true)
+	token, err := a.authorize(ctx, params, opts, true)
```
sacOO7, line 274: instead of `authorize`, should be renamed to `getAuthToken`, because it certainly doesn't do proper authentication

ably/realtime_conn_spec_test.go
```diff
@@ -3390,6 +3390,185 @@ func TestIdempotentACK(t *testing.T) {
...
+		assertEquals(t, chFailed, ch.State())
+	})
+
+	t.Run("RTC8a3: Authorize waits for connection update", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		connectedMsg := intercept(ctx, ably.ActionConnected)
+
+		authorizeDone := make(chan struct{})
+		go func() {
```
sacOO7, line 3531: open goroutine

ably/auth.go
sacOO7: I was going through the file, maybe we need to rename methods that doesn't do actual authorization, some of the code is being used for token related validation. Maybe we should extract it out to be more clear about what's happening behind the scenes

sacOO7:
Also, not sure why all tests are failing : (

ably/realtime_conn.go
tcard: `onClientAuthorize` isn't a great name but `onClientTokenReceived` would be also misleading, because this is supposed to be called when the `Authorize` public method is explicitly called, not any time the client gets a token.

Renamed to `onExplicitAuthorize` at 730615d.

ably/realtime_conn_spec_test.go
tcard: Yes, we wait for `authorizeDone` below, so we check there isn't a leak.

ably/auth.go
tcard: What do you mean it doesn't do proper authentication?

In any case, could we move this to a separate pull request? This code was already there.

tcard:
This test is failing a lot:

```
--- FAIL: TestRealtimePresence_Sync250 (6.78s)
    realtime_presence_test.go:115: members2: clientID="client149" not found in presence map
```

But it seems unrelated to this PR, and it passes on my machine. Opened https://github.com/ably/ably-go/issues/338.

ably/realtime_conn_spec_test.go
sacOO7: Still, its a explicit go routine, we can follow standard `rgGroup`

ably/auth.go
sacOO7: Yes, that would be great, or rather I feel its a small change, we can refactor it if we can

ably/realtime_conn_spec_test.go
tcard: Don't really see the point but done at eb709f3.

tcard:
@sacOO7 Anything blocking approval?

ably/realtime_conn.go
sacOO7: I still think naming can be improved. I will approve it for now

sacOO7 (approved):

PR #355, opened 2021-07-09
Release/1.2.0

QuintinWillison:
Congratulations and thanks to @tcard, @sacOO7, @gernest and @lmars for all your work in getting this release to where we are now! 🚀

### DIFF
--- a/ably/realtime_conn_spec_test.go
+++ b/ably/realtime_conn_spec_test.go
@@ -107,7 +107,7 @@ func Test_RTN2_WebsocketQueryParams(t *testing.T) {
 		t.Parallel()
 		requestParams := setup()
 		lib := requestParams["lib"]
-		assertDeepEquals(t, []string{"go-1.2.0-apipreview.4"}, lib)
+		assertDeepEquals(t, []string{"go-1.2.0"}, lib)
 	})
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,14 +184,12 @@ Version 1.2.0 is out of pre-release!
 
 **Fixed bugs:**
 
-- Please ignore this issue \(testing sync\) [\#203](https://github.com/ably/ably-go/issues/203)
 - Rest.request call timing out doesn't result in the errorMessage being set in the httpPaginatedResponse? [\#192](https://github.com/ably/ably-go/issues/192)
 - Channel attach fails in the DISCONNECTED state [\#189](https://github.com/ably/ably-go/issues/189)
 - TestFixConnLeak\_ISSUE89 is flaky [\#132](https://github.com/ably/ably-go/issues/132)
 
 **Closed issues:**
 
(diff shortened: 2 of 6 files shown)


sacOO7:
LGTM : )

sacOO7 (approved):

AndyNicks (approved):
LGTM

niksilver (approved):
LGTM

ably/realtime_conn_spec_test.go
```diff
@@ -107,7 +107,7 @@ func Test_RTN2_WebsocketQueryParams(t *testing.T) {
 		t.Parallel()
 		requestParams := setup()
 		lib := requestParams["lib"]
-		assertDeepEquals(t, []string{"go-1.2.0-apipreview.4"}, lib)
+		assertDeepEquals(t, []string{"go-1.2.0"}, lib)
```
lmars, line 110: Would this be better as `"go-" + libraryVersion` to avoid changing it every time we release?

lmars (approved):

ben-xD (approved):

ably/realtime_conn_spec_test.go
sacOO7: We generally hardcode strings in the expected assertions, so I don't think we should change it.

sacOO7: It's not a good idea to put logic inside test for value being asserted against, can lead to non-failing test ...

QuintinWillison: I believe this is modelled on what we do elsewhere, where the point of the test is _not_ to use the same constant to test with. I'm going to leave this as is for now but am happy for suggestions for how to perhaps approach this differently for subsequent releases.

lmars: I know this is merged so we can discuss separately, but I believe the point of this test is to make sure "the library version" appears in the query param, and "the library version" is whatever is in the `libraryVersion` constant, not the static string of characters `1.2.0`.

Changing the value of "the library version" to, say, `1.2.1`, does not break the implemented functionality (i.e. that the library version appears in the query param, regardless of what its value is), so the test really shouldn't break in that case.

sacOO7: That also means there's no point of having test if it's not going to fail.

lmars: This test _should not_ fail when we bump the release version, because bumping the release version _does not_ break the functionality being tested here (i.e. that the current library version appears in the `lib` query param).

If, say, someone changed the code so that it sets a `library` query param rather than a `lib` query param, then that is something that should fail this test, because then the functionality has been broken, since as per the spec, it should be `lib` not `library`, but the spec doesn't say the value of the query param should always equal `1.2.0`.

Bumping a release version really shouldn't involve changing tests in the same changeset.

PR #356, opened 2021-07-11
RTC8b/c, RTN22: Server-initiated AUTH

tcard:
Fixes #228.

PR #357, opened 2021-07-12
RTC8a: client-requested reauthorization while CONNECTED (to `main` branch)

tcard:
Contains just the work that was originally done under https://github.com/ably/ably-go/pull/336.

### DIFF
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -68,6 +68,10 @@ type Auth struct {
 	host     string       // a host part of AuthURL
 	clientID string       // clientID of the authenticated user or wildcard "*"
 
+	// onExplicitAuthorize is the callback that Realtime sets to reauthorize with the
+	// server when Authorize is explicitly called.
+	onExplicitAuthorize func(context.Context, *TokenDetails)
+
 	serverTimeOffset time.Duration
 
 	// ServerTimeHandler when provided this will be used to query server time.
@@ -76,7 +80,8 @@ type Auth struct {
 
 func newAuth(client *REST) (*Auth, error) {
 	a := &Auth{
-		client: client,
+		client:              client,
+		onExplicitAuthorize: func(context.Context, *TokenDetails) {},
 	}
 	method, err := detectAuthMethod(a.opts())
 	if err != nil {
@@ -266,8 +271,13 @@ func (a *Auth) Authorize(ctx context.Context, params *TokenParams, setOpts ...Au
 		opts = applyAuthOptionsWithDefaults(setOpts...)
 	}
 	a.mtx.Lock()
-	defer a.mtx.Unlock()
-	return a.authorize(ctx, params, opts, true)
+	token, err := a.authorize(ctx, params, opts, true)
+	a.mtx.Unlock()
+	if err != nil {
+		return nil, err
+	}
+	a.onExplicitAuthorize(ctx, token)
+	return token, nil
 }
 
 func (a *Auth) authorize(ctx context.Context, params *TokenParams, opts *authOptions, force bool) (*TokenDetails, error) {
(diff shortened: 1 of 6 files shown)


ably/auth.go
```diff
@@ -266,8 +271,13 @@ func (a *Auth) Authorize(ctx context.Context, params *TokenParams, setOpts ...Au
 		opts = applyAuthOptionsWithDefaults(setOpts...)
 	}
 	a.mtx.Lock()
-	defer a.mtx.Unlock()
-	return a.authorize(ctx, params, opts, true)
+	token, err := a.authorize(ctx, params, opts, true)
+	a.mtx.Unlock()
```
lmars, line 275: It feels wrong to hold a mutex for the duration of a remote request, what exactly is this protecting? Does this block anything that needs access to the current auth details whilst the request is in flight?

ably/realtime_conn.go
```diff
@@ -880,6 +880,33 @@ func (c *Connection) reauthorize(arg connArgs) {
...
+		}
+
+		c.send(&protocolMessage{
+			Action: actionAuth,
+			Auth:   &authDetails{AccessToken: token.Token},
+		}, nil)
+
+		select {
+		case <-ctx.Done():
+		case <-changes:
```
lmars, line 905: If this receives a failed event, should that not be passed back to the original caller of `Authorize`? Otherwise, why does `Authorize` need to block on this call?

ably/auth.go
tcard: Auth remembers (caches) the last token, among other auth-related stuff. That's what this mutex protects.

I'm not sure if anything relies on this lock covering the whole request, but it makes some intuitive sense to me because otherwise several parallel authorization requests could be launched that would then race to set the current token, and then callers wouldn't agree on which the current token is, etc. It probably could be changed to a RWMutex, though.

(Note that this isn't new in this PR nor related to RTC8a, so, if needed, can be analyzed and fixed separately.)

ably/realtime_conn.go
tcard: True. It makes sense and RTC8a3 makes it clear.

To avoid conflicts with #356, which builds on top of this, do you mind if I handle this there?

ably/auth.go
lmars: > Auth remembers (caches) the last token, among other auth-related stuff. That's what this mutex protects.

Ok, so I do think a mutex lock should always wrap _just_ the memory addresses it's protecting in an obvious way, since in my experience, the further away the wrapping gets, the more likely we are to encounter deadlocks.

> several parallel authorization requests could be launched that would then race to set the current token, and then callers wouldn't agree on which the current token is, etc. It probably could be changed to a RWMutex, though.

Ok I understand, in that case then I don't think a mutex is the correct primitive to use, I would suggest a [`singleflight.Group`](https://pkg.go.dev/golang.org/x/sync/singleflight#Group.Do).

> Note that this isn't new in this PR nor related to RTC8a, so, if needed, can be analyzed and fixed separately.

SGTM 👍

ably/realtime_conn.go
lmars: Sure 👍

lmars (approved):

ably/realtime_conn.go
tcard: See https://github.com/ably/ably-go/pull/356/commits/1b2142e58b49788fe236afad8b65e195b378bcf9.

ably/auth.go
tcard: Singleflight doesn't really fit here because the params passed to Authorize could be different, so calls wouldn't be expected to be idempotent. You shouldn't just return the return value of another concurrent call.

I've linked to this conversation in a relevant issue: https://github.com/ably/ably-go/issues/184

PR #360, opened 2021-07-15
Ably agent header

### LINKED ISSUE #302: Implement RSC7d (Ably-Agent header)
see [PR](https://github.com/ably/docs/pull/1034) and [spec point](https://ably-docs.herokuapp.com/client-lib-development-guide/features/#RSC7d) for context.

sacOO7:
- Fixed #302

### DIFF
--- a/ably/proto_http.go
+++ b/ably/proto_http.go
@@ -6,10 +6,12 @@ const (
 	ablyLibHeader          = "X-Ably-Lib"
 	ablyErrorCodeHeader    = "X-Ably-Errorcode"
 	ablyErrorMessageHeader = "X-Ably-Errormessage"
-	libraryVersion         = "1.2.0-apipreview.4"
+	libraryVersion         = "1.2.0"
 	libraryName            = "go"
 	libraryString          = libraryName + "-" + libraryVersion
 	ablyVersion            = "1.2"
 	ablyClientIDHeader     = "X-Ably-ClientId"
 	hostHeader             = "Host"
+	ablyAgentHeader        = "Ably-Agent"
+	ablyAgentIdentifier    = "ably-go/" + libraryVersion
 )
--- a/ably/rest_client.go
+++ b/ably/rest_client.go
@@ -695,6 +695,7 @@ func (c *REST) newHTTPRequest(ctx context.Context, r *request) (*http.Request, e
 	req.Header.Set("Accept", protocol) //spec RSC19c
 	req.Header.Set(ablyVersionHeader, ablyVersion)
 	req.Header.Set(ablyLibHeader, libraryString)
+	req.Header.Set(ablyAgentHeader, ablyAgentIdentifier)
 	if c.opts.ClientID != "" && c.Auth.method == authBasic {
 		// References RSA7e2
 		h := base64.StdEncoding.EncodeToString([]byte(c.opts.ClientID))


lmars (changes requested):
This should also include the removal of `X-Ably-Lib`

ably/proto_http_windows.go
```diff
@@ -0,0 +1,16 @@
...
+	"strconv"
+
+	"golang.org/x/sys/windows"
+)
+
+func goOSIdentifier() string {
+	versionInfo := windows.RtlGetVersion()
+	semanticVersion := strconv.Itoa(int(versionInfo.MajorVersion)) + "." +
+		strconv.Itoa(int(versionInfo.MinorVersion)) + "." +
+		strconv.Itoa(int(versionInfo.BuildNumber))
```
lmars, line 14: Probably easier to use `fmt.Sprintf`:

```go
v := windows.RtlGetVersion()
return fmt.Sprintf("%s/%d.%d.%d", runtime.GOOS, v.MajorVersion, v.MinorVersion, v.BuildNumber)
```

ably/realtime_client_test.go
```diff
@@ -47,6 +49,42 @@ func TestRealtime_RealtimeHost(t *testing.T) {
 	}
 }
 
+func TestReatime_RSC7_AblyAgent(t *testing.T) {
+	t.Parallel()
+
+	app, err := ablytest.NewSandbox(nil)
```
lmars, line 55: Why does this test need a Sandbox app?

ably/realtime_client_test.go
```diff
@@ -47,6 +49,42 @@ func TestRealtime_RealtimeHost(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer app.Close()
+
+	t.Run("RSC7d3 : Should set fallback host header", func(t *testing.T) {
```
lmars, line 61: This test description is incorrect.

ably/realtime_client_test.go
```diff
@@ -47,6 +49,42 @@ func TestRealtime_RealtimeHost(t *testing.T) {
...
+		defer server.Close()
+		serverURL, _ := url.Parse(server.URL)
+
+		client, err := ably.NewRealtime(
+			app.Options(ably.WithEnvironment(ablytest.Environment),
+				ably.WithTLS(false),
+				ably.WithUseTokenAuth(true),
+				ably.WithRealtimeHost(serverURL.Host))...)
+
+		defer client.Close()
```
lmars, line 77: This `defer` should be after the error check.

ably/realtime_conn_spec_test.go
```diff
@@ -100,14 +100,14 @@ func Test_RTN2_WebsocketQueryParams(t *testing.T) {
...
-		assertDeepEquals(t, []string{"1.2"}, libVersion)
+		assertDeepEquals(t, []string{ably.AblyVersion}, libVersion)
 	})
 
 	t.Run("RTN2g: library and version should be included as the value of lib param", func(t *testing.T) {
 		t.Parallel()
 		requestParams := setup()
 		lib := requestParams["lib"]
-		assertDeepEquals(t, []string{"go-1.2.0"}, lib)
+		assertDeepEquals(t, []string{"go-" + ably.LibraryVersion}, lib)
```
lmars, line 110: [As mentioned](https://github.com/ably/ably-go/pull/360#pullrequestreview-715264554), we shouldn't be sending the deprecated lib value anymore.

ably/rest_client_test.go
```diff
@@ -270,11 +270,56 @@ func TestRSC7(t *testing.T) {
...
+		}))
+		defer server.Close()
+
+		opts := []ably.ClientOption{
+			ably.WithEnvironment(ablytest.Environment),
+			ably.WithTLS(false),
+			ably.WithUseTokenAuth(true),
 		}
+		// set up the proxy to forward all requests except a specific fallback to the server,
+		// whilst that fallback goes to the regular endpoint
```
lmars, line 297: What does this comment have to do with testing the Ably-Agent header?

ably/rest_client_test.go
```diff
@@ -270,11 +270,56 @@ func TestRSC7(t *testing.T) {
...
+
+		var agentHeaderValue string
+		proxy := func(r *http.Request) (*url.URL, error) {
+			agentHeaderValue = r.Header.Get(ably.AblyAgentHeader)
+			return serverURL, nil
+		}
+
+		opts = append(opts, ably.WithHTTPClient(&http.Client{
+			Transport: &http.Transport{
+				Proxy:        proxy,
```
lmars, line 308: Why does this test need to use a proxy?

ably/rest_client_test.go
```diff
@@ -270,11 +270,56 @@ func TestRSC7(t *testing.T) {
...
+			},
+		}))
+
+		client, err := ably.NewREST(app.Options(opts...)...)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		channel := client.Channels.Get("remember_fallback_host")
+		channel.Publish(context.Background(), "ping", "pong")
```
lmars, line 319: Would it be simpler to call [`client.Time`](https://pkg.go.dev/github.com/ably/ably-go@v1.2.0/ably#REST.Time)?

ably/websocket.go
```diff
@@ -60,6 +60,7 @@ func dialWebsocket(proto string, u *url.URL, timeout time.Duration) (*websocketC
 // dialWebsocketTimeout dials the websocket with a timeout.
 func dialWebsocketTimeout(uri, protocol, origin string, timeout time.Duration) (*websocket.Conn, error) {
 	config, err := websocket.NewConfig(uri, origin)
+	config.Header.Set(ablyAgentHeader, getAblyAgentIdentifier())
```
lmars, line 63: This should come after the error check in case `config` is nil when an error is returned.

ably/proto_http.go
```diff
@@ -12,4 +16,27 @@ const (
 	ablyVersion            = "1.2"
 	ablyClientIDHeader     = "X-Ably-ClientId"
 	hostHeader             = "Host"
+	ablyAgentHeader        = "Ably-Agent"                // RSC7d
+	ablySDKIdentifier      = "ably-go/" + libraryVersion // RSC7d1
 )
+
+var ablyAgentIdentifier = "" // Need to be calculated at runtime
```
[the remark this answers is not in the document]
sacOO7: Yeah, not sure why I forgot it. very stupid mistake ; (

ably/proto_http_unix.go
```diff
@@ -0,0 +1,20 @@
...
+package ably
+
+import (
+	"bytes"
+	"runtime"
+
+	"golang.org/x/sys/unix"
+)
+
+func goOSIdentifier() string {
```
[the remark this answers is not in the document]
sacOO7: I think part of it has to do with it's only being used from `proto_http` , once file renamed to something completely different, it's difficult to find files in project view on the side

ably/realtime_client_test.go
sacOO7: It needs for generating `auth token`, if not used, it fails internally with implicit auth request

ably/realtime_conn_spec_test.go
sacOO7: Can you please mention which headers have been deprecated

ably/rest_client_test.go
sacOO7: Yes, misleading comment

ably/rest_client_test.go
sacOO7: Yes, I strongly remember one of the issue, where I could only read headers as a part of proxy. Well, its working in realtime test, there is no point to keep the same in rest tests

ably/rest_client_test.go
sacOO7: Done

ably/websocket.go
sacOO7: Done

ably/proto_http_unix.go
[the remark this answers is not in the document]
lmars: I don't think a file should be named based on the names of the files that call the functions in it, it should be named based on what's in the file, and here there is nothing in the file that's got anything to do with "proto_http". If you feel strongly though, feel free to leave it named this way.

ably/realtime_client_test.go
lmars: Can't it just use a fake key? This puts needless load on our Sandbox cluster.

ably/realtime_conn_spec_test.go
lmars: The whole [RSC7b](https://docs.ably.io/client-lib-development-guide/features/#RSC7b) spec item has been deprecated, so we should no longer be sending the "lib version" in either the `X-Ably-Lib` header or the `lib` query parameter.

sacOO7: Done

ably/realtime_client_test.go
sacOO7: Actually I am not quite sure, how to do it

lmars: For example:

```go
client, err := ably.NewRealtime(ably.WithKey("xxx:xxx"))
```

sacOO7: I will check with existing implementation

sacOO7: `WithKey` does't work, because it doesn't suppoer non-TLS, had to use `withToken`, so done

PR #365, opened 2021-07-28
Clarify event emitter concurrency documentation

tcard:
Wording was too terse and ambiguous before.

### DIFF
--- a/ably/doc.go
+++ b/ably/doc.go
@@ -24,9 +24,12 @@
 // The OffAll method is like Off, except it is like calling all the "off"
 // functions.
 //
-// Calls to handlers for a single or for subsequent events may be concurrent,
-// but are always ordered. That is, if an event is fired after another, the same
-// handler will be called in the same order that the events were fired.
+// Each handler is assigned its own sequential queue of events. That is, any
+// given handler function will not receive calls from different goroutines that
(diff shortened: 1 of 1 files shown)


marklewin (approved):
LGTM!

ably/doc.go
```diff
@@ -24,9 +24,12 @@
 // The OffAll method is like Off, except it is like calling all the "off"
 // functions.
 //
-// Calls to handlers for a single or for subsequent events may be concurrent,
-// but are always ordered. That is, if an event is fired after another, the same
-// handler will be called in the same order that the events were fired.
+// Each handler is assigned its own sequential queue of events. That is, any
```
sacOO7, line 27: Does `handlers emit events here sequentially (FIFO)` ? If it does, it should be mentioned explicitly.

tcard: It says: "you can count on events or messages to be delivered to the handler in the same order they were emitted". That's pretty explicit, isn't it?

sacOO7: Mentioning, queue or just FIFO makes things explicit at a good level. That said, I will keep that change to be optional. But, I strongly feel it will help emitter behaviour more better.

sacOO7 (approved):

PR #370, opened 2021-08-06
Don't set message encoding for valid utf-8 string data

lmars:
See explanation in https://github.com/ably/docs/issues/1165.

### DIFF
--- a/ably/proto_message.go
+++ b/ably/proto_message.go
@@ -44,16 +44,12 @@ func (m Message) withEncodedData(cipher channelCipher) (Message, error) {
 		return m, nil
 	}
 
-	switch d := m.Data.(type) {
-	case string:
-		if utf8.ValidString(d) {
-			m.Encoding = mergeEncoding(m.Encoding, encUTF8)
-		} else {
-			// If string isn't UTF-8, convert to []byte to encode it as base64
-			// below.
-			m.Data = []byte(d)
-		}
+	// If string isn't UTF-8, convert to []byte to encode it as base64
+	// below.
+	if d, ok := m.Data.(string); ok && !utf8.ValidString(d) {
+		m.Data = []byte(d)
 	}
+
 	switch d := m.Data.(type) {
 	case string:
 	case []byte:
--- a/ably/proto_message.go
+++ b/ably/proto_message.go
@@ -83,10 +83,10 @@ func (m Message) withEncodedData(cipher channelCipher) (Message, error) {
 	if err != nil {
 		return Message{}, fmt.Errorf("encrypting message data: %w", err)
 	}
-	// if the plain text is a string, then add utf-8 to the list of encodings
-	// to indicate to the client-side decoder that the plain text should be
-	// further decoded into a utf-8 string after being decrypted.
-	if _, ok := m.Data.(string); ok {
+	// if the plain text is a valid utf-8 string, then add utf-8 to the list
+	// of encodings to indicate to the client-side decoder that the plain text
+	// should be further decoded into a utf-8 string after being decrypted.
+	if s, ok := m.Data.(string); ok && utf8.ValidString(s) {
 		m.Encoding = mergeEncoding(m.Encoding, encUTF8)
 	}
 	m.Data = base64.StdEncoding.EncodeToString(e)
--- a/ably/proto_message.go
+++ b/ably/proto_message.go
@@ -84,9 +84,10 @@ func (m Message) withEncodedData(cipher channelCipher) (Message, error) {
 	if err != nil {
 		return Message{}, fmt.Errorf("encrypting message data: %w", err)
 	}
+	// if the plain text is a string, then add utf-8 to the list of encodings
+	// to indicate to the client-side decoder that the plain text should be
+	// further decoded into a utf-8 string after being decrypted.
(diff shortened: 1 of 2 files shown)


sacOO7:
@lmars , seems `utf8` encoding is also used at one more place. Also, prev. comment has been added as to no idea why it's there.
Can you check if it's needed
https://github.com/ably/ably-go/blob/4c447dc7dce81ae8b1f1cab93beff5cef599c7eb/ably/proto_message.go#L87-L91

sacOO7:
Also, we can remove this //todo comment, since proto is merged into ably 
https://github.com/ably/ably-go/blob/4c447dc7dce81ae8b1f1cab93beff5cef599c7eb/ably/proto_message.go#L42

lmars:
@sacOO7 thanks, I've clarified the requirement to add utf-8 encoding to encrypted strings in 99a3f68, and removed that outdated comment in a57270a.

ably/proto_message.go
```diff
@@ -88,9 +83,10 @@ func (m Message) withEncodedData(cipher channelCipher) (Message, error) {
 	if err != nil {
 		return Message{}, fmt.Errorf("encrypting message data: %w", err)
 	}
+	// if the plain text is a string, then add utf-8 to the list of encodings
+	// to indicate to the client-side decoder that the plain text should be
+	// further decoded into a utf-8 string after being decrypted.
```
sacOO7, line 88: Correct me if I am wrong, according to @SimonWoolf UTF-8 represents binary payload. So, we shouldn't mention UTF-8 encoding in the first place, because we are sending payload as a string (Encoding doesn't matter, it's a string in the end). I am not sure, but I feel this might case issue on js client, since it will expect binary payload as a part of UTF-8 encoding.

sacOO7: Also, not sure `if check` makes sense or not, Since, it's at a point where `_, ok := m.Data.(string)` will always return true

lmars: > UTF-8 represents binary payload

That is correct, and when the client decrypts this cipher-text, they will get a binary payload, and thus we need to include the utf-8 encoding to further indicate that this decrypted binary payload is in fact a utf-8 encoded string.

This is distinctly different to when the publisher simply publishes a string, because then it's just a string, there is no binary payload involved.

Does that make sense? This should definitely be clarified when addressing the linked docs issue (https://github.com/ably/docs/issues/1165).

lmars: > not sure if check makes sense or not, Since, it's at a point where _, ok := m.Data.(string) will always return true

Let's keep the check, as it makes the logic explicit that it only applies to string payloads, and does not rely on other logic in the file that may change in the future.

sacOO7: Then you need to modify `if` check, since current one will always return true. 
`Data` field of `Message` is currently being mutated, to the point, it will always store encoded string. So, it's difficult to guess original encoding binary/text.

sacOO7: Then you need to modify `if` check, since current one will always return true. 
`Data` field of `Message` is currently being mutated, to the point, it will always store encoded string. So, it's difficult to guess original encoding binary/text.

sacOO7: Also, there should be one more check whether its a valid `utf-8` string.

lmars: > Then you need to modify if check, since current one will always return true.

As mentioned in my last comment, I'm going to keep the check, as it defends against logic changing elsewhere and utf-8 encoding being added when `m.Data` is not a string.

> Data field of Message is currently being mutated, to the point, it will always store encoded string. So, it's difficult to guess original encoding binary/text.

Since this encoding logic is iterative, there is no need to know the encoding of the _original_ payload, just whatever the payload of `m.Data` is at the time of adding the next encoding in the chain.

For example, if `m.Data` is originally some arbitrary object, then it will get serialised as JSON and the `json` encoding added [here](https://github.com/ably/ably-go/blob/main/ably/proto_message.go#L73-L74). Now, from this point on, the payload _is_ a utf-8 encoded string, it doesn't matter from that point onwards that the original payload was not a string, since that's been covered by the addition of the `json` encoding.

This is explicitly tested [here](https://github.com/ably/ably-go/blob/main/ably/proto_message_test.go#L76-L86), and this PR should not change those semantics.

> Also, there should be one more check whether its a valid utf-8 string, then only specify UTF-8 encoding.

Good spot, updated in 153c932 👍

sacOO7: I am resolving this conversation for the time being. Though I am not sure, if UTF-8 will work or not.

sacOO7:
@lmars , not sure but I feel when we are using cipher, we are adding extra layer of encoding for byte [] type, which increases payload size. I was thinking maybe you can improve on that one.

lmars:
@sacOO7 

> this is not part of the fix. but I just feel we can include it as a part of fix

Yes, so whilst what you describe could potentially be a problem, let's deal with that separately, the priority here is fixing the behaviour we know is broken. If you're happy that we've fixed the reported issue, please approve this PR so we can merge and release it.

sacOO7 (approved):

sacOO7:
Fixes incomplete  https://github.com/ably/ably-go/pull/107

PR #383, opened 2021-09-01
Fix CI

QuintinWillison:
I'm going to work on getting checks against this pull request green before attempting to gain confidence in any of the other open pull requests.

[This has been failing](https://github.com/ably/ably-go/actions?query=branch%3Amain) for months. 🤯


QuintinWillison:
ok, sorry for the noise but removing reviewers until I find out why this is failing.

QuintinWillison:
For all checks to pass this must be nearly there in terms of getting us CI stability... Before I land this PR I will create issues against the individual commits I had to add where tests were skipped as refactoring / fixing those tests is now technical backlog for this repository.

paddybyers (approved):
I have not reviewed in detail, but I support the process of reinstating green CI.

Going forward I think a fuller PR description would be useful so we know what the aims and scope and approach of this PR are, without having to look at the code.

I also think we should stick more closely to our internal guidance on commit comments, because these are verging on unhelpful.

lmars (commented):
LGTM

PR #432, opened 2021-10-31
Custom test error message

sacOO7:
- Added test error message formatter
- Updated `assertEquals` with optional formatter


sacOO7:
@tomkirbygreen I have referenced most of the error formatter code from the popular go testing library https://github.com/stretchr/testify. It is also to make sure, go developers are familiar with the general error format.

PR #437, opened 2022-01-27
Add `ConnectionKey` to message

ken8203:
If you are publishing from a REST client and want to prevent one particular realtime client (which has echoMessages set to false) from receiving the message, as if that client was the one that published the message, you can do that by setting the connectionKey.

https://faqs.ably.com/is-it-possible-to-prevent-messages-published-being-echoed-back-to-the-publishing-client

PR #443, opened 2022-02-03
Add connection key to messages.

### LINKED ISSUE #442: Add ConnectionKey to message
If you are publishing from a REST client and want to prevent one particular realtime client (which has echoMessages set to false) from receiving the message, as if that client was the one that published the message, you can do that by setting the connectionKey.

┆sprint: SDK 2022-Q1-S3

Rosalita:
This fixes #442  a customer previously raised a PR to add ConnectionKey to the message. However the value of this new connection key field was never populated, so it still was not possible for a rest client to publish on behalf of a realtime client. This PR populates the connection key when rest messages are published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ if err != nil {
 
 // A REST client can publish messages on behalf of a realtime client
 // by providing the connection key of the realtime client.
-err := channel.Publish(ctx, "temperature", "12.7", ably.PublishWithConnectionKey("108yZeOBAB7iva!QQJ2CnCSa6A4ql22-a7443108yZeOBAB7iva"))
+err := channel.Publish(ctx, "temperature", "12.7", ably.PublishWithConnectionKey("ConnectionKeyOfPublisherClientHere"))
 if err != nil {
         panic(err)
 }
--- a/README.md
+++ b/README.md
@@ -224,6 +224,13 @@ err = channel.PublishMultiple(ctx, []*ably.Message{
 if err != nil {
         panic(err)
 }
+
+// A REST client can publish messages on behalf of a realtime client
+// by providing the connection key of the realtime client.
+err := channel.Publish(ctx, "temperature", "12.7", ably.PublishWithConnectionKey("108yZeOBAB7iva!QQJ2CnCSa6A4ql22-a7443108yZeOBAB7iva"))
+if err != nil {
+        panic(err)
+}
 ```
 
 ### Querying the History
--- a/ably/rest_channel.go
+++ b/ably/rest_channel.go
@@ -52,23 +52,13 @@ func newRESTChannel(name string, client *REST) *RESTChannel {
 }
 
 // Publish publishes a message on the channel.
-func (c *RESTChannel) Publish(ctx context.Context, name string, data interface{}) error {
-	//Check to see if data received is of type message.
-	if msg, ok := data.(Message); ok {
-		// Publish the message with connection key.
-		return c.PublishMultiple(ctx, []*Message{
-			{Name: name, Data: data, ConnectionKey: msg.ConnectionKey},
-		})
-	}
-	// Only the name and data are published.
-	return c.PublishMultiple(ctx, []*Message{
-		{Name: name, Data: data},
-	})
+func (c *RESTChannel) Publish(ctx context.Context, name string, data interface{}, options ...PublishMultipleOption) error {
+	return c.PublishMultiple(ctx, []*Message{{Name: name, Data: data}}, options...)
 }
 
 // PublishMultiple publishes multiple messages in a batch.
-func (c *RESTChannel) PublishMultiple(ctx context.Context, messages []*Message) error {
-	return c.PublishMultipleWithOptions(ctx, messages)
+func (c *RESTChannel) PublishMultiple(ctx context.Context, messages []*Message, options ...PublishMultipleOption) error {
+	return c.PublishMultipleWithOptions(ctx, messages, options...)
 }
 
 // PublishMultipleOption is an optional parameter for
@@ -76,7 +66,15 @@ func (c *RESTChannel) PublishMultiple(ctx context.Context, messages []*Message) 
 type PublishMultipleOption func(*publishMultipleOptions)
 
 type publishMultipleOptions struct {
-	params map[string]string
+	connectionKey string
+	params        map[string]string
+}
+
+// PublishWithConnectionKey allows a message to be published for a specified connectionKey.
+func PublishWithConnectionKey(connectionKey string) PublishMultipleOption {
+	return func(options *publishMultipleOptions) {
+		options.connectionKey = connectionKey
+	}
 }
 
 // Params adds query parameters to the resulting HTTP request to the REST API.
@@ -139,6 +137,13 @@ func (c *RESTChannel) PublishMultipleWithOptions(ctx context.Context, messages [
 		}
 		query = "?" + queryParams.Encode()
 	}
+
+	if connectionKey := publishOpts.connectionKey; connectionKey != "" {
+		for _, msg := range messages {
+			msg.ConnectionKey = connectionKey
+		}
+	}
+
 	res, err := c.client.post(ctx, c.baseURL+"/messages"+query, messages, nil)
 	if err != nil {
 		return err
--- a/ably/rest_channel.go
+++ b/ably/rest_channel.go
@@ -53,9 +53,16 @@ func newRESTChannel(name string, client *REST) *RESTChannel {
 
 // Publish publishes a message on the channel.
 func (c *RESTChannel) Publish(ctx context.Context, name string, data interface{}) error {
-	msg := data.(Message)
+	//Check to see if data received is of type message.
+	if msg, ok := data.(Message); ok {
+		// Publish the message with connection key.
+		return c.PublishMultiple(ctx, []*Message{
+			{Name: name, Data: data, ConnectionKey: msg.ConnectionKey},
+		})
+	}
+	// Only the name and data are published.
 	return c.PublishMultiple(ctx, []*Message{
-		{Name: name, Data: data, ConnectionKey: msg.ConnectionKey},
+		{Name: name, Data: data},
 	})
 }
 
--- a/ably/rest_channel.go
+++ b/ably/rest_channel.go
@@ -53,8 +53,9 @@ func newRESTChannel(name string, client *REST) *RESTChannel {
 
 // Publish publishes a message on the channel.
 func (c *RESTChannel) Publish(ctx context.Context, name string, data interface{}) error {
+	msg := data.(Message)
 	return c.PublishMultiple(ctx, []*Message{
-		{Name: name, Data: data},
+		{Name: name, Data: data, ConnectionKey: msg.ConnectionKey},
 	})
 }
 


Rosalita:
I've converted this PR to draft as I don't like that the tests are saying:

panic: interface conversion: interface {} is string, not ably.Message [recovered]
	panic: interface conversion: interface {} is string, not ably.Message
	
Going to take a look into this.

ably/rest_channel.go
```diff
@@ -53,8 +53,9 @@ func newRESTChannel(name string, client *REST) *RESTChannel {
 
 // Publish publishes a message on the channel.
 func (c *RESTChannel) Publish(ctx context.Context, name string, data interface{}) error {
+	msg := data.(Message)
```
sacOO7, line 56: You need to add optional type assertion, current one is forced. Instead, you can do
```
msg, ok := data.(Message)
if (ok) {
connectionKey := data.connectionKey
}
```

sacOO7: `data` is always not of type `Message`, it can be a string too

Rosalita: Thank you @sacOO7 I have addressed your comment.

sacOO7:
@Rosalita Can we add tests for four different scenarios and the bug we were testing before? I think it would be really helpful to understand the fix too.

ably/rest_channel.go
```diff
@@ -53,6 +53,14 @@ func newRESTChannel(name string, client *REST) *RESTChannel {
 
 // Publish publishes a message on the channel.
 func (c *RESTChannel) Publish(ctx context.Context, name string, data interface{}) error {
+	//Check to see if data received is of type message.
+	if msg, ok := data.(Message); ok {
```
sacOO7, line 57: Actually, I am skeptical about `Message` type being used as a public struct. My understanding for the same is, it's supposed to hold message data. Other metadata is modified internally, like `timestamp`, `connectionId`, `clientID` etc.
So, we shouldn't be passing it from outside. Also `data interface {}` as a struct can hold any type of data, adding `connectionKey` to the struct object shouldn't force to use `Message` type. Also, it feels redundant to use `connectionKey` as a part of an actual data message. My personal opinion would be to add extra param to Publish. Since, go doesn't support optional params, as per suggestions by go devs, a new method can be added like this
`PublishWithConnectionKey(ctx context.Context, name string, data interface{}, connectionKey string)` or 
`PublishOnBehalfOf(ctx context.Context, name string, data interface{}, connectionKey string)`. The second name seems more verbose and easy to understand in the public interface itself, instead of making things confusing by including `connectionKey` inside `data` ( user needs to look at documentation explicitly). 

The only issue I see with this is, it's not defined as a public interface in the spec. Though, go language limits reuse of same interface method due to lack of optional param support.

sacOO7: @lmars I would like your input for the same, also you can point out if I am missing something.

Rosalita:
I see two outstanding queries for this PR.

1) Tests - Sachin has asked that tests are added. The following should be considered. There is only 1 scenario which applies to the code changed by this PR. This scenario is:

- Given that Client A is realtime with echo messages set to false.
- And Client B is a rest client that knows the connection key of client A.
- And all clients have performed Get on the same channel name.
- When Client B publishes a message to the channel and provides the connection key of client A e.g. client B is publishing on behalf of client A
- Then Client A does not receive the message as echo messages is set to false and the message is originating from a source with the same connection key as client A. 

This kind of test would be higher up the testing pyramid than a simple unit test as it involves keeping a realtime client and a rest client synchronised with each other. This kind of test would need to exist at a layer higher in the testing pyramid than unit tests.  I've seen these kind of tests before called either end to end tests or an integration tests. At this layer, tests come with a higher cost in terms of both implementation and maintenance than regular unit tests. These kind of tests also have a tendency to be less stable than unit tests as they have to stay synchronised with each other. 
Which raises the following questions:

1a. Do we have any existing tests that test at end to end or integration layer?
1b. If we do have these kind of tests, how stable are they?

We should be pragmatic about not making the fragility of the existing test suite worse. e.g. adding new tests which sometimes appear to fail at random.

2) Sachin has also commented about implementation detail and asked Lewis for his input. He suggests making a new method called publishWithConnectionKey.

I have a question about this. My question is because there are multiple fields that can be present on a message e.g. `ClientID`, `ConnectionID`, `ConnectionKey`, `Timestamp`. Would this path of creating a method called `publishWithConnectionKey` lead us to creating a second method called `publishWithTimestamp` and a third called `publishWithConnectionKeyAndTimestamp`, or even `publishWithConnectionKeyAndTimestampAndClientID`. It feels that working from the message struct, which is available to be unmarshalled, instead of creating lots of methods is a good approach here. However, I am new to this code base and may be missing the reason why this proposed implementation change to create a new method is a great idea. @sacOO7 do you have a working example of your proposed implementation change that fixes this issue? 

We should also await comments from @lmars

sacOO7:
I would like to quickly answer some of those questions 
1a. Do we have any existing tests that test at end to end or integration layer? 
Most of the tests we have are integration tests using ably sandbox API (https://ably.com/documentation/client-lib-development-guide/test-api). We are not using any mocking framework for go, hence few unit tests. It seems we are following the testing trophy (where integration tests are more than unit tests) instead of the testing pyramid.

Rosalita:
@lmars Would you be able to take a look at this PR today? 

Regarding the two comments made by @sacOO7 

1. I'm currently looking into the tests on the project to see if I can find a way to sympathetically build on top of the existing tests without making test instability even worse than it currently is. It is acknowledged that the current test model has drifted from the ideal "pyramid" model.

2. Regarding the comment around implementation details. This comment doesn't contain any points I can immediately action on this PR. The conversation around the fact that a struct called `Message` is public. Feels like addressing this would be a change to the existing architecture of the SDK, which is unrelated to this bug fix. e.g. the architecture was already like this before attempting to fix the bug. 

This PR is only to fix a reported bug. It should not be blocked by requests for architectural changes. If architectural changes are needed to address any existing technical debt, these should be on a new ticket. The advantage of doing this is that any refactoring which needs to be done to correct technical debt can be done without making any changes to functionality. Thus keeping PRs smaller and easier for reviewers to review.

lmars:
Sorry for the delay.

To start with some context, we did some upfront design on what the public facing API should look like in https://github.com/ably/ably-go/issues/197. As part of that, we decided to adopt functional options (see the "Options" section).

So it seems here, we should have a `PublishWithOptions` method on `RESTChannel`, and there should be a `PublishWithConnectionKey` option for publishing a message with an explicit connection key.

Calling that would look like this:

```go
channel.PublishWithOptions(
	"foo",
	"bar",
	ably.PublishWithConnectionKey("connectionKey"),
)
```

On the testing front, ideally we'd have a test that checks that the above code results in a message being sent with a populated `connectionKey`, and ideally that would be done by having a mock server that we publish to and then assert what it received. Testing that the realtime client which that connection key came from doesn't actually receive the message is more a test of the Ably system itself, not something that needs to be tested in every SDK (and indeed, such a test already exists in our internal realtime test suite).

I don't think we currently have anything in place to do that kind of mocked server testing, and so given our current priorities, I'd accept a manual check that the change fixes the bug rather than a test being added to this PR, and focus put into improving the test suite separately.

lmars:
I just looked and we do already have [`RESTChannel.PublishMultipleWithOptions`](https://pkg.go.dev/github.com/ably/ably-go/ably#RESTChannel.PublishMultipleWithOptions), so I suggest also adding `RESTChannel.PublishWithOptions`, and consider deprecating [`PublishMultipleOption`](https://pkg.go.dev/github.com/ably/ably-go/ably#PublishMultipleOption) in favour of just `PublishOption` (the options are the same whether you're publishing one message or multiple messages).

sacOO7:
@lmars really thanks for the input! Though I would suggest adding a test that covers atleast one scenario. Otherwise, it will be difficult to know if it's still working in the future if there are changes to `Publish` API. Also, https://github.com/ably/ably-go/pull/437 didn't contain any tests, so we had to manually test if it was working. It will also be helpful for future go devs. working on this repo. to understand the full context of the bug.

sacOO7:
Also, it will be great if we can update `README` based on changes in public interfacing API.

ably/rest_channel.go
Rosalita: I have reworked the implementation of this bug fix to use a functional option to provide the connection key.

Rosalita:
Regarding 1) This ticket has been created to add a test for a REST client publishing on behalf of a Realtime client. https://github.com/ably/ably-go/issues/448

Regarding 2) The implementation of this bug fix has been reworked to use functional options.

Rosalita:
> Also, it will be great if we can update `README` based on changes in public interfacing API.

This has also been done.

Rosalita:
@lmars @sacOO7 this is now ready for review again :)

README.md
```diff
@@ -224,6 +224,13 @@ err = channel.PublishMultiple(ctx, []*ably.Message{
 if err != nil {
         panic(err)
 }
+
+// A REST client can publish messages on behalf of a realtime client
+// by providing the connection key of the realtime client.
+err := channel.Publish(ctx, "temperature", "12.7", ably.PublishWithConnectionKey("108yZeOBAB7iva!QQJ2CnCSa6A4ql22-a7443108yZeOBAB7iva"))
```
sacOO7, line 230: I am not sure if we usually put secrets in the README, just a suggestion - 
`PublishWithConnectionKey("ConnectionKeyOfPublisherClientHere"));`

ably/rest_channel.go
```diff
@@ -52,23 +52,29 @@ func newRESTChannel(name string, client *REST) *RESTChannel {
...
-		{Name: name, Data: data},
-	})
+func (c *RESTChannel) Publish(ctx context.Context, name string, data interface{}, options ...PublishMultipleOption) error {
+	return c.PublishMultiple(ctx, []*Message{{Name: name, Data: data}}, options...)
 }
 
 // PublishMultiple publishes multiple messages in a batch.
-func (c *RESTChannel) PublishMultiple(ctx context.Context, messages []*Message) error {
-	return c.PublishMultipleWithOptions(ctx, messages)
+func (c *RESTChannel) PublishMultiple(ctx context.Context, messages []*Message, options ...PublishMultipleOption) error {
```
sacOO7, line 60: There doesn't seem to be any difference in the method signature of ` PublishMultiple` and `PublishMultipleWithOptions`. 
I don't think we need to change code at `60-61`. Instead add one more method `PublishWithOptions`. So, we will have 4 methods in total.
```
1. Publish -> Calls PublishMultiple
2. PublishWithOptions -> Calls PublishMultipleWithOptions
3. PublishMultiple -> Calls PublishMultipleWithOptions with no options 
4. PublishMultipleWithOptions -> Has real business logic as is, no need to change it
```

README.md
Rosalita: This is just an example of realistic data, but I have no strong opinion and happy to make this change in order to get this PR through review. I've just pushed this change for you.

ably/rest_channel.go
sacOO7: If we don't want to have `WithOptions` methods, we can just keep **`first`** and **`third`**, accepting optional params at the end.

README.md
sacOO7: Thanks !!

ably/rest_channel.go
lmars: If we do decide to drop the `PublishMultipleWithOptions` function (which I agree will now be redundant), we need to deprecate it and advise what the alternative is, rather than removing it straight away.

sacOO7: Yeah, that's what I meant. Since it's a public interface, we will deprecate it for the time being. Thanks for the reminder @lmars.

Rosalita: If you are happy @sacOO7 please can you resolve this conversation.

ably/rest_channel.go
Rosalita: @sacOO7 please can you resolve this conversation if you are happy that this code has been re-worked to use functional options .

ably/rest_channel.go
sacOO7: No, I think we need to make changes. Can you go through the conversation again?

Rosalita: What are the changes you think we need to make here?

sacOO7: I suggest having four methods as mentioned. It's more explicit and verbose for SDK users.

Rosalita: As this is a bug fix for an issue a customer is facing, I would advocate for making the smallest amount of change possible to resolve the issue.

QuintinWillison (approved):
Thanks. I can see that this moves the codebase forwards, clearly representing an improvement.

@sacOO7 Thanks for your inputs on this PR, however we're keen to get this landed so we can be responsive to a customer who is waiting. For that reason, if you feel there are outstanding conversations to be had then could you please start issue(s) for them. Thanks!

@Rosalita Please go ahead and get this merged. 🎉

ably/rest_channel.go
QuintinWillison: Resolving this conversation on @sacOO7's behalf so that @Rosalita can land this PR.

ably/rest_channel.go
QuintinWillison: Resolving this conversation on @sacOO7's behalf so that @Rosalita can land this PR.

ably/rest_channel.go
lmars: I did the deprecation in https://github.com/ably/ably-go/pull/450.

PR #453, opened 2022-02-14
Remove useless switch

Morganamilo:
This switch just matched the presence event and for each case set a new
varable to whatever the presence event was. There's no need to swtich as
the new varable will always equal the original.


Rosalita (approved):
A very logical change, happy to approve.

Peter-Maguire (approved):

PR #455, opened 2022-02-14
Fix hang when joining existing channel with no presence

Morganamilo:
When getting a channel the sync mutex would be unconditionally locked to
wait for sync. But sync's are only sent if the channel is new or has
presence.

This means the mutex would remain locked forever waiting for a sync that
isn't coming.

### DIFF
--- a/ably/realtime_presence.go
+++ b/ably/realtime_presence.go
@@ -96,7 +96,7 @@ func (pres *RealtimePresence) onAttach(msg *protocolMessage) {
 	pres.mtx.Lock()
 	defer pres.mtx.Unlock()
 	switch {
-	case msg.Flags.Has(flagHasPresence) || serial != "":
+	case msg.Flags.Has(flagHasPresence):
 		pres.syncStart(serial)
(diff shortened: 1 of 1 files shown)


Morganamilo:
Hey so I'm not sure if this PR is *right*.

This check was explicitly added back in 558854b2 (2015) where it said:

>The reason was buggy sync check condition which did not cover scenario
where sync flag was missing but sync serial present.

However looking at the spec I can't see any mention of a sync serial being sent in an attatched message. The spec only seems to talk about the sync bit.

As far as I can tell though there's always a serial sent even when no sync is coming.

So anyway, looks like the person who wrote this originally is no longer at ably. Some insight from some one who's been here longer would be nice.

lmars (approved):
I've read the description, and I unfortunately can't provide any extra context here, but I think the change seems reasonable 👍

Rosalita (approved):

stmoreau:
I am sure we have considered/discussed that, but I am curious. Would it be worth adding some unit tests (even if they're written from scratch) before merging this change?

PR #460, opened 2022-02-18
Add detailed description of project tests to readme.md

### LINKED ISSUE #458: Test improvements - Move integration tests to their own package.
Related to #447 

Now we have clear separation between unit tests and integration tests. We should consider moving the integration tests to their own package.

We could make a package called `integration_test` and move the the integration tests there. 

If we used a `TestMain` function inside that package, we could add test set up and teardown there. Test setup could set the environment variables that the integration tests require. This would allow us to delete the bash script that is currently used to run the test.

An example of using TestMain: 

```
func TestMain(m *testing.M) {

	flag.Parse()
	// Test setup

	// Run tests
	exitCode := m.Run()

	// Test teardown
	os.Exit(exitCode)
}
```

The advantages of moving integration tests to their own package:

* There would be fewer test files in the root of the project making it easier to work with tests.
* It would also improve developer experience as they could be run using the Go tool chain ( `go test ./integration_test` instead of having to run via a bash script)
* We would no longer need to use build tags at the top of each test file to separate unit tests and integration tests.

Acceptance criteria:

* Integration tests are moved to a new package.
* Integration tests can be run in CI without the need for a bash script. *(stretch goal)*
* Build tags are removed from test files.
* CI pipeline steps are updated so they no longer rely on build tags.
* Project readme file is updated to explain how to run all tests.

┆sprint: SDK 2022-Q1-S4

Rosalita:
As it is not possible to move the integration tests to their own package. Have updated the readme file with a detailed description of how tests work on this project. This resolves #458


Peter-Maguire (approved):

README.md
```diff
@@ -338,10 +338,27 @@ As of release 1.2.0, the following are not implemented and will be covered in fu
 
 ### Tests
 
-This project contains two types of test. Unit tests and integration tests. The integration tests depend on a git submodule being present. To clone the repo with submodules `git clone git@github.com:ably/ably-go.git --recurse-submodules` 
+This project contains two types of test. Test which use the `ablytest` package and tests which dont. 
 
-- Unit tests can be run with `go test -tags=unit ./...`
-- Integration tests can be run using `scripts/test.sh`.
+#### Unit tests
```
Morganamilo, line 343: Headings should be followed by a blank line (in my opinion anyway)

README.md
```diff
@@ -338,10 +338,27 @@ As of release 1.2.0, the following are not implemented and will be covered in fu
...
+
+Please note that these tests are not true integration tests as rather than using the public API, they rely on `export_test.go` to expose private functionality so that it can be tested. These tests exist in files which are suffixed `_integration_test.go`. They run twice in the CI pipeline at the steps `Integration Tests with JSON Protocol` and `Integration Tests with MessagePack Protocol`. To run these tests locally, they have a dependency on a git submodule being present, so it is necessary clone the project with `git clone git@github.com:ably/ably-go.git --recurse-submodules`. A bash script is used to run these tests with the command `scripts/test.sh --protocol application/json` or `scripts/test.sh --protocol application/x-msgpack` depending on which protocol they are to be run for. It is also necessary to clean the test cache in between runs of these tests which can be done with the command `go clean -testcache`.
```
Morganamilo, line 354: The long commands make this look a bit unwieldy. Maybe they could be blocks instead:

Please note that these tests are not true integration tests as rather than using the public API, they rely on `export_test.go` to expose private functionality so that it can be tested. These tests exist in files which are suffixed `_integration_test.go`. They run twice in the CI pipeline at the steps `Integration Tests with JSON Protocol` and `Integration Tests with MessagePack Protocol`.

To run these tests locally, they have a dependency on a git submodule being present, so it is necessary clone the project with:

```
git clone git@github.com:ably/ably-go.git --recurse-submodules
```
A bash script is used to run these tests with the commands 

```
scripts/test.sh --protocol application/json
scripts/test.sh --protocol application/x-msgpack
```

Depending on which protocol they are to be run for. It is also necessary to clean the test cache in between runs of these tests which can be done with the command:

```
go clean -testcache
```

README.md
Rosalita: Thank you @Morganamilo blank line is now added.

README.md
Rosalita: Thankyou @Morganamilo I have just pushed these changes :)

Morganamilo (approved):

Issue #456, opened 2022-02-15
Test improvements - Standardise test assertions


Rosalita:
Currently this project uses two different ways of asserting in its tests. 

In some places like `error_test.go` the `github.com/stretchr/testify/assert` package is used to assert values are equal with `assert.Equal`, values are true with `assert.True` etc.

This project also appears to have 'rolled its own' assertion code in `ably_test.go` where there are functions named `assertEquals`, `assertTrue` etc. Our own assertion code in `ably_test.go` is not covered by any unit tests.

We should standardise our approach to making assertions in tests. As we already use the `/testify/assert` package, which has 15.5k stars on GitHub, we should not have to reinvent the wheel and also write our own assertion code. 

The advantages of using a widely adopted assertion package like `/testify/assert` are that this package is already covered by many of unit tests, is actively maintained and updated. It's API has been well designed and thought out making tests that use it highly readable.

I would propose we move from partial adoption of `/testify/assert` and fully adopt this package. This would allow us to delete the assertion code in `ably_test.go`.

Rosalita:
Internal discussion raised the following point:
- We definitely want to be consistent in our approach to assertions. @lmars and @Rosalita both believe we should fully adopt `testify/assert`
- There was a concern from @sacOO7  that adding `testify/assert` as a dependency would create version conflicts for downstream users, however it was noted that `testify/assert` already is already a dependency of this project. So if conflict was going to happen, it's already happened.
- The version of `testify/assert` this project currently uses is `1.4.0`. `testify/assert` accidentally released a breaking change in `1.6.0`, which was reverted in `1.6.1`.  So given there are no breaking changes between `1.4.0` and `1.7.0` there should be nothing to stop this dependency being updated to `1.7.1` in this project.

It would be great to get feedback from @QuintinWillison on this proposed change when he returns from annual leave.

PR #465, opened 2022-02-25
Part 1 of standardising test assertions

Rosalita:
- Test assertions which were using the assertions in `ably_test.go` now use package `testify/assert`
- Have removed the assertion code from `ably_test.go` that duplicates functionality in package `testify/assert`
- Converted all test assertions in `auth_integration_test.go` to use `testify/assert`

Please note, not all test assertions are standardised yet, this PR is only part 1 of #456. Taking the approach of breaking this task down into parts so PRs are smaller and easier for reviewers to review.

### DIFF
--- a/ably/auth_integration_test.go
+++ b/ably/auth_integration_test.go
@@ -44,11 +44,11 @@ func TestAuth_BasicAuth(t *testing.T) {
 	defer safeclose(t, app)
 
 	_, err := client.Time(context.Background())
-	assert.Nil(t, err,
+	assert.NoError(t, err,
 		"Expected client.Time to return a nil error, got %v", err)
 
 	_, err = client.Stats().Pages(context.Background())
-	assert.Nil(t, err,
+	assert.NoError(t, err,
 		"Expected client.Stats().Pages to return a nil error, got %v", err)
 	assert.Equal(t, 2, rec.Len(),
 		"Expected rec.Len to return 2, got %d", rec.Len())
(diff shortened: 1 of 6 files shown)


ably/auth_integration_test.go
```diff
@@ -124,51 +113,44 @@ func TestAuth_TokenAuth(t *testing.T) {
...
 
 	beforeAuth := time.Now().Add(-time.Second)
-	if _, err := client.Time(context.Background()); err != nil {
-		t.Fatalf("client.Time()=%v", err)
-	}
-	if _, err := client.Stats().Pages(context.Background()); err != nil {
-		t.Fatalf("client.Stats()=%v", err)
-	}
+	_, err := client.Time(context.Background())
+	assert.Nil(t, err,
```
Peter-Maguire, line 117: I think it should be assert.NoError when checking an error, functionally the same but describes what is being checked a little better

ably/auth_integration_test.go
```diff
@@ -182,57 +164,48 @@ func TestAuth_TokenAuth_Renew(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("NewREST()=%v", err)
-	}
-	if _, err := client.Stats().Pages(context.Background()); err == nil {
-		t.Fatal("want err!=nil")
-	}
+	assert.Nil(t, err,
+		"NewREST()=%v", err)
+	_, err = client.Stats().Pages(context.Background())
+	assert.NotNil(t, err)
```
Peter-Maguire, line 205: Same as previous but with assert.Error instead

Peter-Maguire (changes requested):

ably/auth_integration_test.go
Rosalita: Thank you @Peter-Maguire I agree and have pushed these changes.

Peter-Maguire (approved):

QuintinWillison (approved):
I've only skim read, but looks sensible. I'm assuming the `assert.Equal` replacements for `assertDeepEquals` are deep in nature (e.g. for an array, equality is checked for each member), but I've not looked that up.

PR #466, opened 2022-02-25
Fix / not being escaped by rest client

### LINKED ISSUE #463: Failure to get channel with slash
However joining a channel works it seems to involve a url built from the channel name. /'s in the name are not escaped and result in creating an invalid url. If you're creating the channel it seems to work.

┆sprint: Backlog

Morganamilo:
Fixes #463

### DIFF
--- a/ably/paginated_result.go
+++ b/ably/paginated_result.go
@@ -17,16 +17,18 @@ const (
 )
 
 type paginatedRequest struct {
-	path   string
-	params url.Values
+	path    string
+	rawPath string
+	params  url.Values
 
 	query queryFunc
 }
 
-func (r *REST) newPaginatedRequest(path string, params url.Values) paginatedRequest {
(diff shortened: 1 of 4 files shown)


Morganamilo:
Whoops missed that function is also called from other files. Need to fix those too.

Morganamilo:
Also I'm not a fan of how this works exactly. Go's url auto escapes urls. If we tell it not to we have to also provide an unescaped version to the url for go to security check? for us. So this ends up with us passing the same url twice basically and it's a bit annoying.

Rosalita:
I've been able to recreate this issue and understand how the fix works. I agree it's annoying having to pass both the path and the rawPath but I think this is the right solution here. I'm happy to give my approval 👍

Rosalita (approved):

lmars (approved):

Peter-Maguire (approved):

PR #473, opened 2022-03-15
Tests/standardise assertions pt4

Rosalita:
More work completed towards #456

### DIFF
--- a/ably/realtime_client_integration_test.go
+++ b/ably/realtime_client_integration_test.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/ably/ably-go/ably"
 	"github.com/ably/ably-go/ablytest"
+
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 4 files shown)


Morganamilo:
I see and have been seeing so many work in progress commits. I'd generally consider that bad and messy. Is there a reason it's not just one commit? Or multiple commits with better tittles?

Rosalita:
> I see and have been seeing so many work in progress commits. I'd generally consider that bad and messy. Is there a reason it's not just one commit? Or multiple commits with better tittles?

It's usually because I have had to change tasks and want to save my progress. Would you like me to rebase and squash them all into just 1 commit to reduce the noise in the commit history?

mohyour:
> 

You can create a commit before switching, then update that commit when you're back on the task.

Morganamilo:
> It's usually because I have had to change tasks and want to save my progress. Would you like me to rebase and squash them all into just 1 commit to reduce the noise in the commit history?

that's what i always do.  `git rebase -i main`

Morganamilo:
For wip commits i just --ammend over it.

Rosalita:
I've squashed everything into 1 commit.

Morganamilo (approved):

mohyour (approved):

PR #478, opened 2022-03-22
Add missing pagination methods

### LINKED ISSUE #120: HTTPPaginatedResult and PaginatedResult are missing some methods
It appears https://github.com/ably/ably-go/blob/develop/ably/http_paginated_response.go is missing various methods, such as `IsLast`, `HasNext` and `First`. They should have these methods to match the other libraries.

Once added, we need to also re-add these methods to the docs. Currently they are commented out with the indicator `GO_REPLACEMENT` to easily find and uncomment them.

┆sprint: SDK 2022-Q1-S6

mohyour:
Fixes #120 
First doesn't appear to be missing

### DIFF
--- a/ablytest/pagination.go
+++ b/ablytest/pagination.go
@@ -73,7 +73,11 @@ func testPagination(request reflect.Value, expectedItems []interface{}, perPage 
 		var gotItems []interface{}
 		pageNum := 1
 		expectedFullPages := len(expectedItems) / perPage
+		pageMap := make(map[int]map[string]bool)
 		for pages.next() {
+			pageMap[pageNum] = map[string]bool{}
+			pageMap[pageNum]["hasNext"] = pages.hasNext()
+			pageMap[pageNum]["isLast"] = pages.isLast()
 			if (pageNum <= expectedFullPages && len(pages.items()) != perPage) ||
 				(pageNum > expectedFullPages && len(pages.items()) >= perPage) {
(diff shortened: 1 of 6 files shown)


Rosalita:
I was just reading the package level documentation for paginated results at https://pkg.go.dev/github.com/ably/ably-go/ably#hdr-Paginated_results

I don't know why, but there's an example there which looks like it's a `TODO` place holder. It's in a file called `doc_test.go` which has been there for over a year.

Given that you have been working with pagination, it would be really nice to add a code example of how a user can use the `PaginatedResults` object and including the new methods you have added (and remove the old `TODO` example).  

Do you think we should create a new ticket for adding this documentation example?

mohyour:
> 
I think a new ticket is better. Once this is approved, then I can add the code examples and update the docs repository.

ablytest/pagination.go
```diff
@@ -73,14 +73,32 @@ func testPagination(request reflect.Value, expectedItems []interface{}, perPage
 		var gotItems []interface{}
 		pageNum := 1
 		expectedFullPages := len(expectedItems) / perPage
+		pageMap := make(map[int]map[string]bool)
```
Rosalita, line 76:
```suggestion
		pageMap := make(map[int]map[string]bool, expectedFullPages)
```
As we know how many pages we are expecting, we should probably provide a size when making the page map. This will let the compiler know to create a map which is allocated at least enough memory to hold the number of expected pages before memory is allocated again.

Peter-Maguire (approved):

PR #480, opened 2022-03-25
Update Go version to 1.17 in go.mod, add Go 1.18 to versions tested

### LINKED ISSUE #474: Support Go 1.18
Go 1.18 was officially released yesterday 🎉 

The project readme says
```
## Supported Versions of Go

Whenever a new version of Go is released, Ably adds support for that version. The [Go Release Policy](https://golang.org/doc/devel/release#policy) supports the last two major versions. This SDK follows the same policy of supporting the last two major versions of Go.

```

To support 1.18, we should take the following actions:
- Increase the minimum version of Go required to use the SDK by bumping the version of Go in the go.mod file from 1.16 to 1.17 
- Add Go version 1.18 to the GitHub Workflow so all tests run against Go 1.18
- Remove Go version 1.16 from the GitHub Workflow so tests no longer run against Go 1.16

┆sprint: SDK 2022-Q1-S6

Rosalita:
- Added Go version `1.18` to Github workflow and removed Go version `1.16`
- Updated Go version in `go.mod` to `1.17` and ran `go mod tidy`

Resolves #474


Morganamilo:
I don't see a reason to remove the old build tags right away. It's no hassle to keep both right?

Rosalita:
> I don't see a reason to remove the old build tags right away. It's no hassle to keep both right?

Sorry to say that I don't see a good reason to save them. They are essentially dead code now we have moved on from `1.16` ☠️

Morganamilo:
While we only target the last 2 versions of Go, I don't see a reason to needlessly break things for people on older versions even if they are technically not supported. But I guess it doesn't really matter.

Rosalita:
I just tested creating a realtime client with an older version of Go `1.15.6`. Looks like I was wrong about the go version in `go.mod` being minimum Go version.

On main branch of ably-go, creating a realtime client works for Go `1.15.6` while the `go.mod` contains `1.16`.

On this branch when the old the build tags were removed, Go `1.15.6` failed to create a realtime client, giving an error that states "module requires Go 1.17".

I've restored the old build tags as this will cause less pain for any users on an unsupported version of Go < 1.17

Morganamilo (approved):

PR #485, opened 2022-03-31
Handle errors in Request()

Morganamilo:
Currently errors like 500 response codes are not checked and don't
return an error.

Request is still broken but may as well push this fix as I've moved to other stuff for the time being.


Morganamilo:
I think the tests may actually be failing for real here. Need to look into it.

Rosalita (approved):

PR #489, opened 2022-04-05
Do not persist authorise attributes force & timestamp

### LINKED ISSUE #54: Do not persist authorise attributes force & timestamp
A recent discovery in the iOS library has unearthed a small shortcoming in the spec, corrected at [https://github.com/ably/docs/commit/f768f0db76020d945cf34974ef7d14d36113f0cd](https://github.com/ably/docs/commit/f768f0db76020d945cf34974ef7d14d36113f0cd)

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-151) by [Unito](https://www.unito.io)
┆sprint: SDK 2022-Q1-S8

mohyour:
Fixes #54

### DIFF
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -311,9 +311,10 @@ func (a *Auth) authorize(ctx context.Context, params *TokenParams, opts *authOpt
 	// RSA10j, RSA10g - override existing tokenParams and authOptions, ignore timestamp and queryTime
 	a.opts().TokenDetails = tok
 	a.params = params
+	// Refreshing or renewing a token, opt won't be nil.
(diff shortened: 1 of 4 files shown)


Rosalita:
The test only checks one scenario where token parameters has a timestamp and the AuthWithQueryTime option is true. 

It would be good to expand this test to cover more scenarios e.g. when the token parameters has no timestamp, when AuthWithQueryTime is false, when there are no auth options etc.

Rosalita (approved):

sacOO7 (approved):

PR #492, opened 2022-04-14
Fix Request()

### LINKED ISSUE #486: AblyREST.Request() does not work
It looks like the paging for Request() doesn't actually save the results anywhere. So no matter what data is sent back calling `.Items()` on the results will always turn up empty.

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-1708) by [Unito](https://www.unito.io)
┆sprint: SDK 2022-Q2-S2 14/04

Morganamilo:
Fixes #486

### DIFF
--- a/ably/rest_client.go
+++ b/ably/rest_client.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/ably/ably-go/ably/internal/ablyutil"
+	"github.com/ugorji/go/codec"
 )
 
 var (
@@ -33,6 +34,23 @@ func query(fn func(context.Context, string, interface{}) (*http.Response, error)
 	return func(ctx context.Context, path string) (*http.Response, error) {
 		return fn(ctx, path, nil)
 	}
+}
+
+type Raw []byte
+
+func (r *Raw) UnmarshalJSON(data []byte) error {
+	*r = data
+	return nil
+}
+
+func (r *Raw) CodecEncodeSelf(*codec.Encoder) {
+	panic("Raw cannot be used as encoder")
+}
+
+func (r *Raw) CodecDecodeSelf(decoder *codec.Decoder) {
+	var raw codec.Raw
+	decoder.MustDecode(&raw)
+	*r = []byte(raw)
 }
 
 // RESTChannels provides an API for managing collection of RESTChannel. This is
@@ -417,7 +435,7 @@ func (r RESTRequest) Pages(ctx context.Context) (*HTTPPaginatedResponse, error) 
 // See "Paginated results" section in the package-level documentation.
 type HTTPPaginatedResponse struct {
 	PaginatedResult
-	items jsonRawArray
+	items Raw
 }
 
 func (r *HTTPPaginatedResponse) StatusCode() int {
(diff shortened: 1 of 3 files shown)


ably/rest_client.go
```diff
@@ -35,6 +36,23 @@ func query(fn func(context.Context, string, interface{}) (*http.Response, error)
 	}
 }
 
+type Raw []byte
```
Rosalita, line 39: As an exported type with exported methods, please can we add a comment line above the type to explain what this represents. Also a comment line above each exported method to explain their use. Main reason for adding this is to improve the auto generated docs at [pkg.go.dev](https://pkg.go.dev/github.com/ably/ably-go/ably)

Morganamilo: Actually I don't think this is meant to be exported whoops. Thanks.

ably/rest_client.go
```diff
@@ -35,6 +36,23 @@ func query(fn func(context.Context, string, interface{}) (*http.Response, error)
...
+
+func (r *Raw) UnmarshalJSON(data []byte) error {
```
Rosalita, line 41: This function signature here can return an error, but I can see the code always returns nil. Is there some error handling missing here? Unmarshalling JSON is usually something that can go wrong and generate an error. How would that error be returned?

Morganamilo: The raw type just pretends to decode but returns the original bytes.

https://pkg.go.dev/encoding/json#RawMessage

Will be commented as  I forgot to comment Raw.

Rosalita: I see, so `Raw` has this method so that it can pass through the `Unmarshaler` interface.

ably/rest_client.go
```diff
@@ -35,6 +36,23 @@ func query(fn func(context.Context, string, interface{}) (*http.Response, error)
...
+	*r = data
+	return nil
+}
+
+func (r *Raw) CodecEncodeSelf(*codec.Encoder) {
+	panic("Raw cannot be used as encoder")
```
Rosalita, line 47: I'm not sure the reason to use panic here. Panic will kill execution and cause an exit with a non-zero status, which feels a bit severe. Assuming this method is needed to satisfy an interface, is there a better way to handle the scenario when `CodecEncodeSelf` is called?

Morganamilo: This function will never be called. This is the same panic line taken from `rest_channel.go`

Rosalita: I don't like the `panic` there either, but at least we are being consistent :joy: 

Analysing our use of `panic` in this SDK is probably a bigger task for another day.

ably/rest_client.go
```diff
@@ -470,12 +488,29 @@ func (p *HTTPPaginatedResponse) HasNext(ctx context.Context) bool {
...
+	/// Turn non arrays into arrays so it fits the output kind
+	if typ == "application/json" {
+		token, _ := json.NewDecoder(bytes.NewReader(p.items)).Token()
+		if token != json.Delim('[') {
+			p.items = append(Raw{'['}, p.items...)
+			p.items = append(p.items, ']')
+		}
+	} else if typ == "application/x-msgpack" {
+		if (p.items[0]&0xf0 != 0x90) && p.items[0] != 0xdc && p.items[0] != 0xdd {
+			p.items = append(Raw{0x91}, p.items...)
```
Rosalita: I'm having a hard time reading lines 509 and 510. Maybe some commenting is needed here?

`0x91` looks like a single left quote ( https://bytetool.web.app/en/ascii/code/0x91/ ) I'm not sure how appending this character converts the raw bytes to an array. Please can you talk me through these lines.

Morganamilo: messagepack is a binary format. The nibble 9 is used for the array type, followed by the length. So 91 is an array of length one.

ably/rest_client.go
```diff
@@ -470,12 +490,29 @@ func (p *HTTPPaginatedResponse) HasNext(ctx context.Context) bool {
...
+
+	/// Turn non arrays into arrays so it fits the output kind
+	if typ == "application/json" {
+		token, _ := json.NewDecoder(bytes.NewReader(p.items)).Token()
+		if token != json.Delim('[') {
+			p.items = append(raw{'['}, p.items...)
+			p.items = append(p.items, ']')
+		}
+	} else if typ == "application/x-msgpack" {
+		if (p.items[0]&0xf0 != 0x90) && p.items[0] != 0xdc && p.items[0] != 0xdd {
```
Rosalita, line 511: Needs a comment for this if statement line, something like 
"Check the first byte to see if it's one of message packs three array types."

lmars (approved):

Issue #479, opened 2022-03-24
Once test observability is in place try to turn all tests back on.


Rosalita:
A number of integration tests on ably-go have been turned off due to either failing or being flaky.  Some of these tests were turned off quite a while ago. It's been observed that some of these tests work when run locally.

We should try turn all the tests back on so that we can collect data on which tests are problematic.

Some of the tests might fail 100% of the time. If that is the case, we should probably turn these tests off again and raise a ticket to investigate.

Some of the tests might fail some of the time. If that is the case then the observability dash board should show us the worst tests for flakiness. We can then raise tickets to address these tests.

- We should also consider closing any outstanding PRs that turn tests off.
- We should also consider doing an audit of existing github issues for failing/flaky tests so that we don't create duplicate issues.

┆Issue is synchronized with this [Jira Research](https://ably.atlassian.net/browse/SDK-1693) by [Unito](https://www.unito.io)
┆sprint: SDK 2022-Q2-S3 28/04

Rosalita:
Closing issue as resolved by https://github.com/ably/ably-go/pull/493

Issue #438, opened 2022-01-28
Investigate re-enabling flaky tests.


Rosalita:
The following tests have been identified as flakey.

```
Test_RTN4a_ConnectionEventForStateChange
TestAuth_BasicAuth
TestAuth_ClientID
```

Recently a PR was raised to this code base, out of 4 test runs for that PR the following was observed.

Attempt #1 all tests failed to start - this run was ignored.

Attempt #2 the follow test failed on Go 1.16

```
=== RUN   Test_RTN4a_ConnectionEventForStateChange
=== RUN   Test_RTN4a_ConnectionEventForStateChange/on_CONNECTING
    ably_test.go:66: safeclose 1: failed to close *ablytest.Sandbox: Delete "https://sandbox-rest.ably.io/apps/_tmp_aRmoog": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
=== RUN   Test_RTN4a_ConnectionEventForStateChange/on_CONNECTED
    realtime_conn_spec_test.go:148: timed out waiting for channel receive
    ably_test.go:66: safeclose 0: failed to close ablytest.realtimeIOCloser: context deadline exceeded
=== RUN   Test_RTN4a_ConnectionEventForStateChange/on_DISCONNECTED
--- FAIL: Test_RTN4a_ConnectionEventForStateChange (182.20s)
    --- PASS: Test_RTN4a_ConnectionEventForStateChange/on_CONNECTING (61.86s)
    --- FAIL: Test_RTN4a_ConnectionEventForStateChange/on_CONNECTED (68.89s)
    --- FAIL: Test_RTN4a_ConnectionEventForStateChange/on_DISCONNECTED (51.44s)
panic: request error: Service Unavailable ("{\"message\":\"Ably's routing layer was unable to service this request\",\"code\":50305,\"statusCode\":503,\"href\":\"https://help.ably.io/error/50000\"}") [recovered]
	panic: request error: Service Unavailable ("{\"message\":\"Ably's routing layer was unable to service this request\",\"code\":50305,\"statusCode\":503,\"href\":\"https://help.ably.io/error/50000\"}")
```

And the following test failed on Go 1.17

```
=== RUN   TestAuth_BasicAuth
--- FAIL: TestAuth_BasicAuth (0.53s)
[...]

PR #493, opened 2022-04-19
Turn flaky tests on  and fix failing tests

mohyour:
#438
#479

### DIFF
--- a/ably/rest_presence_spec_integration_test.go
+++ b/ably/rest_presence_spec_integration_test.go
@@ -19,21 +19,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-/*
-FAILING TEST
-https://github.com/ably/ably-go/pull/383/checks?check_run_id=3489733889#step:7:711
-
-=== RUN   TestPresenceHistory_RSP4_RSP4b3/limit=20
---- FAIL: TestPresenceHistory_RSP4_RSP4b3 (95.01s)
-    --- FAIL: TestPresenceHistory_RSP4_RSP4b3/limit=2 (42.06s)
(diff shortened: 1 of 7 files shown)


ably/rest_presence_spec_integration_test.go
```diff
@@ -315,3 +289,28 @@ func sortPresenceByClientID(items []interface{}) {
 		return items[i].(*ably.PresenceMessage).ClientID < items[j].(*ably.PresenceMessage).ClientID
 	})
 }
+
+func sortData(data string) string {
```
Rosalita, line 293: `sortData` appears to be a helper function for the integration tests. I have a few comments to make specifically about this function.

1. Firstly, its name is unclear to me. It says it's sorting data, but I can't see how it is sorting data. It's definitely transforming data, but to me sorting is when the order of more than 1 piece of data may or may not change. This function only receives one piece of data so how is it being sorted? Maybe there's a better name for this function, one that's based on it's intent.

2. Reading the logic in this function, the intent is not obviously clear to me, there are also no comments to guide the reader. Are we trying to encode a string to base64 here? What is the purpose of this helper?

3. For any complex logic, we should have an internal unit test. I would like to see a unit test around this helper function that shows some example inputs and outputs. At the moment, when I feed strings into this function I get strings back out. e.g for an input of "100" the output prints as "e" but the output also has a length of 8. (https://go.dev/play/p/fao4XcVOBli) 

Feeding a variety of inputs in sometimes null characters come back, is this intentional?

mohyour: The function is no longer needed and has been removed

Rosalita (approved):

PR #495, opened 2022-04-20
Enable and fix some skipped tests

Morganamilo:


ably/realtime_conn_spec_integration_test.go
```diff
@@ -2707,9 +2624,18 @@ func TestRealtimeConn_RTN19b(t *testing.T) {
...
 	}
-	for _, expect := range msgs {
+	var gots []*ably.ProtocolMessage
+	for range msgs {
 		var got *ably.ProtocolMessage
 		ablytest.Instantly.Recv(t, &got, out, t.Fatalf)
+		gots = append(gots, got)
+	}
+
+	sort.Slice(msgs, func(i, j int) bool { return  msgs[i].Action < msgs[j].Action})
```
Rosalita, line 2634:
```suggestion

```
I'm not a fan of changing expected values as its usually a code smell in unit tests. In this case however I don't think this line is needed at all because `msgs` is hardcoded in a way that is already sorted by action order. Suggest deleting this line.

ably/realtime_conn_spec_integration_test.go
```diff
@@ -2731,9 +2657,19 @@ func TestRealtimeConn_RTN19b(t *testing.T) {
...
-	for _, expect := range msgs {
+
+	gots = nil
+	for range msgs {
 		var got *ably.ProtocolMessage
 		ablytest.Instantly.Recv(t, &got, out, t.Fatalf)
+		gots = append(gots, got)
+	}
+
+	sort.Slice(msgs, func(i, j int) bool { return  msgs[i].Action < msgs[j].Action})
```
Rosalita, line 2668:
```suggestion

```
Again I don't think we need to sort something thats already hardcoded in a sorted order.

Rosalita (approved):

PR #499, opened 2022-05-03
Fix sync250

Morganamilo:
The loop would wait for 250 messges but it's possable one client may
send a present message as well as an enter. So instead wait for messages
from 250 clients.

### DIFF
--- a/ably/realtime_presence_integration_test.go
+++ b/ably/realtime_presence_integration_test.go
@@ -57,13 +57,7 @@ func TestRealtimePresence_Sync(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-/*
-FAILING TEST
-Go 1.15, 1 Sep 2021: https://github.com/ably/ably-go/pull/383/checks?check_run_id=3485574655#step:9:763
-*/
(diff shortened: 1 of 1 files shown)


ably/realtime_presence_integration_test.go
```diff
@@ -97,19 +91,28 @@ func TestRealtimePresence_Sync250(t *testing.T) {
...
 	}
 
 	err = contains(members2, clients...)
+
+	if err != nil {
+		for k, v := range members2 {
+			fmt.Println(k, v)
+		}
+	}
+
```
Rosalita, line 115: I can see `members2` is printed on line 116 as part of the assertion if this test fails. Is this a debug print that was left in and should be taken out?

Morganamilo: Yes probably. I wrote this before the summit and forgot about it.

Rosalita: No worries, please delete the first print of `members2` and I'm happy to approve 👍

Rosalita (approved):
👍

mohyour (approved):

mohyour:
Fixes #411

PR #500, opened 2022-05-03
fix failing test

mohyour:
Fix `TestRealtimeConn_BreakConnLoopOnInactiveState`
Unskip `TestRealtime_DontCrashOnCloseWhenEchoOff`

### DIFF
--- a/ably/realtime_conn_integration_test.go
+++ b/ably/realtime_conn_integration_test.go
@@ -204,10 +204,8 @@ https://github.com/ably/ably-go/pull/383/checks?check_run_id=3489733889#step:7:5
     --- PASS: TestRealtimeConn_BreakConnLoopOnInactiveState/closed (6.32s)
 */
 func TestRealtimeConn_BreakConnLoopOnInactiveState(t *testing.T) {
-	t.Skip("FAILING TEST")
 
 	for _, action := range []ably.ProtoAction{
-		ably.ActionDisconnect,
 		ably.ActionError,
 		ably.ActionClosed,
 	} {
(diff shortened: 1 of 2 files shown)


ably/realtime_conn_integration_test.go
```diff
@@ -191,23 +191,9 @@ func TestRealtimeConn_ReceiveTimeout(t *testing.T) {
...
-        realtime_conn_test.go:222: called Receive again; expected end of connection loop
-        ably_test.go:66: safeclose 0: failed to close ablytest.realtimeIOCloser: context deadline exceeded
-    --- PASS: TestRealtimeConn_BreakConnLoopOnInactiveState/error (6.07s)
-    --- PASS: TestRealtimeConn_BreakConnLoopOnInactiveState/closed (6.32s)
-*/
 func TestRealtimeConn_BreakConnLoopOnInactiveState(t *testing.T) {
-	t.Skip("FAILING TEST")
 
 	for _, action := range []ably.ProtoAction{
-		ably.ActionDisconnect,
```
Rosalita, line 210: What is the reason to stop testing that no more messages are received after a disconnection? It looks like taking this out reduces the scope of this test.

Morganamilo: Disconnect is a placeholder action that's not actually meant to do anything. https://ably.com/docs/client-lib-development-guide/protocol#actions

Rosalita: Thanks for the link. Sorry one more question, do you think this test should also be testing the client receiving the detached action here? If we think that should be covered can create a new issue so its not forgotten. Adding a new test scenario would be out of scope for this work to stabilise the existing test.

Rosalita: I can see we already have tests that send messages with `ably.ActionDetached` https://github.com/ably/ably-go/blob/main/ably/realtime_channel_spec_integration_test.go#L1016 so this feels like it isn't needed here

PR #378, opened 2021-08-15
Fix Message Spec

sacOO7:
- Fixes #377 
- Fixes #58
- Fixes #293

PR #501, opened 2022-05-05
Refactored code and tests for TM3 message encoding/decoding

Morganamilo:
This is taken from https://github.com/ably/ably-go/pull/378. Once this is merged that PR can be closed.

### DIFF
--- a/ably/proto_message_test.go
+++ b/ably/proto_message_test.go
@@ -19,7 +19,7 @@ func (custom) MarshalJSON() ([]byte, error) {
 	return json.Marshal([]string{"custom"})
 }
 
-func TestMessage(t *testing.T) {
+func TestMessage_EncodeDecode_TM3(t *testing.T) {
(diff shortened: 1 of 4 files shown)


ably/proto_message_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/ably/ably-go/ably"
+	"github.com/ably/ably-go/ablytest"
```
Rosalita, line 12: If `ablytest` package is imported here, this file which previously only contained unit tests will now contain some integration tests. 

This file still has the build tags to run in CI at the unit test stage. We should probably leave this file alone as a place for unit tests and create a new file `proto_message_integration_test.go` for the new tests added in this PR so it can have the build tags for the integration stage of CI 
```
//go:build !unit
// +build !unit
```

Morganamilo: Done, thanks.

Rosalita (approved):
👍

sacOO7 (approved):

PR #513, opened 2022-05-12
Prefix names of junit test reports with run order

### LINKED ISSUE #507: Difficult to tell which testsuite of CI tests ran in the test observability report.
The tests on this project run in CI in the order
1. Unit Tests
2. Integration Tests with JSON Protocol
3. Integration Tests with MessagePack Protocol

However these test results appear in the test observability server in a different order
 1. Integration Tests with JSON Protocol
 2. Integration Tests with MessagePack Protocol
 3. Unit Tests

It looks like this is because the upload action here matches the file names on the pattern `*.junit` 
https://github.com/ably-labs/test-observability-action/blob/7b03dd9cc6366038c1046118a2c389cb0be8bd13/index.js#L13

Then does a for each loop over the result, which would iterate them in alphabetical order of file names.
We currently name these files
*  `json-integration.junit`
* `msgpack-integration.junit`
* `unit.junit`

I would propose a quick fix to rename these files so their iteration order in the test observability server matches the order they are run in our CI pipeline. Prefixing them with a number would work.

Rosalita:
Resolves #507

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -41,12 +41,12 @@ jobs:
       - name: Unit Tests
         run: |
           set -o pipefail
-          go test -v -tags=unit ./... |& tee >(~/go/bin/go-junit-report > unit.junit)
+          go test -v -tags=unit ./... |& tee >(~/go/bin/go-junit-report > 1-unit.junit)
 
       - name: Integration Tests with JSON Protocol
         run: |
           set -o pipefail
-          scripts/test.sh --protocol application/json |& tee >(~/go/bin/go-junit-report > json-integration.junit)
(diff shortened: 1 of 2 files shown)


Morganamilo:
Did the commit get lost? All I see this do is change white space.

Morganamilo:
Oh wait this is still a draft. Maybe that's why.

Rosalita:
@Morganamilo yes this is still in draft, its currently only a whitespace change as I needed to check if the test observability server had broken. It appears a PR that merged around 17:00 yesterday broke the test observability upload.

Rosalita:
No longer needed as original issue has been closed and https://github.com/ably-labs/test-observability-server/issues/51 has been raised

PR #515, opened 2022-05-13
Split integration tests into separate workflow

owenpearson:
Moves the integration tests into a separate workflow file. Naming convention comes from our [SDK engineering guidance](https://github.com/ably/engineering/blob/main/sdk/github.md#filenames).

This has two main advantages over keeping them in the check workflow:
1. Tests will run in parallel on separate github runners so CI runs should be much faster. Based on the first few runs it seems to have gone from on average of about 35 minutes down to about 17.
2. Will allow us to re-run integration tests for one protocol without having to re-run tests for the other protocol if they're already passing, which should mean much less re-runs required to see all tests passing.

#### For reviewers
- The arcane syntax used in integration-test.yml line 36 comes from https://github.com/actions/runner/issues/409#issuecomment-727565588 which is a workaround for the fact that github workflow syntax doesn't have a proper ternary if statement. An alternative could be to move that logic into a shell script file and just pass in `${{ matrix.protocol }}` as an arg. Leave a comment if you think this would be worth doing.
- I'm not all that familiar with how test-observability works. The uploading seems to still work as expected but bear in mind that I might have inadvertently changed some part of the existing setup.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,6 +1,3 @@
-# Based upon:
-#   https://github.com/actions/starter-workflows/blob/main/ci/go.yml
-
 on:
   pull_request:
   push:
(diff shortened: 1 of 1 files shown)


Rosalita (approved):
I think this is a great idea. It will reduce the amount of time CI currently takes to run and should definitely reduce the need to retry the whole CI run if a test randomly fails.

PR #521, opened 2022-05-17
#509 add support to get lifecycle status

### LINKED ISSUE #509: Add support to get channel lifecycle status
Add support to get channel lifecycle status 

Channel lifecycle status is described at [Channel status / REST / [Docs]](https://ably.com/docs/rest/channel-status#metadata-rest)

It is currently possible to retrieve channel lifecycle status in ably-go by using REST.Request and marshalling the result into a slice of empty interface.

**Example:**

```go
result, _ := restClient.Request("get", "/channels/test").Pages(context.Background())

result.Next(ctx)

var items []interface{}
if err := result.Items(&items); err != nil {
	fmt.Println(err)
}

fmt.Printf("%+v\n", items)
// [map[channelId:test name:test status:map[isActive:true occupancy:map[metrics:map[connections:1 presenceConnections:1 presenceMembers:1 presenceSubscribers:1 publishers:1 subscribers:1]]]]]
```

**Out of Scope**
Realtime subscription to Channel Lifecycle Events. This ticket is only for supporting REST client ‘Get’.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-1851) by [Unito](https://www.unito.io)
┆sprint: SDK 2022-Q2-S4 12/05

mohyour:
Fixes #509


ably/rest_channel_spec_integration_test.go
```diff
@@ -162,6 +162,24 @@ func TestHistory_Direction_RSL2b2(t *testing.T) {
...
 
+func TestGetChannelLifecycleStatus_RSL8(t *testing.T) {
+	app, realtime := ablytest.NewRealtime()
+	app2, rest := ablytest.NewREST()
+	defer app.Close()
+	defer app2.Close()
+	channel := realtime.Channels.Get("lifecycle:test")
+	restchannel := rest.Channels.Get("lifecycle:test")
+
+	channel.Publish(context.Background(), "event", "data")
```
lmars, line 173: The returned error should be checked.

ably/rest_channel.go
```diff
@@ -157,6 +157,82 @@ func (c *RESTChannel) PublishMultipleWithOptions(ctx context.Context, messages [
...
+	Connections         int
+	PresenceConnections int
+	PresenceMembers     int
+	PresenceSubscribers int
+	Publishers          int
+	Subscribers         int
+}
+
+// Status - returns ChannelDetails representing information for a channel
+func (c *RESTChannel) Status(ctx context.Context) (ChannelDetails, error) {
```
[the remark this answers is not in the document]
Rosalita: 👍  for this. It also means that if an error is returned, `nil` can be returned for the channel details instead of an empty `ChannelDetails{}`

ably/rest_channel.go
```diff
@@ -157,6 +160,50 @@ func (c *RESTChannel) PublishMultipleWithOptions(ctx context.Context, messages [
...
+	}
+	res, err := c.client.MakeHTTPRequest(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	var channelDetails ChannelDetails
+
+	// Decode response body
+	dec := codec.NewDecoder(res.Body, new(codec.MsgpackHandle))
```
lmars, line 200: I think here we just want to call [`decodeResp`](https://github.com/ably/ably-go/blob/main/ably/rest_client.go#L864-L873) which will check the `Content-Type` header to determine whether to decode msgpack or JSON.

ably/rest_channel_spec_integration_test.go
```diff
@@ -162,6 +162,21 @@ func TestHistory_Direction_RSL2b2(t *testing.T) {
...
+	channel := rest.Channels.Get("lifecycle:test")
+
+	ctx := context.Background()
+	err := channel.Publish(ctx, "event", "data")
+	assert.NoError(t, err)
+	status, err := channel.Status(ctx)
+	assert.NoError(t, err)
+	assert.NotNil(t, status.ChannelId)
+	assert.True(t, status.Status.IsActive)
+	assert.Equal(t, "lifecycle:test", status.ChannelId)
```
Rosalita, line 177:
```suggestion
	ctx := context.Background()
	app, rest := ablytest.NewREST()
	defer app.Close()
	
	t.Run("Test Channel Status after Publish", func(t *testing.T) {
		channel := rest.Channels.Get("lifecycle:test")
		err := channel.Publish(ctx, "event", "data")
		assert.NoError(t, err)
		status, err := channel.Status(ctx)
		assert.NoError(t, err)
		assert.NotNil(t, status.ChannelId)
		assert.True(t, status.Status.IsActive)
		assert.Equal(t, "lifecycle:test", status.ChannelId)
	})
```
There's been issues with some of the existing tests where the test setup was failing if it was retried (on going work to address this in https://github.com/ably/ably-go/issues/526). 

So I think we need to split the test setup where the app is instantiated from the main body of the test. This can be done by putting the main body of the test inside a sub-test and leaving the app instantiation outside of the sub-test.

ably/rest_client.go
```diff
@@ -360,30 +360,32 @@ func (c *REST) Request(method string, path string, o ...RequestOption) RESTReque
...
+	var lastResponse *http.Response
+	resp, err := c.doWithHandle(ctx, req, func(resp *http.Response, out interface{}) (*http.Response, error) {
+		// Save the resp but return an error on bad status to trigger fallback
+		lastResponse = resp
+		return c.handleResponse(resp, nil)
+	})
+	if err != nil && lastResponse != nil {
+		return lastResponse, nil
+	}
+	return resp, err
```
lmars, line 386: I looked for other uses of `REST.doWithHandle`, and there is [`REST.do`](https://github.com/ably/ably-go/blob/main/ably/rest_client.go#L599-L601) which I think does exactly what this method is doing (i.e. sending a request, and passing the response to `REST.handleResponse`).

Also, it looks like `REST.handleResponse` will call `decodeResp` if `req.Out` is set, so that should avoid the need to do it in `RESTChannel.Status`.

Also, this is why I'm encouraging a refactor here, there is just way too much indirection here!

README.md
```diff
@@ -289,6 +289,16 @@ if err := pages.Err(); err != nil {
 }
 ```
 
+### Getting the channel details
```
Rosalita, line 292:
```suggestion
### Getting the channel status
```
The example code below this heading uses the word status

Rosalita (approved):
Thank you for addressing my comments. I'm happy to approve :)

lmars (approved):
Nice!

PR #531, opened 2022-05-26
Remove dependency on test.sh bash script to run the tests

### LINKED ISSUE #530: Remove dependency on test.sh bash script to run the tests
Currently this project relies on a script called `test.sh` to run the tests, which is a little unexpected. Usually Go projects will use the `go test` command to start the tests.

Looking at the `test.sh` script, it can set an environment variable for to run the tests for JSON or message pack then start running all the tests. Or it can run all the tests for both environment variables.

To reduce test flakiness, we now run tests for each protocol in its own Github runner (as running tests for both protocols in the same run was prone to test flakiness).

It would be possible to define the setting of environment variables as a step inside the pipeline with something like

```
  - name: Set env var for protocol
        run: echo "ABLY_PROTOCOL=${{ matrix.protocol }}" >> $GITHUB_ENV
```

Then inside the pipeline instead of running the `test.sh` the `go test` command could be used e.g. `go test -tags=integration -p 1 -race -v -timeout 120m ./...`

This would be a more readable, transparent way of running the tests in the pipeline and allow us to remove the dependency on a bash script which could then be deleted.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1932) by [Unito](https://www.unito.io)
┆sprint: SDK 2022-Q2-S5 26/05

Rosalita:
Resolves #530


scripts/test.sh
```diff
@@ -1,61 +0,0 @@
-#!/usr/bin/env bash
-#
-# A script to run the tests.
```
sacOO7, line 3: Actually, we were using this script to run all of the tests on the local, before checking them on CI. There are certain go test flags, which are easy to miss out otherwise while running tests.

.github/workflows/integration-test.yml
```diff
@@ -30,10 +30,13 @@ jobs:
       - name: Install go-junit-report
         run: go install github.com/jstemmer/go-junit-report@latest
 
+      - name: Set env var for ablytest
+        run: echo "ABLY_PROTOCOL=${{ matrix.protocol == 'json' && 'application/json' || 'application/x-msgpack' }}" >> $GITHUB_ENV
```
lmars, line 34: This could just be done within the script below rather than a separate step:

```yaml
run: |
  set -o pipefail
  export ABLY_PROTOCOL=${{ matrix.protocol == 'json' && 'application/json' || 'application/x-msgpack' }}
  go test -tags=integration ...

```

scripts/test.sh
Rosalita: Hi @sacOO7 please be aware that if you were using `test.sh` to run all the tests on this project, this would take twice as long to run all the tests as it now does to run them in CI (where they run in parallel as each suite has it's own runner).

The commands to run the test can always be copied directly from the pipeline's GitHub action .yml files. The project should not have to rely on a bash script when the pipeline can be expressed directly in the GitHub action .yml file. Please see #530 for more information and let me know if you still have any objection to the removal of this script.

.github/workflows/integration-test.yml
```diff
@@ -29,11 +29,12 @@ jobs:
 
       - name: Install go-junit-report
         run: go install github.com/jstemmer/go-junit-report@latest
-
+
```
lmars, line 32: This whitespace change isn't needed

lmars (approved):

.github/workflows/integration-test.yml
Rosalita: Looks like my IDE went rogue and put a tab in there. Removed now.

scripts/test.sh
sacOO7: `The commands to run the test can always be copied directly from the pipeline's GitHub action .yml files`. If we are planning to do this, can we update the `README` file for the same? I can see the `README` file section -> https://github.com/ably/ably-go#integration-tests still mentions the use of `scripts/test.sh`. Also, it will be great if we mention relevant flags as specified by the runner.

Morganamilo:
I've wanted to do this for a while. The script is long and only has one line of value with some params change. The line can just be added to the readme if needed.

scripts/test.sh
Rosalita: Good spot with the readme file. I've just pushed an update to remove reference to `test.sh` and add the commands to run the tests.

sacOO7: Nice work 👍

sacOO7 (approved):
LGTM

PR #534, opened 2022-05-30
Increased unit test coverage

Rosalita:
This PR raises overall unit test coverage for package `ably-go` from 22.4% to 27.8%
`realtime_client.go` had 15.6% coverage, it now has 40.6% coverage.
`realtime_channel.go` had 4.5% coverage, it now has 12.2% coverage.

### DIFF
--- a/ably/realtime_channel_internal_test.go
+++ b/ably/realtime_channel_internal_test.go
@@ -1,0 +1,199 @@
+//go:build !integration
+// +build !integration
+
+package ably
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestChannelOptionChannelWithCipherKey(t *testing.T) {
+	tests := map[string]struct {
+		key            []byte
+		expectedResult *channelOptions
+	}{
+		"Can inject a cipher key of length 128 into cipher params": {
+			key: []byte{82, 27, 7, 33, 130, 101, 79, 22, 63, 95, 15, 154, 98, 29, 114, 19},
+			expectedResult: &channelOptions{
+				Cipher: CipherParams{
+					Algorithm: CipherAES,
+					KeyLength: 128,
+					Key:       []uint8{0x52, 0x1b, 0x7, 0x21, 0x82, 0x65, 0x4f, 0x16, 0x3f, 0x5f, 0xf, 0x9a, 0x62, 0x1d, 0x72, 0x13},
+					Mode:      CipherCBC,
+				},
+			},
+		},
+
+		"Can inject a cipher key of length 256 into cipher params": {
+			key: []byte{82, 27, 7, 33, 130, 101, 79, 22, 63, 95, 15, 154, 98, 29, 114, 19, 10, 23, 45, 56, 76, 29, 111, 23, 93, 22, 44, 66, 88, 43, 72, 42},
+			expectedResult: &channelOptions{
+				Cipher: CipherParams{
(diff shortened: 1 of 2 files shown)


ably/realtime_channel_internal_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	for testName, test := range tests {
+		t.Run(testName, func(t *testing.T) {
+			var opt ChannelOption
+			if len(test.modes) == 1 {
+				opt = ChannelWithModes(test.modes[0])
+			} else {
+				opt = ChannelWithModes(test.modes[0], test.modes[1])
+			}
```
lmars, line 133:
```suggestion
			opt := ChannelWithModes(test.modes...)
```

ably/realtime_channel_internal_test.go
```diff
@@ -0,0 +1,200 @@
...
+		t.Run(testName, func(t *testing.T) {
+
+			//These mocks must be created here because type RealtimeChannel
+			//contains a sync.Mutex which is a value that cannot be copied.
+			//TODO: investigate representing sync.Mutex with sync.Locker interface.
+			//see https://github.com/ably/ably-go/issues/535
+
+			mockedExistingChannel := RealtimeChannels{
+				chans: map[string]*RealtimeChannel{
+					"existing": {
```
lmars, line 168: This test feels a little too brittle.

If what we want to ensure is that calling `Get` twice for the same name returns the same object, I'd expect a test like:

```go
channel1 := channels.Get("test")
channel2 := channels.Get("test")
assert.Same(t, channel1, channel2)
```

Rosalita: As this is a unit test, I don't agree with this suggestion. This change would make the test tautological. See https://randycoulman.com/blog/2016/12/20/tautological-tests/ 

In unit tests, assertions absolutely must be hard coded, never calculated from function calls. If the unit test was implemented this way, with the valid string "test" as it's input to Get, it will never fail, so it has no value as a test.

What you are suggesting however, to call twice and check the outputs are the same sounds, would be a great idea for a fuzzing test. Fuzz testing is a new feature in Go 1.18. A fuzz test would use a corpus to generate the string which is passed into Get, it would automatically mutate that string and keep calling, checking for bad inputs and automatically save any inputs that cause unexpected panic. 

We don't have any fuzz tests yet, do you think it's worth starting to add them?

lmars: > I've raised https://github.com/ably/ably-go/issues/536 to start a new experiment to add the first fuzz test to the project.

Great, thanks 👍 

> I have simply removed this test for now.

I do really think it's worthwhile having test coverage for returning an existing channel, so I don't think we need to remove the test, just perhaps make it less brittle?

Rosalita: I know that this PR is not ideal as it is an attempt to retro-fits unit tests after code has been written, which is never a good approach. Some of the "units" are especially tricky to test as testability was not a consideration when they were built. Also this project is not set up to use interfaces which makes mocking harder.

While we are unable to reach agreement on some of these tests, it does feel like we do agree on some of them.  

In the interests of progress, I would like to get the tests we can agree on merged.

If you have any good ideas for how to test returning an existing channel I would be open to discussion on a new PR.

lmars: Ok no problem 👍

lmars (approved):

PR #548, opened 2022-06-23
Update action

Morganamilo:


owenpearson (approved):

Rosalita (commented):
There are three tags at https://github.com/ably/test-observability-action/tags
v1 and v1.0.1 were both created around the same time. What is the difference between them? Why is v1 used here instead of v1.0.1?

Morganamilo:
v1 points to the latest v1.x.x release.

PR #556, opened 2022-10-10
Update to canonical doc

### LINKED ISSUE #558: Add intro blurb to the Go static site generated
To standardise all SDK docs and help customers understand the difference between auto-generated docs and other API reference docs available on the main site, we’ve worked on a generic piece of text to include on the intro page of all SDKs. As part of this ticket, the following text needs to be added to the intro page of the Go SDK.

```go
# Ably Go Client Library SDK API Reference

The Go Client Library SDK supports a realtime and a REST interface. The Go API references are generated from the [Ably Go Client Library SDK source code](https://github.com/ably/ably-go/) using [godoc](https://pkg.go.dev/golang.org/x/tools/cmd/godoc) and structured by classes.

The realtime interface enables a client to maintain a persistent connection to Ably and publish, subscribe and be present on channels. The REST interface is stateless and typically implemented server-side. It is used to make requests such as retrieving statistics, token authentication and publishing to a channel.

View the [Ably docs](https://ably.com/docs/) for conceptual information on using Ably, and for API references featuring all languages. The combined [API references](https://ably.com/docs/api/) are organized by features and split between the [realtime](https://ably.com/docs/api/realtime-sdk) and [REST](https://ably.com/docs/api/rest-sdk) interfaces.
```

### LINKED ISSUE #560: Add/update docstring comments in Go SDK
## Background

We wanted to update the experience our developer audiences have with our API Reference documentation. This is currently only hosted at https://ably.com/docs/api . When we looked into a long-term process to keep the API Ref content up to date, we realised that it needs to come from each SDK separately as features get implemented. We also wanted to ensure that the descriptions of various methods and properties across different SDKs are shown in a standard format.

To enable the above outcomes for our customers, we came up with the following process

image

As you can see from the workflow above, we concluded that each SDK needs to have docstring comments and a static site (generated automatically using specific doc tools per SDK) - the content in this static site will be used by our tech writers to keep the content on the main docs website up to date and accurate.

## What needs to be done now?

To make sure that the comments are standardised across various SDKs, our tech writers have built a table of descriptions which are language/ SDK agnostic(based on the feature spec and IDL). You can find this at: https://github.com/ably/canonical-api-reference-prototyping/blob/main/descriptions.md. Based on this table, the following needs to be done in the Go SDK:

1. Mark all existing comments as legacy comments (so they can be reviewed separately if they weren’t already covered in the canonical table as there may be some Go SDK/language-specific comments)
2. Copy over the canonical descriptions to the right places in the SDK (update if a previous comment already exists)
3. Update copied over language-agnostic comments and covert them to language-specific ones

We’ll use [godoc](https://go.dev/blog/godoc)  to then generate a static site using these comments. This is a separate ticket but good to be aware of in case it requires the comments to be written in a certain format.

--

Please refer to the ably-js PR for an understanding of how the conversion to a specific language was done in JS https://github.com/ably/ably-js/pull/1026 . If you feel like there’s a better approach to doing this please add a comment and we can have the SDK team review/approve.

sacOO7:
- Fixes #560 
- Fixes #558 
- Related refactorings https://github.com/ably/sdk-api-reference/pull/29

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,27 @@ Because this package uses `internal` packages, all fork development has to happe
 7. push to the branch: `git push fork my-new-feature`
 8. create a new Pull Request.
 
+### Local web-based godoc
+
+1. Install [godoc](https://pkg.go.dev/golang.org/x/tools/cmd/godoc) globally via `go get` and run at root
+
+```bash
+  godoc -http=:8000
+```
+- Open the link http://localhost:8000/ for viewing the documentation.
+
+2. Export `godoc` using [gopages](https://pkg.go.dev/github.com/johnstarich/go/gopages#section-readme) (only works on unix based systems)
+```bash
+  gopages -brand-description "Go client library for Ably realtime messaging service." -brand-title "Ably Go SDK"
+```
+- `godoc html` is exported to `dist` and can be served using `python3 http.server`
+
+```bash
+  cd dist
+  py -m http.server 8000
+```
+- Open the link http://localhost:8000/ for viewing the documentation.
+
 ### Running Tests
 
 This project contains two types of test. Test which use the `ablytest` package and tests which dont. 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ Because this package uses `internal` packages, all fork development has to happe
 ```
 - Open the link http://localhost:8000/ for viewing the documentation.
 
-2. Export `godoc` using [gopages](https://pkg.go.dev/github.com/johnstarich/go/gopages#section-readme) (only works on unix based systems)
+2. Export `godoc` using latest version of [gopages](https://pkg.go.dev/github.com/johnstarich/go/gopages#section-readme)
 ```bash
   gopages -brand-description "Go client library for Ably realtime messaging service." -brand-title "Ably Go SDK"
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Because this package uses `internal` packages, all fork development has to happe
 7. push to the branch: `git push fork my-new-feature`
 8. create a new Pull Request.
 
-### Local web-based godoc
+### Previewing Godoc Locally
 
 1. Install [godoc](https://pkg.go.dev/golang.org/x/tools/cmd/godoc) globally via `go get` and run at root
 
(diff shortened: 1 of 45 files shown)


README.md
```diff
@@ -291,6 +291,27 @@ if err != nil {
 fmt.Print(status, status.ChannelId)
 ```
 
+## Local web-based godoc
```
amnonbc, line 294: I am not sure about including this in the README, which is really documentation for people who are using the library rather than people developing the library. If we include a "Local web-based godoc" section, 
we should at least add a sentence on why anyone would want to use this, and in what way this is different from any other Go library.

sacOO7: - If developers want to update documentation for `ably-go`, say for a `method` or `class`, they need a way to `preview` it locally, without pushing to `CI` first, so that it's easy to make a quick correction, rather than waiting for CI to update it.
- For me personally, I had to look around for tools to generate the doc, most of the time answers were confusing or not available since the community is small. It's better if we have these standard tools documented somewhere.
- Also, I strongly believe this is a must setup while working on a new feature/refactoring where it will be needed to add documentation for the code, developer can have a quick preview and can correct it based on the presentation.

amnonbc: OK, so I would say, lets put this under a new section called "Contributing" or something similar,
to make it clear that this is aimed at the small minority of people who are _contributing to the library itself_, rather 
than _using_ the library to add realtime abilities to their own system.

amnonbc: This should be moved to [CONTRIBUTING.md](https://github.com/ably/ably-go/blob/main/CONTRIBUTING.md)

sacOO7: Moved as per suggestion 👍

CONTRIBUTING.md
```diff
@@ -11,6 +11,27 @@ Because this package uses `internal` packages, all fork development has to happe
 7. push to the branch: `git push fork my-new-feature`
 8. create a new Pull Request.
 
+### Local web-based godoc
```
QuintinWillison, line 14: I feel that "web-based" is a bit misleading here. Also, this title doesn't quite align with the next heading "Running Tests". So, perhaps this might be better:

```suggestion
### Previewing Godoc Locally
```

QuintinWillison (approved):
I've not reviewed the added API commentaries line by line because my understanding is that they're sourced from canonical references. I have one small suggestion and, of course, checks need to be green.

CONTRIBUTING.md
sacOO7: done

Issue #347, opened 2021-07-08
Surface API reference docs


QuintinWillison:
We may be able to use [gopages](https://johnstarich.com/go/gopages/pkg/github.com/johnstarich/go/gopages/), as suggested by lmars [here (internal)](https://ably-real-time.slack.com/archives/C0114KK7S6R/p1623926815099500?thread_ts=1623922381.097100&cid=C0114KK7S6R).

To be uploaded to `sdk.ably.com` using our [Upload Action](https://github.com/ably/sdk-upload-action), modelled on similar workflows we have in other repositories (e.g. [ably-flutter](https://github.com/ably/ably-flutter/blob/main/.github/workflows/docs.yml) and [ably-asset-tracking-android](https://github.com/ably/ably-asset-tracking-android/blob/main/.github/workflows/docs.yml)).

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1027) by [Unito](https://www.unito.io)

sacOO7:
Related to #559

PR #561, opened 2022-11-18
Publish godoc for every commit

### LINKED ISSUE #559: Set up the tool and generate static site with Go docstring comments
See the description in [https://ably.atlassian.net/browse/SDK-2950](https://ably.atlassian.net/browse/SDK-2950|smart-link) for the background on this. 

We’ll use [https://go.dev/blog/godoc](https://go.dev/blog/godoc)  to generate a static site. As part of this workflow, can you please set up and add it as part of the build workflow? The following are related resources for this workflow:

* [https://github.com/ably/engineering/blob/main/sdk/sdk.ably.com.md](https://github.com/ably/engineering/blob/main/sdk/sdk.ably.com.md)
* [https://github.com/ably/ably-flutter/blob/main/.github/workflows/docs.yml](https://github.com/ably/ably-flutter/blob/main/.github/workflows/docs.yml)

sacOO7:
- Added `docs.yml` github workflow file.
- Added steps to generate static html site and upload the same on ABLY AWS BUCKET.
- Fixed #559 
- Fixed https://github.com/ably/ably-go/issues/347

Update - Fixing styling for godoc depends on -> https://github.com/ably/sdk-upload-action/issues/32

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,11 +29,11 @@ jobs:
           go get github.com/johnstarich/go/gopages
           go install github.com/johnstarich/go/gopages
 
-      - uses: ably/sdk-upload-action@v1
-        id: sdk-upload-prempt
-        with:
-          mode: preempt
-          artifactName: godoc
+#      - uses: ably/sdk-upload-action@v1
+#        id: sdk-upload-prempt
+#        with:
+#          mode: preempt
+#          artifactName: godoc
 
       - name: Build Documentation
         run: >
@@ -41,7 +41,8 @@ jobs:
           -source-link "https://github.com/ably/ably-go/blob/${{ github.sha }}/ably/{{slice .Path 5}}{{if .Line}}#L{{.Line}}{{end}}"
           -brand-description "Go client library for Ably realtime messaging service."
           -brand-title "Ably Go SDK"
-          -base $BASE_URL
+          -base ./../../../../..
+#          -base $BASE_URL
 
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,6 +26,7 @@ jobs:
       - name: Download Packages
         run: |
           go install golang.org/x/tools/cmd/godoc@latest
+          go get github.com/johnstarich/go/gopages
           go install github.com/johnstarich/go/gopages
 
       - name: Build Documentation
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -40,7 +40,7 @@ jobs:
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-region: eu-west-2
-          role-to-assume: arn:aws:iam::${{ secrets.ABLY_AWS_ACCOUNT_ID_SDK }}:role/ably-sdk-builds-ably-js
+          role-to-assume: arn:aws:iam::${{ secrets.ABLY_AWS_ACCOUNT_ID_SDK }}:role/ably-sdk-builds-ably-go
           role-session-name: "${{ github.run_id }}-${{ github.run_number }}"
 
       - name: Upload Documentation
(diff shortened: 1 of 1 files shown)


sacOO7:
One of the questions would be whether to lock versioning for `godoc` and `gopages`. 
This would avoid possible bugs getting introduced in the future due to the release of new versions.

amnonbc:
> One of the questions would be whether to lock versioning for `godoc` and `gopages`. This would avoid possible bugs getting introduced in the future due to the release of new versions.

Yes, I think so.
godoc and gopages are dependencies which we want to treat like all other dependancies - update them as a developer action, rather than as an accidental byproduct of doing a new build.

Unrelated, but we should remove Go 1.17 from the build matrix (as it is no longer supported upstream)
and add Go 1.19

sacOO7:
> > One of the questions would be whether to lock versioning for `godoc` and `gopages`. This would avoid possible bugs getting introduced in the future due to the release of new versions.
> 
> Yes, I think so. godoc and gopages are dependencies which we want to treat like all other dependancies - update them as a developer action, rather than as an accidental byproduct of doing a new build.
> 
> Unrelated, but we should remove Go 1.17 from the build matrix (as it is no longer supported upstream) and add Go 1.19
> 
> p.s. I have updated the build matrix in #563
 
Locked dependencies as suggested!

.github/workflows/docs.yml
```diff
@@ -25,9 +25,9 @@ jobs:
 
       - name: Download Packages
         run: |
-          go install golang.org/x/tools/cmd/godoc@latest
-          go get github.com/johnstarich/go/gopages
-          go install github.com/johnstarich/go/gopages
+          go install golang.org/x/tools/cmd/godoc@v0.3.0
+          go get github.com/johnstarich/go/gopages@v0.1.11
```
amnonbc, line 29: If you do `go install github.com/johnstarich/go/gopages@v0.1.11`
Then you don't need to `go get github.com/johnstarich/go/gopages@v0.1.11`

sacOO7: Seems to be working, done 👍

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+#          artifactName: godoc
+
+      - name: Build Documentation
+        run: >
+          gopages
+          -source-link "https://github.com/ably/ably-go/blob/${{ github.sha }}/ably/{{slice .Path 5}}{{if .Line}}#L{{.Line}}{{end}}"
+          -brand-description "Go client library for Ably realtime messaging service."
+          -brand-title "Ably Go SDK"
+          -base ./../../../../..
+#          -base $BASE_URL
```
QuintinWillison, line 44: It would be nice to remove this commented out line, given there's no accompanying commentary explaining why it's been kept here.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.19
+
+      - name: Download Packages
+        run: |
+          go install golang.org/x/tools/cmd/godoc@v0.3.0
+          go install github.com/johnstarich/go/gopages@v0.1.16
+
+#      - uses: ably/sdk-upload-action@v1
```
QuintinWillison, line 31: Please remove this commented out step.

QuintinWillison (changes requested):
Was the issue with base URL sorted out? We're yet to complete https://github.com/ably/sdk-upload-action/issues/32 🤔

.github/workflows/docs.yml
sacOO7: > Was the issue with base URL sorted out? We're yet to complete [ably/sdk-upload-action#32](https://github.com/ably/sdk-upload-action/issues/32) 🤔

It's temporarily sorted by specifying the baseUrl. For permanent fix, we need to resolve the issue.

sacOO7: removed, but we need to update this once we fix https://github.com/ably/sdk-upload-action/issues/32

.github/workflows/docs.yml
sacOO7: removed

Peter-Maguire (approved):

QuintinWillison (approved):

PR #563, opened 2022-11-22
Test against go 1.19, and drop support for 1.17

amnonbc:
See https://github.com/ably/ably-go#supported-versions-of-go

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.17, 1.18]
+        go-version: [1.18, 1.19]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -30,7 +30,14 @@ jobs:
         run: go get -t -v ./ably/...
       
       - name: Format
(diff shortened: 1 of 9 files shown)


QuintinWillison (commented):
Yet to see what CI checks are failing and why... however I do wonder if [this line in `go.mod`](https://github.com/ably/ably-go/blob/5707c707ee53be695956daa28e9817655cfb20de/go.mod#L17) should also be modified as part of this change.

amnonbc:
> Yet to see what CI checks are failing and why... 

go 1.19 slightly changes the output of gofmt
(in order to acomodate, links, markdown etc.)
So this changes the output of gofmt...

I can't really think of any fix other than removing this test, or only run it for go-1.19
and perhaps putting it back, once go 1.20 comes out and we have obsoleted 1.18
and run gofmt on everything (which is something I don't want to do while other folk have PRs in flight).

What do you think?
We do want to ensure that all code has passed gofmt,
but we have got to ensure that we support the the latest Go.

> however I do wonder if [this line in go.mod](https://github.com/ably/ably-go/blob/5707c707ee53be695956daa28e9817655cfb20de/go.mod#L17) should also be modified as part of this change.

I don't think it matters, until we start using go-1.18 or 1.19 language features (e.g. Generics).

QuintinWillison (approved):

Issue #574, opened 2022-12-22
ably-go should not convert `[]byte` messages to strings


amnonbc:

amnonbc:
```
func TestUnmarshallByte(t *testing.T) {
	buf := []byte{
		0xc4,     // bin8
		2,        // len
		'a', 'a', // bytes
	}
	var target interface{}

	err := UnmarshalMsgpack(buf, &target)
	require.NoError(t, err)
	assert.IsType(t, []byte{}, target,
		"bin8 should be decoded as []byte, but instead we got %T", target)
}
```

In  github.com/ugorji/go/codec@v1.1.9/msgpack.go:494
```
		case bd == mpBin8, bd == mpBin16, bd == mpBin32:
			fauxUnionReadRawBytes(d, &d.d, n, d.h.RawToString)
```

which calls 
```
func fauxUnionReadRawBytes(dr decDriver, d *Decoder, n *fauxUnion, rawToString bool) {
	if rawToString {
		n.v = valueTypeString
		n.s = string(dr.DecodeBytes(d.b[:], true))
	} else {
		n.v = valueTypeBytes
		n.l = dr.DecodeBytes(nil, false)
	}
}
```

amnonbc:
I added a bug report to https://github.com/ugorji/go/issues/387

amnonbc:
See https://github.com/ably/ably-go/pull/581 as a sample fix.

PR #581, opened 2023-01-10
Data sent as []byte should be decoded as []byte

amnonbc:
Fixes https://github.com/ably/ably-go/issues/574

The spec states:

[(RSL6a)](https://sdk.ably.com/builds/ably/specification/main/features/#RSL6a) All messages received will be decoded automatically based on the encoding field and the payloads will be converted into the format they were originally sent using i.e. binary, string, or JSON

In this case the message was sent as []byte so should have been received as as []byte.

This PR fixes the behaviour when receiving a `[]byte` message.

The problem is that merging this PR could cause user code to panic at runtime, if it assumes the old incorrect behaviour 
and does something like

`payload := msg.Data.(string)`

Is this a breaking change which requires a new major version?
How many customers are using ably-go to receive realtime messages? 
The SDK team will need to decide.


sacOO7 (changes requested):
Can you add one more msgpack integration test that sends binary data and checks on another side if data is received in the required format?

amnonbc:
> Can you add one more msgpack integration test that sends binary data and checks on another side if data is received in the required format?

done in  68c0afe

sacOO7:
Are we waiting for anymore changes on this PR? I remember there was issue the msgpack library for encoding/decoding data?

amnonbc:
> Are we waiting for anymore changes on this PR? I remember there was issue the msgpack library for encoding/decoding data?

Yes, I am adding some more fixture tests. Should be done by the end of the weekend.

But the SDK team will need to discuss what we do with this PR, as it will break current behaviour (see PR comment).

amnonbc:
I am going to close this for now as it is a breaking change, which will cause currently working code to panic at runtime.
This change is too small to justify a v2 api on its own. So I will close the PR for now, but link it to the bug, as an example of how the bug can be fixed when the SDK team does a v2.

PR #588, opened 2023-03-06
Run unit tests based on msgpack_test_fixtures

amnonbc:
Run unit tests using fixtures from  https://github.com/ably/ably-common/blob/main/test-resources/messages-encoding.json and https://github.com/ably/ably-common/blob/main/test-resources/msgpack_test_fixtures.json
Ensure that the json and msgpack messages are decoded correctly, and that re-encoding the decoded message 
gets you back to the original fixture.

### DIFF
--- a/ably/proto_message_decoding_test.go
+++ b/ably/proto_message_decoding_test.go
@@ -1,0 +1,152 @@
+package ably
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/ably/ably-go/ably/internal/ablyutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type fixture struct {
+	Data             string `json:"data"`
+	Encoding         string `json:"encoding"`
+	ExpectedType     string `json:"expectedType"`
+	ExpectedValue    any    `json:"expectedValue"`
+	ExpectedHexValue string `json:"expectedHexValue"`
+}
+
+func loadFixtures() ([]fixture, error) {
+	var dec struct {
+		Messages []fixture
+	}
+
+	// We can't embed the fixtures as go:embed forbids embedding of resources above the current directory.
+	text, err := os.ReadFile("../common/test-resources/messages-encoding.json")
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.Unmarshal(text, &dec)
+	return dec.Messages, err
+}
+
+func Test_decodeMessage(t *testing.T) {
+	fixtures, err := loadFixtures()
+	require.NoError(t, err, "failed to load test fixtures")
+	for _, f := range fixtures {
+		t.Run(f.Data, func(t *testing.T) {
+			msg := Message{
+				Data:     f.Data,
+				Encoding: f.Encoding,
+			}
+			decodedMsg, err := msg.withDecodedData(nil)
+			require.NoError(t, err)
+			switch f.ExpectedType {
+			case "string":
+				assert.IsType(t, "string", decodedMsg.Data)
+				assert.Equal(t, f.ExpectedValue, decodedMsg.Data)
+			case "jsonObject":
+				assert.IsType(t, map[string]any{}, decodedMsg.Data)
+				assert.Equal(t, f.ExpectedValue, decodedMsg.Data)
+
+			case "jsonArray":
+				assert.IsType(t, []any{}, decodedMsg.Data)
+				assert.Equal(t, f.ExpectedValue, decodedMsg.Data)
+			case "binary":
+				assert.IsType(t, []byte{}, decodedMsg.Data)
+				if f.ExpectedHexValue != "" {
+					value, err := hex.DecodeString(f.ExpectedHexValue)
+					require.NoError(t, err)
+					assert.Equal(t, value, decodedMsg.Data)
+				} else if f.ExpectedValue != nil {
+					assert.Equal(t, []byte(f.ExpectedValue.(string)), decodedMsg.Data)
+				}
+			}
+
+			// Test that the re-encoding of the decoded message gives us back the original fixture.
+			reEncoded, err := decodedMsg.withEncodedData(nil)
+			require.NoError(t, err)
+			assert.Equal(t, f.Encoding, reEncoded.Encoding)
+			if f.Encoding == "json" {
+				require.IsType(t, "", reEncoded.Data)
+				// json fields could be re-ordered so assert.Equal could fail.
+				assert.JSONEq(t, f.Data, reEncoded.Data.(string))
+			} else {
+				assert.Equal(t, f.Data, reEncoded.Data)
+			}
+		})
+	}
+}
+
+type MsgpackTestFixture struct {
+	Name      string `json:"name"`
+	Data      any    `json:"data"`
+	Encoding  string `json:"encoding"`
+	NumRepeat int    `json:"numRepeat"`
+	Type      string `json:"type"`
+	MsgPack   string `json:"msgpack"`
+}
+
+func loadMsgpackFixtures() ([]MsgpackTestFixture, error) {
+	var o []MsgpackTestFixture
+
+	// We can't embed the fixtures as go:embed forbids embedding of resources above the current directory.
+	text, err := os.ReadFile("../common/test-resources/msgpack_test_fixtures.json")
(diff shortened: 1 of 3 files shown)


ably/proto_message_decoding_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/ably/ably-go/ably/internal/ablyutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
lmars, line 13: Should we use either `assert` or `require` rather than a mixture of both?

ably/proto_message_decoding_test.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+type fixture struct {
+	Data          string `json:"data"`
+	Encoding      string `json:"encoding"`
+	ExpectedType  string `json:"expectedType"`
+	ExpectedValue any    `json:"expectedValue"`
+}
+
+func loadFixtures() ([]fixture, error) {
```
lmars, line 23: There is some similar code for loading the crypto fixtures in the `ablytest` package [here](https://github.com/ably/ably-go/blob/main/ablytest/cryptodata.go), is it worth being consistent with that existing code?

ably/proto_message_decoding_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	err = json.Unmarshal(text, &dec)
+	return dec.Messages, err
+}
+
+func Test_decodeMessage(t *testing.T) {
+	fixtures, err := loadFixtures()
+	require.NoError(t, err, "failed to load test fixtures")
+	for _, f := range fixtures {
+		t.Run(f.Data, func(t *testing.T) {
```
lmars, line 42: Is it worth giving each fixture a human readable name rather than using the data as the test name?

ably/proto_message_decoding_test.go
```diff
@@ -0,0 +1,147 @@
...
+			case "jsonObject":
+				assert.IsType(t, map[string]any{}, decodedMsg.Data)
+				assert.Equal(t, f.ExpectedValue, decodedMsg.Data)
+
+			case "jsonArray":
+				assert.IsType(t, []any{}, decodedMsg.Data)
+				assert.Equal(t, f.ExpectedValue, decodedMsg.Data)
+			case "binary":
+				assert.IsType(t, []byte{}, decodedMsg.Data)
+				if f.ExpectedValue == nil {
```
lmars, line 62: Should all fixtures not have an expected value?

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/ably/ably-go
 
+go 1.18
+
```
lmars, line 4: Did you intend to include this change? It seems unrelated

ably/proto_message_decoding_test.go
amnonbc: I am all for reducing dependencies. 
But assert and require do different things.
We use assert to record an error, but allow the test to continue to run, so other failures can be found.
On the other hand use require for things like failure to open the fixture file, where there is no way to continue the error and we need to abort the test now.

ably/proto_message_decoding_test.go
amnonbc: Both functions read test fixture data. But the data does have different schemas, and different encodings.
And LoadCryptoData() returns different data types. 
DNRY is a good principle. But here, a I think in this case, and in loadMsgpackFixtures, a bit of repetition leads to simpler and more understandable code

ably/proto_message_decoding_test.go
amnonbc: That is what I did in the msgpack fixtures.
But the messages-encoding.json file was created by [many years ago](https://github.com/ably/ably-common/pull/17) and does not have a name field. The fixtures in this file were not used in any tests in ably-go, so I decided to add one. But I did not intend to alter this file, which probably is imported by other SDKs.

ably/proto_message_decoding_test.go
amnonbc: if they have a legal decoding, yes
But "3q2+7w==" (which is base64 for 0xdeadbeef) does not have an ExpectedValue in messages-encoding.json
(though it does have an expectedHexValue).

go.mod
amnonbc: The new tests won't compile without this because it includes the `any` type, which was introduced in 1.18.
If you think it is important, I could change all occurrences of `any` to `[]byte`

ably/proto_message_decoding_test.go
lmars: Ok thanks, so let me instead comment in the code, since on a first pass I was under the impression this was not how they were being used and you meant to use `require`.

ably/proto_message_decoding_test.go
lmars: Ok sorry, I didn't mean use that exact method, but perhaps have `ablytest.LoadMessageEncodingFixtures` and `ablytest.LoadMsgpackFixtures` (i.e. put all the test fixture loading code in the same place), but I understand if you don't think that's useful.

ably/proto_message_decoding_test.go
lmars: Ok I was implying we add a name to the fixture file, but I'll leave it to your judgement, happy to also leave as is.

ably/proto_message_decoding_test.go
lmars: So should this be checking `expectedHexValue`?

go.mod
lmars: Ok thanks, so I think this should be in a separate commit with that context in the commit message.

ably/proto_message_decoding_test.go
```diff
@@ -0,0 +1,147 @@
...
+	fixtures, err := loadFixtures()
+	require.NoError(t, err, "failed to load test fixtures")
+	for _, f := range fixtures {
+		t.Run(f.Data, func(t *testing.T) {
+			msg := Message{
+				Data:     f.Data,
+				Encoding: f.Encoding,
+			}
+			decodedMsg, err := msg.withDecodedData(nil)
+			assert.NoError(t, err)
```
lmars, line 49: I think this should be `require.NoError` since any further assertions seem meaningless as `decodedMsg` is a zero value.

go.mod
amnonbc: I will squash merge this anyway, so the commits will be mushed together into one commit.
Should I do a separate PR for this?

ably/proto_message_decoding_test.go
amnonbc: done in https://github.com/ably/ably-go/pull/588/commits/707c94f5163d8d1422cc6efbd0a11e6b102094d5

ably/proto_message_decoding_test.go
amnonbc: I don't really want to change  messages-encoding.json as it is being imported by other SDKs

ably/proto_message_decoding_test.go
amnonbc: I could do that if you want. But it is a trivial function, so I think it is probably better to have it not exported and defined nearer the test that uses it.

lmars: 👍

ably/proto_message_decoding_test.go
lmars: 👍

ably/proto_message_decoding_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type fixture struct {
+	Data          string `json:"data"`
+	Encoding      string `json:"encoding"`
+	ExpectedType  string `json:"expectedType"`
+	ExpectedValue any    `json:"expectedValue"`
+	ExpectedHex   []byte `json:"expectedHex"`
```
lmars, line 21: I thought it was called `expectedHexValue`?

go.mod
lmars: No separate PR no, just merge with two commits?

ably/proto_message_decoding_test.go
amnonbc: well spotted.
Fixed

lmars (approved):

AndyTWF (approved):

PR #590, opened 2023-04-03
add agents option

andydunstall:
Add an optional agents option that includes product/version key-value pairs to include in the library agents identifier, as described by RSC7d6.

Note if the OS cannot be identified, instead of an empty string this uses `unknown` to ensure the agents are always 4th in the header.

### DIFF
--- a/ably/options.go
+++ b/ably/options.go
@@ -390,6 +390,10 @@ type clientOptions struct {
 	Now   func() time.Time
 	After func(context.Context, time.Duration) <-chan time.Time
 
+	// Product/version key-value pairs to include in the agent library
+	// identifiers. This must only be used by Ably-authored SDKs (RSC7d6).
+	Agents map[string]string
+
 	// LogLevel controls the verbosity of the logs output from the library.
 	// Levels include verbose, debug, info, warn and error.
 	// platform specific (TO3b)
@@ -1067,6 +1071,14 @@ func WithLogHandler(handler Logger) ClientOption {
 func WithLogLevel(level LogLevel) ClientOption {
 	return func(os *clientOptions) {
 		os.LogLevel = level
+	}
+}
+
+// WithAgents is used to add product/version key-value pairs to include in the
+// agent library identifiers. This must only be used by Ably-authored SDKs.
+func WithAgents(agents map[string]string) ClientOption {
+	return func(os *clientOptions) {
+		os.Agents = agents
 	}
 }
 
--- a/ably/proto_http.go
+++ b/ably/proto_http.go
@@ -3,6 +3,7 @@ package ably
 import (
 	"fmt"
 	"runtime"
+	"strings"
 )
 
 // constants for rsc7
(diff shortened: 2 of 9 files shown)


lmars:
> Note if the OS cannot be identified, instead of an empty string this uses unknown to ensure the agents are always 4th in the header.

Why is it desirable for it to always be 4th?

lmars:
I think it's fine to be any position, since I believe there is an explicit parser based on [agents.json](https://github.com/ably/ably-common/blob/main/protocol/agents.json) (which you should add the new agent you're adding to, by the way).

cc @ruimffl who can confirm ^^

andydunstall:
> I think it's fine to be any position, since I believe there is an explicit parser based on [agents.json](https://github.com/ably/ably-common/blob/main/protocol/agents.json) (which you should add the new agent you're adding to, by the way).

Ok great thanks - will remove the 'unknown' and update agents.json

ably/options.go
```diff
@@ -1070,6 +1074,14 @@ func WithLogLevel(level LogLevel) ClientOption {
 	}
 }
 
+// WithAgents is used to add product/version key-value pairs to include in the
+// agent library identifiers. This must only be used by Ably-authored SDKs.
+func WithAgents(agents map[string]string) ClientOption {
+	return func(os *clientOptions) {
+		os.Agents = agents
```
sacOO7, line 1081: Instead of passing by reference, can we do `copy by value`, create a new object for agents and use it instead.

sacOO7: https://david-yappeter.medium.com/golang-pass-by-value-vs-pass-by-reference-e48aac8b2716

andydunstall: I was just keeping to the same style as the other params eg `WithTransportParams` just passes the map by reference

andydunstall:
Some tests failing, though different tests fail between runs so looks like CI is just flaky - so will mark ready to review

ably/options.go
mohyour: I think it's better to keep the same style

ably/proto_http.go
```diff
@@ -23,11 +24,20 @@ var goRuntimeIdentifier = func() string {
 	return fmt.Sprintf("%s/%s", libraryName, runtime.Version()[2:])
 }()
 
-var ablyAgentIdentifier = func() string {
+func ablyAgentIdentifier(agents map[string]string) string {
+	elems := []string{
```
mohyour, line 28: Does elems stand for something or just a variable name? Maybe a bit more descriptive

mohyour (changes requested):

ably/proto_http.go
andydunstall: Sure renamed `identifiers`

mohyour (approved):

ably/proto_http.go
```diff
@@ -23,11 +24,20 @@ var goRuntimeIdentifier = func() string {
...
-		return fmt.Sprintf("%s %s", ablySDKIdentifier, goRuntimeIdentifier)
-	} else {
-		return fmt.Sprintf("%s %s %s", ablySDKIdentifier, goRuntimeIdentifier, osIdentifier)
+	if !empty(osIdentifier) {
+		elems = append(elems, osIdentifier)
 	}
-}()
+
+	for product, version := range agents {
+		elems = append(elems, product+"/"+version)
```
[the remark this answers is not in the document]
sacOO7: @andydunstall before merging the PR, please look into this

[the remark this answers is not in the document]
andydunstall: Thanks for pointing me to that!

I've updated to ignore version if empty and tests for rest and realtime

ably/realtime_client_integration_test.go
```diff
@@ -72,6 +72,35 @@ func TestRealtime_RSC7_AblyAgent(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		client, err := ably.NewRealtime(
+			ably.WithEnvironment(ablytest.Environment),
+			ably.WithTLS(false),
+			ably.WithToken("fake:token"),
+			ably.WithUseTokenAuth(true),
+			ably.WithRealtimeHost(serverURL.Host),
+			ably.WithAgents(map[string]string{
+				"foo": "1.2.3",
```
sacOO7, line 93: Please add a test with 
```
"bar": "",
```
It should only accept product without version

sacOO7 (changes requested):
check more comments

sacOO7 (approved):
LGTM : )

PR #592, opened 2023-04-25
Feature/filtered subscription

mohyour:
We are creating a new method called GetDerived so we can create derived/synthetic channels. Presently, we only have filter as an option. The filter option lets us implement filtered subscription on a channel.
DR can be found [here](https://ably.atlassian.net/wiki/spaces/product/pages/2560098305/DR2+Filtered+Subscription+SDK+Implementation+Decision)
Ticket: [DSC-54](https://ably.atlassian.net/browse/DSC-54)

[DSC-54]: https://ably.atlassian.net/browse/DSC-54?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ably/internal/ablyutil/regex.go
+++ b/ably/internal/ablyutil/regex.go
@@ -1,0 +1,39 @@
+package ablyutil
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+)
+
+// derivedChannelMatch provides the qualifyingParam and channelName from a channel regex match for derived channels
+type derivedChannelMatch struct {
+	QualifierParam string
+	ChannelName    string
+}
+
+// This regex check is to retain existing channel params if any e.g [?rewind=1]foo to
+// [filter=xyz?rewind=1]foo. This is to keep channel compatibility around use of
+// channel params that work with derived channels.
+func MatchDerivedChannel(name string) (*derivedChannelMatch, error) {
+	regex := `^(\[([^?]*)(?:(.*))\])?(.+)$`
+	r, _ := regexp.Compile(regex)
+	match := r.FindStringSubmatch(name)
+
+	if len(match) == 0 || len(match) < 5 {
+		err := errors.New("regex match failed")
+		return nil, err
+	}
+	// Fail if there is already a channel qualifier,
+	// eg [meta]foo should fail instead of just overriding with [filter=xyz]foo
+	if len(match[2]) > 0 {
+		err := fmt.Errorf("cannot use a derived option with a %s channel", match[2])
+		return nil, err
+	}
+
+	// Return match values to be added to derive channel quantifier.
+	return &derivedChannelMatch{
+		QualifierParam: match[3],
+		ChannelName:    match[4],
+	}, nil
+}
--- a/ably/internal/ablyutil/regex.go
+++ b/ably/internal/ablyutil/regex.go
@@ -19,6 +19,7 @@ func MatchDerivedChannel(name string) (*derivedChannelMatch, error) {
 	regex := `^(\[([^?]*)(?:(.*))\])?(.+)$`
 	r, err := regexp.Compile(regex)
 	if err != nil {
+		err := errors.New("regex compilation failed")
 		return nil, err
 	}
 	match := r.FindStringSubmatch(name)
--- a/ably/internal/ablyutil/regex.go
+++ b/ably/internal/ablyutil/regex.go
@@ -17,7 +17,10 @@ type derivedChannelMatch struct {
 // channel params that work with derived channels.
 func MatchDerivedChannel(name string) (*derivedChannelMatch, error) {
 	regex := `^(\[([^?]*)(?:(.*))\])?(.+)$`
-	r, _ := regexp.Compile(regex)
+	r, err := regexp.Compile(regex)
+	if err != nil {
+		return nil, err
+	}
 	match := r.FindStringSubmatch(name)
 
 	if len(match) == 0 || len(match) < 5 {
--- a/ably/internal/ablyutil/regex_test.go
+++ b/ably/internal/ablyutil/regex_test.go
@@ -1,0 +1,35 @@
+//go:build !integration
+// +build !integration
+
+package ablyutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMatchDerivedChannel(t *testing.T) {
+	var channels = []struct {
+		name  string
+		input string
+		want  *derivedChannelMatch
+	}{
+		{"valid with base channel name", "foo", &derivedChannelMatch{QualifierParam: "", ChannelName: "foo"}},
+		{"valid with base channel namespace", "foo:bar", &derivedChannelMatch{QualifierParam: "", ChannelName: "foo:bar"}},
+		{"valid with existing qualifying option", "[?rewind=1]foo", &derivedChannelMatch{QualifierParam: "?rewind=1", ChannelName: "foo"}},
+		{"fail with wrong channel option param", "[param=1]foo", nil},
+		{"fail with invalid qualifying option", "[meta]foo", nil},
+		{"fail with invalid regex match", "[failed-match]foo", nil},
+	}
+
+	for _, tt := range channels {
+		t.Run(tt.name, func(t *testing.T) {
+			match, err := MatchDerivedChannel(tt.input)
+			if err != nil {
+				assert.Error(t, err, "An error is expected for the regex match")
+			}
+			assert.Equal(t, tt.want, match, "invalid output received")
+		})
+	}
+}
--- a/ably/internal/ablyutil/regex_test.go
+++ b/ably/internal/ablyutil/regex_test.go
@@ -18,6 +18,8 @@ func TestMatchDerivedChannel(t *testing.T) {
 		{"valid with base channel name", "foo", &derivedChannelMatch{QualifierParam: "", ChannelName: "foo"}},
 		{"valid with base channel namespace", "foo:bar", &derivedChannelMatch{QualifierParam: "", ChannelName: "foo:bar"}},
 		{"valid with existing qualifying option", "[?rewind=1]foo", &derivedChannelMatch{QualifierParam: "?rewind=1", ChannelName: "foo"}},
+		{"valid with existing qualifying option with channel namespace", "[?rewind=1]foo:bar", &derivedChannelMatch{QualifierParam: "?rewind=1", ChannelName: "foo:bar"}},
+		{"fail with invalid param with channel namespace", "[param:invalid]foo:bar", nil},
 		{"fail with wrong channel option param", "[param=1]foo", nil},
 		{"fail with invalid qualifying option", "[meta]foo", nil},
 		{"fail with invalid regex match", "[failed-match]foo", nil},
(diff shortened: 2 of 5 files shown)


ably/realtime_channel.go
```diff
@@ -114,6 +125,46 @@ func (ch *RealtimeChannels) Get(name string, options ...ChannelOption) *Realtime
...
 
+// GetDerived creates a new derived [ably.RealtimeChannel] object for given channel name, using the provided derive options and
+// channel options if any. Returns error if any occurs
+func (ch *RealtimeChannels) GetDerived(name string, deriveOptions DeriveOptions, options ...ChannelOption) (*RealtimeChannel, error) {
+	if deriveOptions.Filter != "" {
+		match, err := matchDerivedChannel(name)
+		if err != nil {
+			return nil, err
+		}
+		filter := url.PathEscape(deriveOptions.Filter)
```
andydunstall, line 136: Should this be `url.QueryEscape`? eg `url.PathEscape` doesn't encode ':' characters

ably/realtime_channel.go
```diff
@@ -114,6 +125,46 @@ func (ch *RealtimeChannels) Get(name string, options ...ChannelOption) *Realtime
...
+		name = fmt.Sprintf("[filter=%s%s]%s", filter, match.qualifierParam, match.channelName)
+	}
+	return ch.Get(name, options...), nil
+}
+
+// This regex check is to retain existing channel params if any e.g [?rewind=1]foo to
+// [filter=xyz?rewind=1]foo. This is to keep channel compatibility around use of
+// channel params that work with derived channels.
+func matchDerivedChannel(name string) (*derivedChannelMatch, error) {
```
andydunstall, line 145: Is it worth adding a unit test for this function to verify all the possible cases are handled? (qualifiers, channel params etc)

ably/realtime_channel.go
mohyour: `url.QueryEscape` causes an `invalid filter expression code` error. An alternative is to use [url.Parse](https://pkg.go.dev/net/url#URL.Parse).

ably/realtime_channel.go
```diff
@@ -50,6 +52,15 @@ type ChannelOption func(*channelOptions)
 // channelOptions wraps ChannelOptions. It exists so that users can't implement their own ChannelOption.
 type channelOptions protoChannelOptions
 
+type DeriveOptions struct {
```
owenpearson, line 55: You should add some docstrings to this public struct.

In fact, we should really be adding new SDK API descriptions to the [sdk-api-reference](https://github.com/ably/sdk-api-reference/blob/main/descriptions.md) so that they're aligned between SDKs.

owenpearson (changes requested):

ably/realtime_channel.go
andydunstall: Ah ok `QueryEscape` encodes spaces as `+`, which is valid for a URL but we can't decode it in JS to get back the original string...

Maybe for now we stick with `PathEscape` but replace `:` with `%3A`? The characters that we need to make sure are properly escaped are `:` (as its used to separate channel scopes) - and `?` and `]` (which both seem ok).

Eg using a filter 'foo == `bar:car`' will fail as an invalid channel name

In hindsight using URL encoding was probably the wrong choice (@mschristensen maybe in the production release we switch to base64 or something instead to avoid any differences between language encodings?)

ably/internal/ablyutil/regex.go
```diff
@@ -0,0 +1,39 @@
...
+	QualifierParam string
+	ChannelName    string
+}
+
+// This regex check is to retain existing channel params if any e.g [?rewind=1]foo to
+// [filter=xyz?rewind=1]foo. This is to keep channel compatibility around use of
+// channel params that work with derived channels.
+func MatchDerivedChannel(name string) (*derivedChannelMatch, error) {
+	regex := `^(\[([^?]*)(?:(.*))\])?(.+)$`
+	r, _ := regexp.Compile(regex)
```
mschristensen, line 20: We should either use `MustCompile` or check for the error rather than fail silently

ably/internal/ablyutil/regex_test.go
```diff
@@ -0,0 +1,35 @@
...
+func TestMatchDerivedChannel(t *testing.T) {
+	var channels = []struct {
+		name  string
+		input string
+		want  *derivedChannelMatch
+	}{
+		{"valid with base channel name", "foo", &derivedChannelMatch{QualifierParam: "", ChannelName: "foo"}},
+		{"valid with base channel namespace", "foo:bar", &derivedChannelMatch{QualifierParam: "", ChannelName: "foo:bar"}},
+		{"valid with existing qualifying option", "[?rewind=1]foo", &derivedChannelMatch{QualifierParam: "?rewind=1", ChannelName: "foo"}},
+		{"fail with wrong channel option param", "[param=1]foo", nil},
```
mschristensen, line 21: let's try a case with a colon in the channel name as well as a qualifier param

ably/realtime_channel_integration_test.go
```diff
@@ -93,6 +93,82 @@ func TestRealtimeChannel_Subscribe(t *testing.T) {
...
+				},
+			},
+		},
+		{
+			Name: "filtered",
+			Data: "filtered messages",
+		},
+		{
+			Name: "filtered",
+			Data: "Another unfiltered message",
```
mschristensen, line 121: This message will be filtered, since header.number != 1234

ably/realtime_channel_integration_test.go
```diff
@@ -93,6 +93,82 @@ func TestRealtimeChannel_Subscribe(t *testing.T) {
...
+	rtSub, unsub2, err := ablytest.ReceiveMessages(realtimeChannel, "filtered")
+	assert.NoError(t, err, "realtimeClient:.Subscribe(context.Background())=%v", err)
+	defer unsub2()
+
+	err = restChannel.PublishMultiple(context.Background(), msg)
+	assert.NoError(t, err, "restClient: PublishMultiple()=%v", err)
+
+	timeout := 15 * time.Second
+
+	err = expectMsg(rtDerivedSub, "filtered", msg[0].Data, timeout, true)
```
mschristensen, line 162: We need to ensure no additional messages are received on this channel

mschristensen (changes requested):

ably/realtime_channel.go
andydunstall: @mohyour if its ok with you will update update realtime to expect base64 instead which should be rolled out to sandbox next week (https://github.com/ably/realtime/pull/4835)

ably/realtime_channel_integration_test.go
mohyour: Yeah, wrong use of filtered term

ably/internal/ablyutil/regex.go
mohyour: I'll check for the error. That's better than have a panic with `MustCompile`

ably/realtime_channel.go
mschristensen: @mohyour I think we should make this change now and merge once tests pass when the realtime change goes to sandbox :) If you want to test your implementation locally, sync with @andydunstall who should be able to spin up a dev cluster with the change that you can run the tests against :)

ably/realtime_channel_integration_test.go
```diff
@@ -93,6 +93,88 @@ func TestRealtimeChannel_Subscribe(t *testing.T) {
...
+	app, err := ablytest.NewSandbox(nil)
+	assert.NoError(t, err)
+	defer app.Close()
+	options := app.Options()
+	restClient, err := ably.NewREST(options...)
+	assert.NoError(t, err)
+	msg := []*ably.Message{
+		{
+			Name: "filtered",
+			Data: "This message will be filtered",
```
mschristensen, line 106: This message will pass the filter

ably/realtime_channel_integration_test.go
```diff
@@ -93,6 +93,88 @@ func TestRealtimeChannel_Subscribe(t *testing.T) {
...
+
+	// Check message received on rtDerivedChannel
+	err = expectMsg(rtDerivedSub, "filtered", msg[0].Data, timeout, true)
+	assert.NoError(t, err)
+	// Ensure that no additional message is received on rtDerivedChannel
+	err = expectMsg(rtDerivedSub, "filtered", "incorrect data", timeout, true)
+	assert.Error(t, err)
+	assert.Equal(t, err, fmt.Errorf("waiting for message name=%q data=%q timed out after %v", "filtered", "incorrect data", timeout))
+	// Check messages received on realtimeChannel
+	err = expectMsg(rtSub, "filtered", msg[0].Data, timeout, true)
```
mschristensen, line 170: Rather than waiting 15s for this test to pass, could we use an "end" message like we do for the js tests?

ably/internal/ablyutil/regex.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// This regex check is to retain existing channel params if any e.g [?rewind=1]foo to
+// [filter=xyz?rewind=1]foo. This is to keep channel compatibility around use of
+// channel params that work with derived channels.
+func MatchDerivedChannel(name string) (*derivedChannelMatch, error) {
+	regex := `^(\[([^?]*)(?:(.*))\])?(.+)$`
+	r, err := regexp.Compile(regex)
+	if err != nil {
+		return nil, err
```
mschristensen, line 22: we should wrap the error

mschristensen (changes requested):

ably/realtime_channel_integration_test.go
```diff
@@ -93,6 +93,98 @@ func TestRealtimeChannel_Subscribe(t *testing.T) {
...
+	assert.NoError(t, err)
+	defer unsub()
+
+	err = restChannel.PublishMultiple(context.Background(), testMessages)
+	assert.NoError(t, err, "restClient: PublishMultiple()=%v", err)
+
+	for i := 0; i < 1; i++ {
+		select {
+		case m := <-filteredMessages:
+			assert.Equal(t, "filtered", m.Name)
```
mschristensen, line 173: This doesn't tell us that there are not more messages that could be received on this channel, as we only check the first message and then stop looking for more

mschristensen (changes requested):
nice, this is getting close!

The main problem is that more messages could be subsequently received on the channels which we wouldn't catch.

We could execute the two for loops in a goroutine and use context cancellation to check when an "end" message is received, something like:

```go
	ctx, cancel := context.WithCancel(ctx)
	var wg sync.WaitGroup
	wg.Add(1)
	go func(ctx context.Context) {
		defer wg.Done()
		for {
			select {
			case m := <-filteredMessages:
				assert.Equal(t, "filtered", m.Name)
			case <-ctx.Done():
				return
			}
		}
	}(ctx)
	wg.Add(1)
	go func(ctx context.Context) {
		defer cancel()
		defer wg.Done()
		for {
			select {
			case m := <-unfilteredMessages:
				if m.Name == "end" {
					return
				}
				assert.Equal(t, "filtered", m.Name)
			case <-ctx.Done():
				return
			}
		}
	}(ctx)
	wg.Wait()
```

mschristensen (approved):
LGTM, nice work!

andydunstall (approved):
LGTM

PR #596, opened 2023-06-01
add publish async method

andydunstall:
Adds a `PublishAsync` method by replacing `chan error` with a callback

See https://ably-real-time.slack.com/archives/C0114KK7S6R/p1685622657245009 for context

Depends on https://github.com/ably/ably-go/pull/595

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,9 +341,6 @@ As of release 1.2.0, the following are not implemented and will be covered in fu
   connection becomes `suspended` and then resumes, and presence members associated with the client will not be
   automatically re-entered.
 
-- Transient realtime publishing is not supported, so a call to `publish()` on a realtime channel will trigger attachment
-  of the channel.
-
 - Inband reauthentication is not supported; expiring tokens will trigger a disconnection and resume of a realtime
   connection.
 
--- a/README.md
+++ b/README.md
@@ -93,6 +93,23 @@ if err != nil {
 }
 ```
 
+`Publish` will block until either the publish is acknowledged or failed to
+deliver.
+
+Alternatively you can use `PublishAsync` which does not block:
+
+```go
+channel.PublishAsync("EventName1", "EventData11", func(err error) {
+	if err != nil {
+		fmt.Println("failed to publish", err)
+	} else {
+		fmt.Println("publish ok")
+	}
+})
+```
+
+Note the `onAck` callback must not block as it would block the internal client.
+
 #### Handling errors
 
 Errors returned by this library may have an underlying `*ErrorInfo` type.
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -585,8 +585,22 @@ func (em ChannelEventEmitter) OffAll() {
 }
 
 // Publish publishes a single message to the channel with the given event name and message payload.
+//
+// This will block until either the publish is acknowledged or fails to deliver.
+//
+// If the context is cancelled before the attach operation finishes, the call
+// returns an error but the publish will carry on in the background and may
+// eventually be published anyway.
 func (c *RealtimeChannel) Publish(ctx context.Context, name string, data interface{}) error {
 	return c.PublishMultiple(ctx, []*Message{{Name: name, Data: data}})
+}
+
+// PublishAsync is the same as Publish except instead of blocking it calls onAck
+// with nil if the publish was successful or the appropriate error.
+//
+// Note onAck must not block as it would block the internal client.
+func (c *RealtimeChannel) PublishAsync(name string, data interface{}, onAck func(err error)) error {
+	return c.PublishMultipleAsync([]*Message{{Name: name, Data: data}}, onAck)
 }
 
 // PublishMultiple publishes all given messages on the channel at once.
@@ -595,6 +609,25 @@ func (c *RealtimeChannel) Publish(ctx context.Context, name string, data interfa
 // returns an error but the publish will carry on in the background and may
 // eventually be published anyway.
 func (c *RealtimeChannel) PublishMultiple(ctx context.Context, messages []*Message) error {
+	listen := make(chan error, 1)
+	onAck := func(err error) {
+		listen <- err
+	}
+	if err := c.PublishMultipleAsync(messages, onAck); err != nil {
+		return err
+	}
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-listen:
+		return err
+	}
+}
+
+// PublishMultipleAsync is the same as PublishMultiple except it calls onAck instead of blocking
+// (see PublishAsync).
+func (c *RealtimeChannel) PublishMultipleAsync(messages []*Message, onAck func(err error)) error {
 	id := c.client.Auth.clientIDForCheck()
 	for _, v := range messages {
 		if v.ClientID != "" && id != wildcardClientID && v.ClientID != id {
@@ -607,11 +640,7 @@ func (c *RealtimeChannel) PublishMultiple(ctx context.Context, messages []*Messa
 		Channel:  c.Name,
 		Messages: messages,
 	}
-	res, err := c.send(msg)
-	if err != nil {
-		return err
-	}
-	return res.Wait(ctx)
+	return c.send(msg, onAck)
 }
 
 // History retrieves a [ably.HistoryRequest] object, containing an array of historical
@@ -624,35 +653,34 @@ func (c *RealtimeChannel) History(o ...HistoryOption) HistoryRequest {
 	return c.client.rest.Channels.Get(c.Name).History(o...)
 }
 
-func (c *RealtimeChannel) send(msg *protocolMessage) (result, error) {
-	if res, enqueued := c.maybeEnqueue(msg); enqueued {
-		return res, nil
+func (c *RealtimeChannel) send(msg *protocolMessage, onAck func(err error)) error {
+	if enqueued := c.maybeEnqueue(msg, onAck); enqueued {
+		return nil
 	}
 
 	if !c.canSend() {
-		return nil, newError(ErrChannelOperationFailedInvalidChannelState, nil)
+		return newError(ErrChannelOperationFailedInvalidChannelState, nil)
 	}
 
-	res, listen := newErrResult()
-	c.client.Connection.send(msg, listen)
-	return res, nil
+	c.client.Connection.send(msg, onAck)
+	return nil
 }
 
-func (c *RealtimeChannel) maybeEnqueue(msg *protocolMessage) (_ result, enqueued bool) {
+func (c *RealtimeChannel) maybeEnqueue(msg *protocolMessage, onAck func(err error)) bool {
 	// RTL6c2
 	if c.opts().NoQueueing {
-		return nil, false
+		return false
 	}
 	switch c.client.Connection.State() {
 	default:
-		return nil, false
+		return false
 	case ConnectionStateInitialized,
 		ConnectionStateConnecting,
 		ConnectionStateDisconnected:
 	}
 	switch c.State() {
 	default:
-		return nil, false
+		return false
 	case ChannelStateInitialized,
 		ChannelStateAttached,
 		ChannelStateDetached,
@@ -660,9 +688,8 @@ func (c *RealtimeChannel) maybeEnqueue(msg *protocolMessage) (_ result, enqueued
 		ChannelStateDetaching:
 	}
 
-	res, listen := newErrResult()
-	c.queue.Enqueue(msg, listen)
-	return res, true
+	c.queue.Enqueue(msg, onAck)
+	return true
 }
 
 func (c *RealtimeChannel) canSend() bool {
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -585,24 +585,15 @@ func (em ChannelEventEmitter) OffAll() {
 }
 
 // Publish publishes a single message to the channel with the given event name and message payload.
-// A callback may optionally be passed in to this call to be notified of success or failure of the operation.
-// When publish is called with this client library, it won't attempt to implicitly attach to the channel,
-// so long as transient publishing is available in the library. Otherwise, the client will implicitly attach (RTL6i).
-//
-// This implicitly attaches the channel if it's not already attached. If the
-// context is canceled before the attach operation finishes, the call
-// returns with an error, but the operation carries on in the background and
-// the channel may eventually be attached and the message published anyway.
 func (c *RealtimeChannel) Publish(ctx context.Context, name string, data interface{}) error {
 	return c.PublishMultiple(ctx, []*Message{{Name: name, Data: data}})
 }
 
 // PublishMultiple publishes all given messages on the channel at once.
 //
-// This implicitly attaches the channel if it's not already attached. If the
-// context is canceled before the attach operation finishes, the call
-// returns with an error, but the operation carries on in the background and
-// the channel may eventually be attached and the message published anyway (RTL6i).
+// If the context is cancelled before the attach operation finishes, the call
+// returns an error but the publish will carry on in the background and may
+// eventually be published anyway.
 func (c *RealtimeChannel) PublishMultiple(ctx context.Context, messages []*Message) error {
 	id := c.client.Auth.clientIDForCheck()
 	for _, v := range messages {
--- a/ably/realtime_channel_integration_test.go
+++ b/ably/realtime_channel_integration_test.go
@@ -49,6 +49,20 @@ func TestRealtimeChannel_Publish(t *testing.T) {
 	assert.NoError(t, err, "Publish()=%v", err)
 }
 
+func TestRealtimeChannel_PublishAsync(t *testing.T) {
+	app, client := ablytest.NewRealtime(nil...)
+	defer safeclose(t, ablytest.FullRealtimeCloser(client), app)
+	err := ablytest.Wait(ablytest.ConnWaiter(client, client.Connect, ably.ConnectionEventConnected), nil)
+	assert.NoError(t, err)
+	channel := client.Channels.Get("test")
+	listen := make(chan error, 1)
+	err = channel.PublishAsync("hello", "world", func(err error) {
+		listen <- err
+	})
+	assert.NoError(t, err, "PublishAsync()=%v", err)
+	assert.NoError(t, <-listen, "onAck=%v", err)
+}
+
 func TestRealtimeChannel_Subscribe(t *testing.T) {
 	app, client1 := ablytest.NewRealtime(nil...)
 	defer safeclose(t, ablytest.FullRealtimeCloser(client1), app)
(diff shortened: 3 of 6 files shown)


ably/realtime_channel.go
```diff
@@ -633,45 +649,43 @@ func (c *RealtimeChannel) History(o ...HistoryOption) HistoryRequest {
...
 
 	if !c.canSend() {
-		return nil, newError(ErrChannelOperationFailedInvalidChannelState, nil)
+		return newError(ErrChannelOperationFailedInvalidChannelState, nil)
 	}
 
-	res, listen := newErrResult()
-	c.client.Connection.send(msg, listen)
-	return res, nil
+	c.client.Connection.send(msg, onAck)
```
sacOO7, line 661: i don't think we don't need to send onAck as an argument here.

sacOO7:
@andydunstall I think we just need to modify small piece of code here 
```

func (c *RealtimeChannel) PublishAsync(name string, data interface{}, ack func(error)) error {
	return c.PublishMultipleAsync([]*Message{{Name: name, Data: data}}, ack)
}

func (c *RealtimeChannel) PublishMultipleAsync(messages []*Message, ack func(error)) error {
	id := c.client.Auth.clientIDForCheck()
	for _, v := range messages {
		if v.ClientID != "" && id != wildcardClientID && v.ClientID != id {
			// Spec RTL6g3,RTL6g4
			return fmt.Errorf("Unable to publish message containing a clientId (%s) that is incompatible with the library clientId (%s)", v.ClientID, id)
		}
	}
	msg := &protocolMessage{
		Action:   actionMessage,
		Channel:  c.Name,
		Messages: messages,
	}
	res, err := c.send(msg)
	if err != nil {
		return err
	}
	errResult := res.(*errResult)
	if errResult == nil {
		ack(nil)
		return nil
	}
	go func() {
		if listenChannel := errResult.listen; listenChannel != nil {
			errResult.listen = nil
			// TODO : Add one more case with timeout to avoid dangling coroutine
			select {
				case errResult.err = <-listenChannel:
			}
			ack(errResult.err)
		}
	}()
	return nil
}
```

Not sure why we need to make so many changes at different places: (

andydunstall:
@sacOO7 I was trying to avoid a goroutine per publish - which seems very inefficient?

sacOO7:
I don't think you can avoid goroutines 🤔

sacOO7:
Based on what's written inside onAck function, it can block internal queue

sacOO7:
I get what you are trying to do. Based on what I read every goroutine consumes around 2KB of stack space. If we are sending one message at a time with the rate of 2k msg/second, it can spawn tons of goroutines waiting for ack. I think a better solution would be to implement goroutine pool that is responsible for calling ack functions.

sacOO7:
@andydunstall There is a lightweight golang workerpool library.
https://github.com/gammazero/workerpool
We can either install this or just include the file https://github.com/gammazero/workerpool/blob/master/workerpool.go

andydunstall:
> it can block internal queue/ msg processing workflow or delay it based on the time it takes to execute onAck method

Could you elaborate? The `onAck` writes an error to a `chan error` of length 1 - which is exactly the same as the behaviour before?

sacOO7:
> > it can block internal queue/ msg processing workflow or delay it based on the time it takes to execute onAck method
> 
> Could you elaborate? The `onAck` writes an error to a `chan error` of length 1 - which is exactly the same as the behaviour before?
> 
> I still don't think we need goroutines - even if we use a pool to avoid being inefficient - won't that just really hurt throughput if each publish has to wait for a free goroutine from a pool
> 
> Plus goroutines seem far more complex and likely to cause issues than just using an `onAck` callback

We are talking about impl for `publishAsync` and not `publish` right?

sacOO7:
The way I see it. `publishAsync` accepts `onAck` as an extra param, which can have any type of implementation that can lead to a blocking nature.
The original `publish` method is anyway blocking and we don't have issues with it.

sacOO7:
> > it can block internal queue/ msg processing workflow or delay it based on the time it takes to execute onAck method
> 
> Could you elaborate? The `onAck` writes an error to a `chan error` of length 1 - which is exactly the same as the behaviour before?
> 
> I still don't think we need goroutines - even if we use a pool to avoid being inefficient - won't that just really hurt throughput if each publish has to wait for a free goroutine from a pool
> 
> Plus goroutines seem far more complex and likely to cause issues than just using an `onAck` callback

The reason we didn't have callbacks till now was to avoid the use of goroutines : (
Since, we are introducing it, we have to use them properly as I suggested earlier,

sacOO7:
> 

Or simplest thing you can do is, maintain a list of all listen channels for all published messages, and execute them one by one under one goroutine.
The only issue would be if one of the channel is taking too much time to get ack, other's have to wait for it. That's why I was suggesting use of goroutine worker pool with limited goroutines and it's completely safe : )

andydunstall:
> which can have any type of implementation that can lead to a blocking nature for internal workflow

Ah ok so you want to handle the users that `onAck` callback blocks? Seems like we could just tell the user to make sure `onAck` doesn't block?

Given we're just adding this for the database connector - where we know `onAck` won't block - 
it might be better if we just vendor the go SDK within the database connector that we can patch instead of trying to merge into main. Trying to add goroutine pools seems like it would add a lot of complexity which isn't worth doing just for us to use `PublishAsync` in the connector (and could hurt throughput and resource usage) cc @mschristensen

sacOO7:
> > which can have any type of implementation that can lead to a blocking nature for internal workflow
> 
> Ah ok so you want to handle a users `onAck` callback that blocks? Seems like we could just tell the user to make sure `onAck` doesn't block?
> 
> Given we're just adding this for the database connector - where we know `onAck` won't block - it might be better if we just vendor the go SDK within the database connector that we can patch instead of trying to merge into main. Trying to add goroutine pools seems like it would add a lot of complexity which isn't worth doing just for us to use `PublishAsync` in the connector (and could hurt throughput and resource usage) cc @mschristensen

I think it would be great if you can test performance with both cases. Sending 5k msg/second and check how ack will exactly work, how much time it takes for processing acks in both cases. The reason we are adding pool is to reduce resource usage and increase throughput, and I don't think it's that complex 🙂.

Anyways, in async-await, all of the tasks are run on a worker pool.

lmars:
The contentious thing here seems to be how to provide ACKs back to the caller when they aren't blocked in a publish call.

The callback approach seems reasonable, although if we don't like that because it can lead to internally blocking the client, rather than a goroutine pool, what if the client exposed an ACK queue that the caller could read from separate to the publish call? That queue can be designed to be non-blocking, then the client never blocks appending an ACK to that queue, and the caller can process that queue however they please (e.g. either using a goroutine pool, or in a single goroutine)?

sacOO7:
@lmars if you can provide sample code implementation for above solution, it will be helpful to get the full picture.

lmars:
@sacOO7 I won't provide a complete implementation, but consider a queue which implements the following interface:

```go
type AckQueue interface {
	// push an ACK message to the queue (non-blocking)
	Push(ack *Ack)

	// read from this channel to wait for some ACKs to process
	Ready() <-chan struct{}

	// call this to iterate to the next ACK, and remove the previous one
	Next() bool

	// call this to get the current ACK
	Ack() *Ack
}
```

and the user processing those ACKs in a separate goroutine:

```go
for {
	select {
	case <-ackQueue.Ready():
		for ackQueue.Next() {
			ack := ackQueue.Ack()
			// process the ACK
		}
	case <-ctx.Done():
		return ctx.Err()
	}
}

```

This could equally be an internal implementation whereby the callbacks passed to `PublishAsync` are pushed onto a queue like this, and a single internal goroutine processes that queue by calling the callback for whatever message got ACKed.

andydunstall:
Using a callback seems to be consistent with the other SDKs? (ably-java, ably-js etc) I'm not sure if ably-java would block the publish if the listener blocks

A non-blocking queue still seems like it would be complex to implement, both within the SDK and for the client

Given the purpose of this is just to improve the database connector throughput, which using a callback is sufficient for given we can make sure it doesn't block, I'm worried about spending too much time blocked trying to make this into a generic method we can merge into main when that isn't a requirement for the connector

So I'm leaning towards just vendoring in the connector and adding a callback based `PublishAsync` but not merge into the SDK - then closing this PR

mschristensen:
It seems to me that any concerns around the onAck callback blocking the client can be mitigated with clear documentation about this behaviour on the method to prevent any misuse?

It sounds like we're all broadly happy the the API, so could we consider this implementation good enough with the above documentation in place? I like Lewis's idea of an internal ack queue, which we could then use to improve this implementation in a subsequent milestone.

andydunstall:
> I'm happy with the callback approach with documentation that it should avoid doing anything blocking 👍

Ok great thanks

@owenpearson is this approach ok with you?

owenpearson:
@andydunstall seems reasonable to me 👍

ably/realtime_channel.go
```diff
@@ -585,25 +585,45 @@ func (em ChannelEventEmitter) OffAll() {
...
+	case err := <-listen:
+		return err
+	}
+}
+
+// PublishMultipleAsync is the same as PublishMultiple except instead of blocking it calls onAck after the
+// message is either acknowledged or fails to publish.
+
+// PublishMultipleAsync is the same as PublishMultiple except it calls onAck instead of blocking
+// (see PublishAsync).
```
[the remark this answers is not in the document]
andydunstall: oops remove previous

lmars (approved):

sacOO7:
Till then can you please README with new method signature

ably/realtime_channel.go
```diff
@@ -585,25 +585,42 @@ func (em ChannelEventEmitter) OffAll() {
...
-// context is canceled before the attach operation finishes, the call
-// returns with an error, but the operation carries on in the background and
-// the channel may eventually be attached and the message published anyway.
 func (c *RealtimeChannel) Publish(ctx context.Context, name string, data interface{}) error {
 	return c.PublishMultiple(ctx, []*Message{{Name: name, Data: data}})
 }
 
+// PublishAsync is the same as Publish except instead of blocking it calls onAck after either:
+// * The message is successfully acknowledged so onAck is called with a nil error
+// * The publish fails so onAck is called with the appropriate error
```
sacOO7, line 594: - PublishAsync is a non-blocking Publish that accepts `onAck` callback fn as an extra arg,
- onAck is called with error as nil/non-nil specifying success/failure respectively

ably/realtime_channel.go
```diff
@@ -585,25 +585,42 @@ func (em ChannelEventEmitter) OffAll() {
...
 // Publish publishes a single message to the channel with the given event name and message payload.
-// A callback may optionally be passed in to this call to be notified of success or failure of the operation.
-// When publish is called with this client library, it won't attempt to implicitly attach to the channel,
-// so long as transient publishing is available in the library. Otherwise, the client will implicitly attach (RTL6i).
-//
-// This implicitly attaches the channel if it's not already attached. If the
-// context is canceled before the attach operation finishes, the call
-// returns with an error, but the operation carries on in the background and
-// the channel may eventually be attached and the message published anyway.
 func (c *RealtimeChannel) Publish(ctx context.Context, name string, data interface{}) error {
```
sacOO7, line 588: Publish publishes a single message to the channel with the given event name and message payload
Publish call remains blocked till the message is published, it returns an error with nil/non-nil value representing success/failure respectively.

sacOO7: Can you update the same doc for `PublishMultiple` ?

ably/realtime_channel.go
```diff
@@ -585,25 +585,42 @@ func (em ChannelEventEmitter) OffAll() {
...
 	return c.PublishMultiple(ctx, []*Message{{Name: name, Data: data}})
 }
 
+// PublishAsync is the same as Publish except instead of blocking it calls onAck after either:
+// * The message is successfully acknowledged so onAck is called with a nil error
+// * The publish fails so onAck is called with the appropriate error
+func (c *RealtimeChannel) PublishAsync(ctx context.Context, name string, data interface{}, onAck func(err error)) error {
```
sacOO7, line 595: No need to pass `ctx` as argument

ably/realtime_channel.go
sacOO7: Can you update doc for the same for `publishMultipleAsync`?

ably/realtime_channel.go
sacOO7: Resolving the comment since we are accepting this implementation

sacOO7 (changes requested):
Can you update README for `publishAsync` and `publishMultipleAsync` 
specifying blocking nature of the onAck and mentioning proper example to avoid it. 
I think channels can be used inside `onAck` in the example.

ably/realtime_channel.go
```diff
@@ -585,25 +585,42 @@ func (em ChannelEventEmitter) OffAll() {
...
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-listen:
+		return err
+	}
+}
+
+// PublishMultipleAsync is the same as PublishMultiple except it calls onAck instead of blocking
+// (see PublishAsync).
+func (c *RealtimeChannel) PublishMultipleAsync(ctx context.Context, messages []*Message, onAck func(err error)) error {
```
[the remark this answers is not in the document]
andydunstall: oops thanks - fixed

ably/realtime_channel.go
andydunstall: How about
```
// Publish publishes a single message to the channel with the given event name and message payload.
//
// This will block until either the publish is acknowledged or fails to deliver.
```

> it returns an error with nil/non-nil value representing success/failure respectively.

That seems t bit redundant?

ably/realtime_channel.go
andydunstall: How about
```
// PublishAsync is the same as Publish except instead of blocking it calls onAck
// with nil if the publish was successful or an appropriate error.
```

andydunstall:
> Can you update README for publishAsync and publishMultipleAsync
specifying blocking nature of the onAck and mentioning proper example to avoid it.
I think channels can be used inside onAck in the example.

I've added a note about `PublishAsync`. It doesn't look like the README currently gives many examples beyond just printing a result - so given theres the godoc assuming just mentioning it exists is sufficient (the method comment says `onAck` must not block)

ably/realtime_channel.go
sacOO7: // Publish publishes a single message to the channel with the given event name and message payload.
// This will block until either the publish is acknowledged (returns error as nil) or fails to deliver ( returns error value )
//If the context is cancelled before the attach operation finishes, the call returns an error but the publish will carry on in the background and may eventually be published anyway.
WDYT?

README.md
```diff
@@ -93,6 +93,9 @@ if err != nil {
 }
 ```
 
+`Publish` will block until either the publish is acknowledged or failed to
+deliver. Alternatively you can use `PublishAsync` which does not block.
```
sacOO7, line 97: I think it would be great if you can add an example to publish asynchronously. It will be great if the example is visible.
for `publishMultipleAsync`, you can just add a note below it.

ably/realtime_channel.go
sacOO7: // PublishAsync is the same as Publish except instead of blocking it calls onAck
// with nil if the publish was successful or with error value if publish fails.
//
// Important Note - code inside onAck will block internal client, it should be written in a way that it shouldn't be blocking and shouldn't take too much time to execute.

README.md
andydunstall: > devs who want to make sure message was published without using normal blockingpublish method

This seems very application specific?

In the connector we're just using it to monitor the number of failed messages
```
channel.PublishAsync(..., func(err error) {
  if (err != nil) {
    metrics.FailedPublishes.Inc()
  }
})
```

Could add the equivalent of the `Publish` example to just print the failed publish but that seems a bit redundant?

ably/realtime_channel.go
andydunstall: sure updated

README.md
sacOO7: yeah, I think you add example 
channel.PublishAsync(..., func(err error) {
  if (err != nil) {
    fmt.print("publish failed with err", err);
  } else {
      fmt.print("publish success")
    }
})

ably/realtime_channel.go
```diff
@@ -585,25 +585,49 @@ func (em ChannelEventEmitter) OffAll() {
...
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-listen:
+		return err
+	}
+}
+
+// PublishMultipleAsync is the same as PublishMultiple except it calls onAck instead of blocking
```
sacOO7, line 628: `PublishMultipleAsync` is the same as `PublishAsync`, only different is, multiple messages can be published.

ably/realtime_channel.go
```diff
@@ -585,25 +585,49 @@ func (em ChannelEventEmitter) OffAll() {
...
+// returns an error but the publish will carry on in the background and may
+// eventually be published anyway.
 func (c *RealtimeChannel) Publish(ctx context.Context, name string, data interface{}) error {
 	return c.PublishMultiple(ctx, []*Message{{Name: name, Data: data}})
 }
 
+// PublishAsync is the same as Publish except instead of blocking it calls onAck
+// with nil if the publish was successful or the appropriate error.
+//
+// Note onAck must not block as it will block the event loop.
```
sacOO7, line 601: I think lot of people will not understand `event loop`, still it would be better if we make it clear in README as a part of example. We can just say, onAck will block the internal client depending on the implementation inside it.

ably/realtime_channel.go
andydunstall: I think we're bikeshedding a bit here slightly rewording comments?

andydunstall:
@sacOO7 I've added that example to the readme and re-worked the comments a few times - are you ok for this to be merged now?

ably/realtime_channel.go
andydunstall: ok

README.md
```diff
@@ -93,6 +93,21 @@ if err != nil {
...
+```go
+channel.PublishAsync("EventName1", "EventData11", func(err error) {
+	if err != nil {
+		fmt.Println("failed to publish", err)
+	} else {
+		fmt.Println("publish ok")
+	}
+})
+```
+
```
sacOO7, line 110: You can add a final note -
```
Note - In the case of PublishAsync, make sure implementation inside `onAck` doesn't block the internal client. 
```

README.md
```diff
@@ -93,6 +93,21 @@ if err != nil {
 }
 ```
 
+`Publish` will block until either the publish is acknowledged or failed to
+deliver.
+
+Alternatively you can use `PublishAsync` which does not block:
```
sacOO7, line 99: Alternatively, you can use `PublishAsync, which doesn't wait for `publish` to finish.

sacOO7 (approved):
LGTM : )

README.md
andydunstall: Done

PR #600, opened 2023-06-09
Release/1.2.12

mschristensen:


owenpearson (changes requested):
I think we should editorialise the changelog entries a bit here:
- "Integration/subscription filters" is a bit cryptic
- Some of these entries aren't user facing so could be removed (eg. changing godocs base-url, msgpack test fixtures)
- Dropping support for go 1.18 should probably be closer to the top since users will probably want to be aware of that.

I'd actually be in favour of removing the "Closed issues" section since everything there is duplicated in "Merged pull requests"

Additionally, is it worth adding the same paragraph that we added to top of the [ably-js release](https://github.com/ably/ably-js/releases/tag/1.2.40)?

mschristensen:
Thanks @owenpearson, updated. Will squash these last 2 commits before merge.

owenpearson (approved):

PR #603, opened 2023-06-22
channel: add realtime encryption

andydunstall:
@owenpearson @sacOO7 we're looking to support encryption with in the database connector, though AFAICT encryption isn't supported for realtime channels

Looking at rest channels, looks like its just doing:
```
		cipher, _ := options.GetCipher()
		var err error
		*m, err = (*m).withEncodedData(cipher)
		if err != nil {
			return fmt.Errorf("encoding data for message #%d: %w", i, err)
		}
```

Though I know why `withEncodedData` isn't run on non-rest channels?

Just adding draft PR to start discussion - this just copies `withEncodedData` into `RealtimeChannel`

### DIFF
--- a/ably/realtime_channel.go
+++ b/ably/realtime_channel.go
@@ -629,10 +629,18 @@ func (c *RealtimeChannel) PublishMultiple(ctx context.Context, messages []*Messa
 // (see PublishAsync).
 func (c *RealtimeChannel) PublishMultipleAsync(messages []*Message, onAck func(err error)) error {
 	id := c.client.Auth.clientIDForCheck()
-	for _, v := range messages {
-		if v.ClientID != "" && id != wildcardClientID && v.ClientID != id {
+	for i, m := range messages {
+		if m.ClientID != "" && id != wildcardClientID && m.ClientID != id {
 			// Spec RTL6g3,RTL6g4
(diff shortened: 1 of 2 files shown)


sacOO7:
Need to check on how encryption works in rest channels.

andydunstall:
> You can check and implement if it's easy, we will review.

Thanks - thats what this PR is :)

owenpearson (commented):
Looks good so far, although I'm hesitant to approve because I don't think we're doing enough to test this at the moment. The test added here would pass even if the library just ignores the cipher param entirely. At a minimum, could we add a test which creates two realtimes with a cipher mismatch and checks that the behaviour is correct (see [ably-js example](https://github.com/ably/ably-js/blob/3915bcd16306bcad1b6ab72b04252be53d892a2a/test/realtime/crypto.test.js#L635-L696). It would also be good to test the behaviour when sending an encrypted message to a client without a cipher and vice versa.

andydunstall:
> test the behaviour when sending an encrypted message to a client without a cipher and vice versa

Sure, extended the test to include a channel without a cipher, the that the message doesn't equal the original

andydunstall:
@owenpearson @sacOO7 can I get a review on this?

owenpearson (approved):
LGTM, thanks

andydunstall:
@owenpearson Thanks can this be merged?

sacOO7 (approved):
LGTM

PR #604, opened 2023-06-27
Fix README for ablytest package usage

### LINKED ISSUE #602: Update README for usage of ablytest package
- Updated README as per comments https://github.com/ably/ably-go/issues/601#issuecomment-1597117245

sacOO7:
Fixed #602

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,6 +309,15 @@ if err != nil {
 }
 fmt.Print(status, status.ChannelId)
 ```
+## Note on ablytest package
+Although `ablytest` package ia available as a part of `ably`, we do not recommend using it as a sandbox for your own testing, since it's specifically intended for client library SDKs.
+It can lead to flaky nature of tests, considering some beta features can be deployed on the `sandbox` environment so that they can be tested before going into production.
+
+You should rather use, `ablyRealtimeClient` by passing the `ABLY_KEY`, which would be using a more stable ably prod environment.
+
+```
+client, err := ably.NewRealtime(ably.WithKey("xxx:xxx"))
+```
 
 ## Resources
 
--- a/README.md
+++ b/README.md
@@ -309,8 +309,8 @@ if err != nil {
 }
 fmt.Print(status, status.ChannelId)
 ```
-## Note on ablytest package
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -309,6 +309,17 @@ if err != nil {
 }
 fmt.Print(status, status.ChannelId)
 ```
+## Note on usage of ablytest package
+Although `ablytest` package is available as a part of `ably`, we do not recommend using it as a sandbox for your own testing, since it's specifically intended for client library SDKs and we don’t provide any guarantees for support or that it will remain publicly accessible.
```
owenpearson, line 313:
```suggestion
Although the `ablytest` package is available as a part of ably-go, we do not recommend using it as a sandbox for your own testing, since it's specifically intended for client library SDKs and we don’t provide any guarantees for support or that it will remain publicly accessible.
```

README.md
```diff
@@ -309,6 +309,17 @@ if err != nil {
...
+It can lead to flaky nature of tests, considering some beta features can be deployed on the `sandbox` environment so that they can be tested before going into production.
```
owenpearson, line 314:
```suggestion
It can lead to unexpected behaviour, since some beta features may be deployed on the `sandbox` environment so that they can be tested before going into production.
```

README.md
```diff
@@ -309,6 +309,17 @@ if err != nil {
...
+
+You should rather use, `ablyRealtimeClient` by passing the `ABLY_KEY`, which would be using a more stable ably prod environment.
```
owenpearson, line 316:
```suggestion
You should rather use, `ably.NewRealtime` by passing the `ABLY_KEY`, which would be using the Ably production environment.
```

README.md
```diff
@@ -309,6 +309,17 @@ if err != nil {
...
+
+```
+client, err := ably.NewRealtime(ably.WithKey("xxx:xxx"))
+```
+
+You can also use a [control api](https://ably.com/docs/control-api) to setup test environment using https://github.com/ably/ably-control-go/.
```
owenpearson, line 322:
```suggestion
You can also use the [control api](https://ably.com/docs/control-api) to setup a test environment using https://github.com/ably/ably-control-go/.
```

owenpearson (approved):

PR #614, opened 2023-07-30
[SDK-3731] Fix websocket read limits

### LINKED ISSUE #610: Allow configuring ably/websocket read limits
The messages we are receiving through the Ably real-time client are somewhat large (seem to max out ~175k), but the default real limit for messages in `nhooyr.io/websocket` is 32,678 bytes. If a message is received larger than this limit, the connection is closed.

Ideally, we'd be able to set this message limit in the `ClientOptions`, but this seems like an implementation-specific configuration and may not apply to other connection mechanisms. 

I also see that the `Dial` is a client option. I'd like to use all of the `ably.websocketConn` (but increase the read limit) without needing to implement my own websocket connection/client functionality.

I'm happy to do some work here too and open a PR, but I want to make sure that 1) this is a change that would be welcome in the package and 2) the option I implement is the right path.

Thanks so much!!

-- Chris

sacOO7:
Fixed #610

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,11 +196,11 @@ if err != nil {
 }
 ```
 
-#### Update MaxMessageSize/read limit for realtime connection
-- Connection MaxMessageSize limit is automatically configured based on `ConnectionDetails.MaxMessageSize` property received when connection is successfully made with ably.
-- This value defaults to 65536(64kb) for normal account, you can increase the limit by raising a ticket at [ably support](https://ably.com/support).
-- Upgrading account to higher limit should automatically update `ConnectionDetails.MaxMessageSize` property and should accordingly set the client side connection readlimit.
-- If you are still facing issues while reading bigger size messages or intentionally want to reduce the limit, you can explicitly update the connection read limit.
+#### Update MaxMessageSize/read limit for realtime message subscription
+- The default `MaxMessageSize` is automatically configured by Ably when connection is established with Ably.
+- This value defaults to 64kb, please [get in touch](https://ably.com/support) if you would like to request a higher limit for your account.
+- Upgrading your account to higher limit will automatically update `MaxMessageSize` property and should accordingly set the client side connection read limit.
+- If you are still facing issues when receiving large messages or intentionally want to reduce the limit, you can explicitly update the connection read limit:
 
 ```go
 client, err = ably.NewRealtime(ably.WithKey("xxx:xxx"))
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -196,6 +196,22 @@ if err != nil {
 }
 ```
 
+#### Update MaxMessageSize/read limit for realtime connection
```
owenpearson, line 199:
```suggestion
#### Update MaxMessageSize/read limit for realtime message subscription
```

README.md
```diff
@@ -196,6 +196,22 @@ if err != nil {
...
+- Connection MaxMessageSize limit is automatically configured based on `ConnectionDetails.MaxMessageSize` property received when connection is successfully made with ably.
```
owenpearson, line 200: I don't think we need to talk about `ConnectionDetails` here, it would be simpler to just say this:
```suggestion
- The default `MaxMessageSize` is automatically configured by Ably when connection is established with Ably.
```

README.md
```diff
@@ -196,6 +196,22 @@ if err != nil {
...
+- This value defaults to 65536(64kb) for normal account, you can increase the limit by raising a ticket at [ably support](https://ably.com/support).
```
owenpearson, line 201:
```suggestion
- This value defaults to 64kb, please [get in touch](https://ably.com/support) if you would like to request a higher limit for your account.
```

README.md
```diff
@@ -196,6 +196,22 @@ if err != nil {
...
+- Upgrading account to higher limit should automatically update `ConnectionDetails.MaxMessageSize` property and should accordingly set the client side connection readlimit.
```
owenpearson, line 202:
```suggestion
- Upgrading your account to higher limit will automatically update `MaxMessageSize` property and should accordingly set the client side connection read limit.
```

README.md
```diff
@@ -196,6 +196,22 @@ if err != nil {
...
+- If you are still facing issues while reading bigger size messages or intentionally want to reduce the limit, you can explicitly update the connection read limit.
```
owenpearson, line 203:
```suggestion
- If you are still facing issues when receiving large messages or intentionally want to reduce the limit, you can explicitly update the connection read limit:
```

README.md
```diff
@@ -196,6 +196,22 @@ if err != nil {
...
+
+```go
+client, err = ably.NewRealtime(ably.WithKey("xxx:xxx"))
+if err != nil {
+        panic(err)
+}
+client.Connection.SetReadLimit(131072) // Set read limit to 128kb, overriding default ConnectionDetails.MaxMessageSize
+```
+
+Note - If connection read limit is less than size of received message, client will throw error "failed to read: read limited at {READ_LIMIT + 1} bytes" and will close the connection.
```
owenpearson, line 213:
```suggestion
Note - If connection read limit is less than size of received message, the client will throw an error "failed to read: read limited at {READ_LIMIT + 1} bytes" and will close the connection.
```

owenpearson (changes requested):
I'm not a golang expert but based on the tests these changes look good to me 👍 a few suggestions for the readme update below.

owenpearson (approved):
LGTM, thanks

PR #621, opened 2023-08-14
[SDK-222] [no connection serial ] + implement missing internal presence map + presencequeue

### LINKED ISSUE #255: RTP2, RTP17, RTP19: PresenceMap
Write spec tests for [RTP2](https://docs.ably.io/client-lib-development-guide/features/#RTP2), [RTP17](https://docs.ably.io/client-lib-development-guide/features/#RTP17), [RTP19](https://docs.ably.io/client-lib-development-guide/features/#RTP19), ensure they pass and that the affected API conforms to [the 1.2 API design](https://github.com/ably/ably-go/issues/197).

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-222) by [Unito](https://www.unito.io)

sacOO7:
- Fixed #255

### DIFF
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -211,6 +211,12 @@ func (c *Connection) SetKey(key string) {
 	c.key = key
 }
 
+func (c *RealtimePresence) GetMembers() map[string]*PresenceMessage {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.members
+}
+
 func (c *Connection) ConnectionStateTTL() time.Duration {
 	return c.connectionStateTTL()
 }
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -217,6 +217,12 @@ func (c *RealtimePresence) GetMembers() map[string]*PresenceMessage {
 	return c.members
 }
 
+func (c *RealtimePresence) InternalMembers() map[string]*PresenceMessage {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.internalMembers
+}
+
 func (c *RealtimePresence) SyncInitial() bool {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -205,14 +205,12 @@ func (c *Connection) PendingItems() int {
 	return len(c.pending.queue)
 }
 
-type MsgWithAckCallback = msgWithAckCallback
-
 // AckAll empties queue and acks all pending callbacks
 func (c *Connection) AckAll() {
 	c.mtx.Lock()
 	cx := c.pending.Dismiss()
 	c.mtx.Unlock()
-	c.log().Debugf("resending %d messages waiting for ACK/NACK", len(cx))
+	c.log().Infof("Ack all %d messages waiting for ACK/NACK", len(cx))
 	for _, v := range cx {
 		v.onAck(nil)
 	}
@@ -224,16 +222,24 @@ func (c *Connection) SetKey(key string) {
 	c.key = key
 }
 
-func (c *RealtimePresence) GetMembers() map[string]*PresenceMessage {
+func (c *RealtimePresence) Members() map[string]*PresenceMessage {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
-	return c.members
+	presenceMembers := make(map[string]*PresenceMessage, len(c.members))
+	for k, pm := range c.members {
+		presenceMembers[k] = pm
+	}
+	return presenceMembers
 }
 
 func (c *RealtimePresence) InternalMembers() map[string]*PresenceMessage {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
-	return c.internalMembers
+	internalMembers := make(map[string]*PresenceMessage, len(c.internalMembers))
+	for k, pm := range c.internalMembers {
+		internalMembers[k] = pm
+	}
+	return internalMembers
 }
 
 func (c *RealtimePresence) SyncInitial() bool {
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -93,6 +93,12 @@ func (c *REST) GetCachedFallbackHost() string {
 	return c.successFallbackHost.get()
 }
 
+func (c *RealtimeChannel) GetChannelSerial() string {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.properties.ChannelSerial
+}
+
 func (c *RealtimeChannel) GetAttachResume() bool {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -217,6 +217,18 @@ func (c *RealtimePresence) GetMembers() map[string]*PresenceMessage {
 	return c.members
 }
 
+func (c *RealtimePresence) SyncInitial() bool {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.syncState == syncInitial
+}
+
+func (c *RealtimePresence) SyncInProgress() bool {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.syncState == syncInProgress
+}
+
 func (c *Connection) ConnectionStateTTL() time.Duration {
 	return c.connectionStateTTL()
 }
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -205,6 +205,19 @@ func (c *Connection) PendingItems() int {
 	return len(c.pending.queue)
 }
 
+type MsgWithAckCallback = msgWithAckCallback
+
+// AckAll empties queue and acks all pending callbacks
+func (c *Connection) AckAll() {
+	c.mtx.Lock()
+	cx := c.pending.Dismiss()
+	c.mtx.Unlock()
+	c.log().Debugf("resending %d messages waiting for ACK/NACK", len(cx))
+	for _, v := range cx {
+		v.onAck(nil)
+	}
+}
+
 func (c *Connection) SetKey(key string) {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
--- a/ably/export_test.go
+++ b/ably/export_test.go
@@ -105,6 +105,12 @@ func (c *RealtimeChannel) SetAttachResume(value bool) {
 	c.attachResume = value
 }
 
+func (c *RealtimeChannel) SetState(chanState ChannelState) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	c.state = chanState
+}
+
 func (opts *clientOptions) GetFallbackRetryTimeout() time.Duration {
 	return opts.fallbackRetryTimeout()
 }
@@ -191,6 +197,12 @@ func (c *Connection) PendingItems() int {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 	return len(c.pending.queue)
+}
+
+func (c *Connection) SetKey(key string) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	c.key = key
 }
 
 func (c *Connection) ConnectionStateTTL() time.Duration {
--- a/ably/proto_protocol_message_test.go
+++ b/ably/proto_protocol_message_test.go
@@ -23,7 +23,7 @@ func TestProtocolMessageEncodeZeroSerials(t *testing.T) {
 	}
 	encoded, err := ablyutil.MarshalMsgpack(msg)
 	assert.NoError(t, err)
-	// expect a 3-element map with both the serial fields set to zero
+	// expect a 2-element map with both the serial fields set to zero
 	expected := []byte("\x82\xa2id\xa4test\xa9msgSerial\x00")
 	assert.True(t, bytes.Equal(encoded, expected),
 		"unexpected msgpack encoding\nexpected: %x\nactual:   %x", expected, encoded)
--- a/ably/proto_protocol_message_test.go
+++ b/ably/proto_protocol_message_test.go
@@ -5,7 +5,6 @@ package ably_test
 
 import (
 	"bytes"
-	"fmt"
 	"strconv"
 	"testing"
 
@@ -137,106 +136,4 @@ func TestIfHasFlg(t *testing.T) {
 		"Should contain flag %v", ably.FlagPublish)
 	assert.False(t, flags.Has(ably.FlagHasBacklog),
 		"Shouldn't contain flag %v", ably.FlagHasBacklog)
-}
-
-func TestPresenceCheckForNewNessByTimestampIfSynthesized_RTP2b1(t *testing.T) {
-	presenceMsg1 := &ably.PresenceMessage{
-		Action: ably.PresenceActionPresent,
-		Message: ably.Message{
-			ID:           "123:12:1",
-			Timestamp:    125,
-			ConnectionID: "987",
-		},
-	}
-	presenceMsg2 := &ably.PresenceMessage{
-		Action: ably.PresenceActionPresent,
-		Message: ably.Message{
-			ID:           "123:12:2",
-			Timestamp:    123,
-			ConnectionID: "784",
-		},
-	}
-	isNewMsg, err := presenceMsg1.IsNewerThan(presenceMsg2)
-	assert.Nil(t, err)
-	assert.True(t, isNewMsg)
-
-	isNewMsg, err = presenceMsg2.IsNewerThan(presenceMsg1)
-	assert.Nil(t, err)
-	assert.False(t, isNewMsg)
-}
-
-func TestPresenceCheckForNewNessBySerialIfNotSynthesized__RTP2b2(t *testing.T) {
-	oldPresenceMsg := &ably.PresenceMessage{
-		Action: ably.PresenceActionPresent,
-		Message: ably.Message{
-			ID:           "123:12:0",
-			Timestamp:    123,
-			ConnectionID: "123",
-		},
-	}
-	newPresenceMessage := &ably.PresenceMessage{
-		Action: ably.PresenceActionPresent,
-		Message: ably.Message{
-			ID:           "123:12:1",
-			Timestamp:    123,
-			ConnectionID: "123",
-		},
-	}
-	isNewMsg, err := oldPresenceMsg.IsNewerThan(newPresenceMessage)
-	assert.Nil(t, err)
-	assert.False(t, isNewMsg)
-
-	isNewMsg, err = newPresenceMessage.IsNewerThan(oldPresenceMsg)
-	assert.Nil(t, err)
-	assert.True(t, isNewMsg)
-}
-
-func TestPresenceMessagesShouldReturnErrorForWrongMessageSerials__RTP2b2(t *testing.T) {
-	// Both has invalid msgserial
-	msg1 := &ably.PresenceMessage{
-		Action: ably.PresenceActionPresent,
-		Message: ably.Message{
-			ID:           "123:1a:0",
-			Timestamp:    123,
-			ConnectionID: "123",
-		},
-	}
-
-	msg2 := &ably.PresenceMessage{
-		Action: ably.PresenceActionPresent,
-		Message: ably.Message{
-			ID:           "123:1b:1",
-			Timestamp:    124,
-			ConnectionID: "123",
-		},
-	}
-	isNewMsg, err := msg1.IsNewerThan(msg2)
-	assert.NotNil(t, err)
-	assert.Contains(t, fmt.Sprint(err), "the presence message has invalid msgSerial, for msgId 123:1a:0")
-	assert.False(t, isNewMsg)
-
-	isNewMsg, err = msg2.IsNewerThan(msg1)
-	assert.NotNil(t, err)
-	assert.Contains(t, fmt.Sprint(err), "the presence message has invalid msgSerial, for msgId 123:1b:1")
-	assert.False(t, isNewMsg)
-
-	// msg2 has valid messageSerial
-	msg2 = &ably.PresenceMessage{
-		Action: ably.PresenceActionPresent,
-		Message: ably.Message{
-			ID:           "123:10:0",
-			Timestamp:    124,
-			ConnectionID: "123",
-		},
-	}
-
-	isNewMsg, err = msg1.IsNewerThan(msg2)
-	assert.NotNil(t, err)
-	assert.Contains(t, fmt.Sprint(err), "the presence message has invalid msgSerial, for msgId 123:1a:0")
-	assert.False(t, isNewMsg)
-
-	isNewMsg, err = msg2.IsNewerThan(msg1)
-	assert.NotNil(t, err)
-	assert.Contains(t, fmt.Sprint(err), "the presence message has invalid msgSerial, for msgId 123:1a:0")
-	assert.True(t, isNewMsg)
 }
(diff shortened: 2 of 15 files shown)


sacOO7:
TODO - Check for channel attach while publishing presence msg

ably/export_test.go
```diff
@@ -193,6 +205,49 @@ func (c *Connection) PendingItems() int {
...
+func (c *Connection) SetKey(key string) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	c.key = key
+}
+
+func (c *RealtimePresence) GetMembers() map[string]*PresenceMessage {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.members
```
lmars, line 230: If this map might get mutated after this call returns, then holding the lock just to return the map isn't enough, the map itself needs to be thread safe.

ably/proto_protocol_message_test.go
```diff
@@ -24,7 +24,7 @@ func TestProtocolMessageEncodeZeroSerials(t *testing.T) {
 	encoded, err := ablyutil.MarshalMsgpack(msg)
 	assert.NoError(t, err)
 	// expect a 3-element map with both the serial fields set to zero
-	expected := []byte("\x83\xB0id\xA4test\xA9msgSerial\x00")
+	expected := []byte("\x82\xa2id\xa4test\xa9msgSerial\x00")
```
lmars, line 27: The comment is now incorrect, this is a two-element map.

ably/realtime_channel.go
```diff
@@ -548,6 +547,17 @@ func (c *RealtimeChannel) SubscribeAll(ctx context.Context, handle func(*Message
 	return unsubscribe, nil
 }
 
+// SubscribeAll registers an event listener for error messages on this channel.
```
lmars, line 550: This comment doesn't match the method.

ably/realtime_channel.go
```diff
@@ -548,6 +547,17 @@ func (c *RealtimeChannel) SubscribeAll(ctx context.Context, handle func(*Message
...
+//
+// See package-level documentation => [ably] Event Emitters for details about messages dispatch.
+func (c *RealtimeChannel) OnError(handle func(*ErrorInfo)) (func(), error) {
+	// unsubscribe deregisters all listeners to error messages on this channel.
+	unsubscribe := c.errorEmitter.OnAll(func(err emitterData) {
+		handle((*ErrorInfo)(err.(*errorMessage)))
+	})
+	return unsubscribe, nil
```
lmars, line 558: Given this function never returns an error, perhaps drop the `err` return value?

ably/realtime_presence.go
```diff
@@ -100,23 +130,60 @@ func (pres *RealtimePresence) syncWait() {
...
 	}
-	return ""
+	return false, syncSequenceId, syncCursor
 }
 
-func (pres *RealtimePresence) onAttach(msg *protocolMessage) {
-	serial := syncSerial(msg)
+func (pres *RealtimePresence) enterMembers(internalMembers []*PresenceMessage) {
+	for _, member := range internalMembers {
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
lmars, line 150: Is there any context as to why this timeout is set to 2 minutes?

ably/realtime_presence.go
```diff
@@ -100,23 +130,60 @@ func (pres *RealtimePresence) syncWait() {
...
-		pres.syncState = syncComplete
-		pres.syncMtx.Unlock()
+	// RTP1
+	if msg.Flags.Has(flagHasPresence) {
+		pres.syncStart()
+	} else {
+		pres.leaveMembers(pres.members) // RTP19a
+		if pres.syncState == syncInitial {
+			pres.syncState = syncComplete
+			pres.syncMtx.Unlock()
```
lmars, line 172: This looks susceptible to deadlock, where does this get locked? Locks and unlocks should really always appear in pairs close to each other.

sacOO7: This is also done because of `syncWait` method, you can check it's usage in comments
https://github.com/ably/ably-go/blob/e66ad6a1bfa0e0e48dccb7c1afd0c282b568d1c2/ably/realtime_presence.go#L93-L99

ably/realtime_presence.go
sacOO7: Honestly there's no context as such, I just didn't want goroutine to get blocked!
Your suggestions will be helpful here !!

ably/export_test.go
```diff
@@ -193,6 +205,49 @@ func (c *Connection) PendingItems() int {
...
 }
 
+type MsgWithAckCallback = msgWithAckCallback
+
+// AckAll empties queue and acks all pending callbacks
+func (c *Connection) AckAll() {
+	c.mtx.Lock()
+	cx := c.pending.Dismiss()
+	c.mtx.Unlock()
+	c.log().Debugf("resending %d messages waiting for ACK/NACK", len(cx))
```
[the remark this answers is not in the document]
sacOO7: Yeah, `dismiss` returns queuedMessages and empties internal queue

ably/export_test.go
sacOO7: actually these methods are only used for `testing`, are not used anywhere else.
All methods defined under `export_test.go` are available only for tests.

ably/realtime_presence.go
```diff
@@ -100,23 +130,60 @@ func (pres *RealtimePresence) syncWait() {
...
+		// RTP17e
+		if err != nil {
+			pres.channel.log().Errorf("Error for internal member presence enter with id %v, clientId %v, err %v", member.ID, member.ClientID, err)
+			pres.channel.emitErrorUpdate(newError(91004, err), true)
+		}
+		cancel()
+	}
+}
+
+func (pres *RealtimePresence) onAttach(msg *protocolMessage, isAttachWithoutMessageLoss bool) {
```
SimonWoolf, line 162: The naming of `isAttachWithoutMessageLoss` seems misleading. That'll be set to true in any situation where you need to re-enter your members, which includes both situations where you do and don't have message loss

ably/realtime_channel.go
```diff
@@ -794,11 +808,13 @@ func (c *RealtimeChannel) notify(msg *protocolMessage) {
...
-		isAttachResumed := msg.Flags.Has(flagResumed)
-		if c.state != ChannelStateAttached || !isAttachResumed { //RTL12
-			c.setState(ChannelStateAttached, newErrorFromProto(msg.Error), isAttachResumed)
+		// RTL12
+		if c.state == ChannelStateAttached && !msg.Flags.Has(flagResumed) {
+			c.emitErrorUpdate(newErrorFromProto(msg.Error), false) // RTL12
+			c.Presence.onAttach(msg, false)
+		} else {
+			c.Presence.onAttach(msg, true)
+			c.setState(ChannelStateAttached, newErrorFromProto(msg.Error), msg.Flags.Has(flagResumed))
```
SimonWoolf, line 817: This conditional looks kinda weird. You'll be in this `else` clause if either of the two conditions in the if fail, but the stuff in the else clause only makes sense if it was the first condition that failed -- eg what if you get an attached with resumed=true while you were already attached? seems like it should be
```
		if c.state == ChannelStateAttached {
			if !msg.Flags.Has(flagResumed) {
				c.emitErrorUpdate(newErrorFromProto(msg.Error), false) // RTL12
				c.Presence.onAttach(msg, false)
			}
		} else {
			c.Presence.onAttach(msg, true)
			c.setState(ChannelStateAttached, newErrorFromProto(msg.Error), msg.Flags.Has(flagResumed))
		}
```

ably/realtime_conn.go
```diff
@@ -593,11 +597,16 @@ func (c *Connection) advanceSerial() {
...
 	hasMsgSerial := msg.Action == actionMessage || msg.Action == actionPresence
 	c.mtx.Lock()
+	// RTP16a - in case of presence msg send, check for connection status and send accordingly
 	switch state := c.state; state {
 	default:
 		c.mtx.Unlock()
 		if onAck != nil {
-			onAck(connStateError(state, nil))
+			if c.state == ConnectionStateClosed {
+				onAck(errClosed)
```
SimonWoolf, line 606: why does `closed` need to be special-cased? can't it be handled by `connStateError(state)` like all the others?

ably/realtime_conn.go
```diff
@@ -606,9 +615,9 @@ func (c *Connection) send(msg *protocolMessage, onAck func(err error)) {
 			if onAck != nil {
 				onAck(connStateError(state, errQueueing))
 			}
+		} else {
+			c.queue.Enqueue(msg, onAck) // RTL4i
```
SimonWoolf, line 619: :+1: good job spotting that bug

ably/realtime_presence.go
```diff
@@ -127,82 +194,147 @@ func (pres *RealtimePresence) SyncComplete() bool {
...
+}
+
+// RTP2a, RTP2b, RTP2c
+func (pres *RealtimePresence) removePresenceMember(memberMap map[string]*PresenceMessage, memberKey string, presenceMember *PresenceMessage) bool {
+	if existingMember, ok := memberMap[memberKey]; ok { // RTP2a
+		isMemberNew, err := presenceMember.IsNewerThan(existingMember) // RTP2b
+		if err != nil {
+			pres.log().Error(err)
+			errorInfo := newError(0, err)
+			pres.channel.errorEmitter.Emit(subscriptionName("error"), (*errorMessage)(errorInfo))
```
SimonWoolf, line 269: as discussed above -- I'd get rid of this

ably/realtime_presence.go
sacOO7: Let's say, if you `attached` channel when previous state was `detached` or `attaching`, do we get `resume` as false in such a case?

sacOO7: So, we should call it `isFreshAttach` wdyt?

ably/realtime_presence.go
sacOO7: Not sure why we even added this in the first place. I am sure there was a concrete reason

sacOO7: You can see in `ably-dotnet`, we have a `realtimechannel` method called `OnError`.
https://github.com/ably/ably-dotnet/blob/c2a514ad2cc9401547c8c6ccb7c23b87101131a5/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs#L559-L567
This gets called whenever there is any error registered on a channel 
https://github.com/ably/ably-dotnet/blob/c2a514ad2cc9401547c8c6ccb7c23b87101131a5/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs#L628

You can see there is an error delegate registration here 
https://github.com/ably/ably-dotnet/blob/c2a514ad2cc9401547c8c6ccb7c23b87101131a5/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs#L57C44-L57C44

sacOO7: `OnError` method usages
1. Presence -> https://github.com/ably/ably-dotnet/blob/c2a514ad2cc9401547c8c6ccb7c23b87101131a5/src/IO.Ably.Shared/Realtime/Presence.cs#L610
2. Sync message -> https://github.com/ably/ably-dotnet/blob/c2a514ad2cc9401547c8c6ccb7c23b87101131a5/src/IO.Ably.Shared/Realtime/ChannelMessageProcessor.cs#L148
3. Channel state errors -> https://github.com/ably/ably-dotnet/blob/c2a514ad2cc9401547c8c6ccb7c23b87101131a5/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs#L628

ably/realtime_channel.go
```diff
@@ -548,6 +547,17 @@ func (c *RealtimeChannel) SubscribeAll(ctx context.Context, handle func(*Message
 	return unsubscribe, nil
 }
 
+// OnError registers an event listener for error messages on this channel.
+//
+// See package-level documentation => [ably] Event Emitters for details about messages dispatch.
+func (c *RealtimeChannel) OnError(handle func(*ErrorInfo)) (func(), error) {
```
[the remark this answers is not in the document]
sacOO7: I believe we do have `OnError` public method for `ably-js` and `ably-dotnet`.

ably/realtime_presence.go
sacOO7: Let me know if we should keep it. If this is not part of the spec, then there's no point in keeping the public method.

sacOO7: Seems we need to do something like this 
https://github.com/ably/ably-java/blob/72992c943498ee2ebefb8acec303718ed7f35ae4/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L1336C1-L1339C6

sacOO7: Okay, I have removed this particular piece of code. 
Let me know if we planning to add `OnError` method to the spec. I think this will be a very specific way to listen to channel-specific errors as we do the same in `ably-dotnet`.

ably/realtime_conn.go
sacOO7: Actually 
https://github.com/ably/ably-go/blob/e66ad6a1bfa0e0e48dccb7c1afd0c282b568d1c2/ably/state.go#L55-L60 map doesn't contain `errClosed`. 
At the same time, we can't update the map to include `errClosed` since it can be a graceful close from the client side. 
If we update the map with `errClosed`, other tests start failing where we don't expect error when client goes into closed state, like when we call `client.close()` and `conn` state changes to `closed` with error.
This error is specific to our usecase, so I didn't include it into the map

ably/proto_protocol_message_test.go
sacOO7: updated

ably/realtime_conn.go
sacOO7: thanks

ably/export_test.go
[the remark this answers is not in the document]
lmars: Ok, but isn't that _not_ resending them, instead emptying the queue and ack'ing them?

ably/export_test.go
lmars: Ok, but even in the tests we should be concerned about making things thread safe, otherwise we'll get failing tests due to races being detected.

ably/realtime_presence.go
lmars: Perhaps it should keep retrying whilst the channel is in a state where we can expect the operation to eventually succeed (i.e. it's not in a terminal state)?

ably/realtime_presence.go
lmars: Ok, that's really not a good use of a mutex, a mutex should be used to synchronise access to shared memory, not to synchronise clients waiting for the result of a network call (i.e. here waiting for a presence sync to complete). For example, what happens if the sync doesn't complete because the channel became detached? Does the mutex stay locked, and all calls to `syncWait` block until the program exits?

It's typical to use the closing of an empty struct channel to unblock multiple callers when some event has happened, and to select on it in a cancelable way:

```go
// a field on RealtimePresence
syncDone := make(chan struct{})

// when the sync has completed
close(p.syncDone)

// wait for the channel to close in syncWait
func (p *RealtimePresence) syncWait(ctx context.Context) error {
select {
case <-p.syncDone:
	return nil
case <-ctx.Done():
	return ctx.Err()
}
```

ably/realtime_presence.go
sacOO7: I just don't want goroutine to go into a dangling state : (
Sure, it can work.

ably/export_test.go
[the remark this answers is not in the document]
sacOO7: Yeah, I will update message for the same

ably/realtime_presence.go
sacOO7: sure, we can update

ably/realtime_channel.go
SimonWoolf: > I think RTL12 should mention explicitly, if attached is received on already attached channel and resume set to true, then it should do nothing

It already does say that. "If and only if the resumed flag is false, this should result in the channel emitting an UPDATE event with a ChannelStateChange object". "If _and only if_"

ably/realtime_channel.go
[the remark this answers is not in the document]
sacOO7: Sorry, i should have updated it. Only dotnet has it

ably/realtime_conn.go
SimonWoolf: > At the same time, we can't update the map to include errClosed since it can be a graceful close from the client side.

I don't follow. If the connection is closed, then most operations you try to do to it (other than reconnect) should fail with a 'connection closed' error, similarly to if it's failed. if connclosed is missing from that then either every place you call is going to need to check if the conn is closed and handle that explicitly, or you have a bug, no?

> E.g. call to client.close(), and conn state changes to closed with error.

No it doesn't because connection state changes don't have errors. They have `reason`s. The reason might be 'the connection was closed', that's not an error, it's telling you the connection is in that state

ably/realtime_presence.go
SimonWoolf: > You can see in ably-dotnet, we have a realtimechannel method called OnError.

I am not a dot.net developer. But my understanding is that if it's not part of the realtime channel interface then it's not part of the public api. It's an internal method, used by the library itself. It's annotated 'public' because it can be called by other parts of the library, but that doesn't make it part of the actual public API.

> Let me know if we planning to add OnError method to the spec.

We're not.

> [Seems we need to do something like this](https://github.com/ably/ably-java/blob/72992c943498ee2ebefb8acec303718ed7f35ae4/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L1336C1-L1339C6)

[We already have that](https://github.com/ably/ably-go/blob/fix/presencemap/ably/realtime_channel.go#L995)

sacOO7: Got it 👍

ably/realtime_presence.go
SimonWoolf: > To put another way, what if the enterClient call is failing due to an unreliable network for 3 minutes, but then it's ok again, shouldn't this then succeed rather than have give up after 2 minutes?

If a message is sent to the connection while it's in a `connected` state, then eventually either it will successfully send or the connection will detect a network issue and become disconnected. If the latter happens, then when it reconnects, the channel will re-attach, which will trigger `onAttach()` and `presence.enterMembers()` to happen again at that time. So we shouldn't need an old enterMembers from some previous attach cycle living on.

ably/realtime_conn.go
sacOO7: I believe it's called `errorReason` =>
https://github.com/ably/ably-go/blob/e66ad6a1bfa0e0e48dccb7c1afd0c282b568d1c2/ably/realtime_conn.go#L48-L51
and same is copied into reason
https://github.com/ably/ably-go/blob/e66ad6a1bfa0e0e48dccb7c1afd0c282b568d1c2/ably/realtime_conn.go#L1001-L1006

There are tests which doesn't treat connection closed as connection failure as such. I mean ideally they shouldn't.
Because, `client.close()` initiates it and gracefully closes the connection accordingly.

ably/realtime_presence.go
lmars: Sorry yes, I'm proposing `context.Background()`, since if I understand Simon correctly, the call will either block until it succeeds, or return an error if the client becomes disconnected, so in effect there is a timeout on this operation based on how long it takes the client to detect a disconnection, there's no need to specify an arbitrary timeout here too.

ably/realtime_conn.go
sacOO7: Do we have a spec item that specifies to set a reason when client explicitly sends a `close` message and  connection is closed by server

ably/realtime_presence.go
sacOO7: Same will happen with `cancelledContext` too. If it fails to send message due to `disconnection`, for next channel attach it will automatically send the message. We don't really care about message failure in current case. Because, it will only fail when channel is in some other state ( maybe because of disconnection ), for next attach, we can enter members again.
There's no need to wait really

ably/realtime_presence.go
sacOO7: Updated as a part of -> https://github.com/ably/ably-go/pull/621/commits/2740853cd88a58e6b0b6332904e61efcb3c4912c

ably/export_test.go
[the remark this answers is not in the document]
sacOO7: updated -> https://github.com/ably/ably-go/pull/621/commits/31b739a380c961ff176671093579c8928e4b9792

ably/export_test.go
sacOO7: updated -> https://github.com/ably/ably-go/pull/621/commits/31b739a380c961ff176671093579c8928e4b9792

ably/realtime_presence.go
lmars: Ok thanks, can we get rid of `syncMtx` altogether?

sacOO7: we did

ably/realtime_channel.go
sacOO7: ```The library must never emit a ChannelState ChannelEvent for a state equal to the previous state``` means it should emit `UpdateEvent` for prevState = newState` instead of `ChannelEvent`. That's what I understood by it.

sacOO7: As per RTL2f =>
```
When a channel ATTACHED ProtocolMessage is received, the ProtocolMessage may contain a RESUMED bit flag indicating that the channel has been resumed. The corresponding ChannelStateChange (either ATTACHED per RTL2a, or UPDATE per RTL12) will contain a resumed boolean attribute with value true if the bit flag RESUMED was included ...
```
I feel there's a gray area here. It doesn't mention if it's a ATTACHED on non-attached channel or ATTACHED on already attached channel. I can only assume if it's ATTACHED on already ATTACHED with resume flag as true, as per RTL2g, it should emit update event on the channel.
RTL12 only describes ATTACHED on already ATTACHED with resume flag as false, other specs are implicitly pointing towards setting a state and emitting update event for the same.

sacOO7: I have updated the code according to your suggestion.
But, I would be honest, I had to read related spec multiple times again and again for this. Still, as you can see, understanding wasn't reflected in the code.
Maybe I will update the spec here https://github.com/ably/specification/pull/167 and we can resume discussion for the same.

ably/realtime_presence.go
sacOO7: Update -> `context.Background()` does seem to be a sensible approach, but is there a way this can go into a lock type of a situation where client is connected for a long time and ack is not received for the entered presence.

sacOO7: That's why I tried putting a `timeout` here, where all `innerPresenceMembers` will be processed without one waiting for another

lmars: Ok, I leave it to your judgement, I'd probably just use `context.Background()`

ably/realtime_presence.go
lmars: Yes sorry, I think I was looking at an outdated diff 👍

lmars (approved):
This looks ok from my perspective, but I defer to @SimonWoolf to review the changes he requested

ably/realtime_channel.go
sacOO7: I have added explicit commit to the specification PR https://github.com/ably/specification/pull/167/commits/248bbc63a29e7393e56c2b72b92ab5097b628e30

sacOO7: Since the change has been made as per suggestion, I am closing the issue

sacOO7:
@lmars merging this PR into integration-2.0, since one more review required for recovery spec related commits. I will re-raise @SimonWoolf comments there

PR #624, opened 2023-10-03
Fix message extras unmarshaled to incompatible json type

zknill:
Message extras should be a json compatible map[string]interface{}, but when marshalling to msgpack and back again, any nested objects become map[interface{}]interface{} which is incompatible with json marshalling.

Set the map type on the msgpack handle to map[string]interface{} so that nested objects are correctly unmarshaled.

### DIFF
--- a/ably/internal/ablyutil/msgpack_test.go
+++ b/ably/internal/ablyutil/msgpack_test.go
@@ -5,6 +5,7 @@ package ablyutil
 
 import (
 	"bytes"
+	"encoding/json"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


ably/proto_message_decoding_test.go
```diff
@@ -107,6 +107,34 @@ func loadMsgpackFixtures() ([]MsgpackTestFixture, error) {
 	return o, err
 }
 
+func Test_msgpackExtrasJsonCompatible(t *testing.T) {
```
lmars, line 110:
```suggestion
func TestMsgpackExtrasJsonCompatible(t *testing.T) {
```

(to be consistent with other test names)

ably/internal/ablyutil/msgpack_test.go
```diff
@@ -31,3 +32,28 @@ func TestMsgpack(t *testing.T) {
 		}
 	})
 }
+
+func Test_msgpackJson(t *testing.T) {
```
lmars, line 36:
```suggestion
func TestMsgpackJson(t *testing.T) {
```

(to be consistent with other test names)

lmars (approved):
LGTM other than test naming 👍

PR #631, opened 2024-02-01
fix: keep client default client read limit when realtime doesn't specify

AndyTWF:
This is mostly for internal purposes, where we have MaxMessageSize set to 0 (read: unlimited). When this happens, the SDK should keep whatever limit it currently has set.

Previously it was trying to set 0 which prevented anything from being read from the connection/

### DIFF
--- a/ably/realtime_channel_integration_test.go
+++ b/ably/realtime_channel_integration_test.go
@@ -294,6 +294,47 @@ func TestRealtimeChannel_ShouldSetProvidedReadLimit(t *testing.T) {
 	assert.Equal(t, int64(2048), client.Connection.ReadLimit())
 }
 
+func TestRealtimeChannel_SetsNoLimitIfServerNoLimits(t *testing.T) {
+	// Mock out the dial
+	dial := DialFunc(func(p string, url *url.URL, timeout time.Duration) (ably.Conn, error) {
+		return connMock{
+			SendFunc: func(m *ably.ProtocolMessage) error {
+				return nil
+			},
+			ReceiveFunc: func(deadline time.Time) (*ably.ProtocolMessage, error) {
+				connDetails := ably.ConnectionDetails{
+					ClientID:           "id1",
+					ConnectionKey:      "foo",
+					MaxFrameSize:       12,
+					MaxInboundRate:     14,
+					MaxMessageSize:     0,
+					ConnectionStateTTL: ably.DurationFromMsecs(time.Minute * 2),
+					MaxIdleInterval:    ably.DurationFromMsecs(time.Second),
+				}
+
+				return &ably.ProtocolMessage{
+					Action:            ably.ActionConnected,
+					ConnectionID:      "connection-id-1",
+					ConnectionDetails: &connDetails,
+				}, nil
+			},
+			CloseFunc: func() error {
+				return nil
+			},
+		}, nil
+	})
+
+	_, c := ablytest.NewRealtime(ably.WithDial(dial))
+
+	// Wait for a little bit for things to settle
+	time.Sleep(1 * time.Second)
+
+	// Check that the connection read limit is the default - due to websocket.go
+	// Only allowing the value to be set if the connection is a certain type
+	// We'll just check -1 here
+	assert.Equal(t, int64(-1), c.Connection.ReadLimit())
+}
+
 func TestRealtimeChannel_ShouldReturnErrorIfReadLimitExceeded(t *testing.T) {
 	app, client1 := ablytest.NewRealtime(ably.WithEchoMessages(false))
 	defer safeclose(t, ablytest.FullRealtimeCloser(client1), app)
--- a/ably/realtime_channel_integration_test.go
+++ b/ably/realtime_channel_integration_test.go
@@ -294,36 +294,29 @@ func TestRealtimeChannel_ShouldSetProvidedReadLimit(t *testing.T) {
 	assert.Equal(t, int64(2048), client.Connection.ReadLimit())
 }
 
-func TestRealtimeChannel_SetDefaultLimitIfNoServerLimit(t *testing.T) {
-	in := make(chan *ably.ProtocolMessage, 1)
-	out := make(chan *ably.ProtocolMessage, 16)
+func TestRealtimeChannel_SetDefaultReadLimitIfServerHasNoLimit(t *testing.T) {
 
-	_, c := ablytest.NewRealtime(ably.WithDial(MessagePipe(in, out)))
-
-	connDetails := ably.ConnectionDetails{
-		ClientID:           "id1",
-		ConnectionKey:      "foo",
-		MaxFrameSize:       12,
-		MaxInboundRate:     14,
-		MaxMessageSize:     0,  // 0 represents no limit on message size
-		ConnectionStateTTL: ably.DurationFromMsecs(time.Minute * 2),
-		MaxIdleInterval:    ably.DurationFromMsecs(time.Second),
+	dial := func(proto string, url *url.URL, timeout time.Duration) (ably.Conn, error) {
+		return ably.DialWebsocket(proto, url, timeout)
 	}
+	wrappedDialWebsocket, interceptMsg := DialIntercept(dial)
 
-	in <- &ably.ProtocolMessage{
-		Action:            ably.ActionConnected,
-		ConnectionID:      "connection-id-1",
-		ConnectionDetails: &connDetails,
-	}
+	connMsgContext, cancel := context.WithCancel(context.Background())
+	connMessageCh := interceptMsg(connMsgContext, ably.ActionConnected)
 
-	// Wait for a little bit for things to settle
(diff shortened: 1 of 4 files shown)


ably/realtime_channel_integration_test.go
```diff
@@ -294,6 +294,47 @@ func TestRealtimeChannel_ShouldSetProvidedReadLimit(t *testing.T) {
 	assert.Equal(t, int64(2048), client.Connection.ReadLimit())
 }
 
+func TestRealtimeChannel_SetsNoLimitIfServerNoLimits(t *testing.T) {
+	// Mock out the dial
+	dial := DialFunc(func(p string, url *url.URL, timeout time.Duration) (ably.Conn, error) {
```
sacOO7, line 299: We can simplify test using `MessagePipe(in, out)` instead of `DialFunc`. Is it fine if I update the test with the same?

AndyTWF: Done in [514dd2f](https://github.com/ably/ably-go/pull/631/commits/514dd2f021fe4856cff16594153dfc6d936878e9) :)

ably/realtime_channel_integration_test.go
```diff
@@ -294,6 +294,37 @@ func TestRealtimeChannel_ShouldSetProvidedReadLimit(t *testing.T) {
...
+	}
+
+	in <- &ably.ProtocolMessage{
+		Action:            ably.ActionConnected,
+		ConnectionID:      "connection-id-1",
+		ConnectionDetails: &connDetails,
+	}
+
+	// Wait for a little bit for things to settle
+	time.Sleep(1 * time.Second)
```
sacOO7, line 320:
```suggestion
	err := ablytest.Wait(ablytest.ConnWaiter(c, c.Connect, ably.ConnectionEventConnected), nil)
	assert.NoError(t, err)
```

ably/realtime_channel_integration_test.go
```diff
@@ -294,6 +294,37 @@ func TestRealtimeChannel_ShouldSetProvidedReadLimit(t *testing.T) {
 	assert.Equal(t, int64(2048), client.Connection.ReadLimit())
 }
 
+func TestRealtimeChannel_SetsNoLimitIfServerNoLimits(t *testing.T) {
```
sacOO7, line 297:
```suggestion
func TestRealtimeChannel_SetDefaultLimitIfNoServerLimit(t *testing.T) {
```

ably/realtime_channel_integration_test.go
```diff
@@ -294,6 +294,37 @@ func TestRealtimeChannel_ShouldSetProvidedReadLimit(t *testing.T) {
...
+	out := make(chan *ably.ProtocolMessage, 16)
+
+	_, c := ablytest.NewRealtime(ably.WithDial(MessagePipe(in, out)))
+
+	connDetails := ably.ConnectionDetails{
+		ClientID:           "id1",
+		ConnectionKey:      "foo",
+		MaxFrameSize:       12,
+		MaxInboundRate:     14,
+		MaxMessageSize:     0,
```
sacOO7, line 308:
```suggestion
		MaxMessageSize:     0,  // 0 represents no limit on message size
```

ably/realtime_channel_integration_test.go
```diff
@@ -294,6 +294,37 @@ func TestRealtimeChannel_ShouldSetProvidedReadLimit(t *testing.T) {
...
+		ConnectionDetails: &connDetails,
+	}
+
+	// Wait for a little bit for things to settle
+	time.Sleep(1 * time.Second)
+
+	// Check that the connection read limit is the default - due to websocket.go
+	// Only allowing the value to be set if the connection is a certain type
+	// We'll just check -1 here
+	assert.Equal(t, int64(-1), c.Connection.ReadLimit())
```
sacOO7, line 325: `c.Connection.ReadLimit()` should return `64kb`. It only sets `-1` if there is error while setting the limit. When it's set to `-1`, limit is `32kb` which is default `nhooyr/websocket` readlimit. Currently, there is an error while setting the limit, because it doesn't use `websocketConn` as such -> https://github.com/ably/ably-go/blob/cc9e260bf265f89cb42fd8f19b62b904fef2d1c8/ably/websocket.go#L122
Seems we need to test it as a integration test without mocking `websocket_conn`

sacOO7: @AndyTWF is it fine if I can write integration test for this? seems mocking won't work in current case : (

sacOO7 (changes requested):

ably/realtime_channel_integration_test.go
AndyTWF: @sacOO7 I'm not sure how we achieve this without mocking - unless there's a way to tell sandbox to send 0, which I doubt we can to just for this test?

sacOO7: I'm thinking ...

sacOO7: One of the option I can think of is, updating `setConnectionReadLimit` method
```
func setConnectionReadLimit(c conn, readLimit int64) error {
	verboseConn, ok := c.(verboseConn)
	if !ok {
		return errors.New("cannot set readlimit for connection, connection does not use verboseConn")
	}
	websocketConn, ok := verboseConn.conn.(*websocketConn)
	if !ok {
		return errors.New("cannot set readlimit for connection, connection does not use nhooyr.io/websocket")
	}
	websocketConn.conn.SetReadLimit(readLimit)
	return nil
}
```
in a way that it recursively finds `websocketConn` as a part of supplied param `conn`. Then we set the readlimit on it. This doesn't put restriction of being `verboseConn` before `websocketConn`.


Then, we can have integration test with `DialFunc` that wraps `DialWebsocket conn`
```
func TestRealtimeConn_SendErrorReconnects(t *testing.T) {
	sendErr := make(chan error, 1)
	closed := make(chan struct{}, 1)
	allowDial := make(chan struct{})

	dial := DialFunc(func(p string, url *url.URL, timeout time.Duration) (ably.Conn, error) {
		<-allowDial
		ws, err := ably.DialWebsocket(p, url, timeout)
		if err != nil {
			return nil, err
		}
		return connMock{
			SendFunc: func(m *ably.ProtocolMessage) error {
				select {
				case err := <-sendErr:
					return err
				default:
					return ws.Send(m)
				}
			},
			ReceiveFunc: ws.Receive,
			CloseFunc: func() error {
				closed <- struct{}{}
				return ws.Close()
			},
		}, nil
	})

	app, c := ablytest.NewRealtime(ably.WithDial(dial))
	defer safeclose(t, ablytest.FullRealtimeCloser(c), app)

	allowDial <- struct{}{}

	err := ablytest.Wait(ablytest.ConnWaiter(c, c.Connect, ably.ConnectionEventConnected), nil)
	assert.NoError(t, err)

	// Cause a send error; expect message to be enqueued and transport to be
	// closed.

	sendErr <- errors.New("fail")

	publishErr := make(chan error, 1)
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
	go func() {
		e := c.Channels.Get("test").Publish(ctx, "test", nil)
		publishErr <- e
	}()

	ablytest.Instantly.Recv(t, nil, closed, t.Fatalf)
	ablytest.Instantly.NoRecv(t, nil, publishErr, t.Fatalf)

	// Reconnect should happen instantly as a result of transport closure.
	allowDial <- struct{}{}

	// After reconnection, message should be published.
	ablytest.Soon.Recv(t, &err, publishErr, t.Fatalf)
	assert.NoError(t, err)
}
```

sacOO7: or if we don't want to modify `setConnectionReadLimit` method, then we need to wrap result from 
```
 ws, err := ably.DialWebsocket(p, url, timeout)
```
inside
 `verboseConn`

sacOO7: @AndyTWF you can decide if all of those efforts are worth it in the end haha : )

sacOO7: Okay, it seems if we can modify `setConnectionReadLimit`, we have option to use 
https://github.com/ably/ably-go/blob/cc9e260bf265f89cb42fd8f19b62b904fef2d1c8/ably/ably_test.go#L461

AndyTWF: The last option probably sounds like the path of least resistance?

sacOO7: Yeah .. seems like it

sacOO7: Okay, I updated the code. Also, unwrapping approach feels a bit cleaner :  )

sacOO7 (approved):
LGTM

sacOO7:
@AndyTWF do try running the test and check if `limit` is actually set  ( with either logging or setting debut points )

AndyTWF:
LGTM - nicely done 👍🏻

PR #647, opened 2024-07-23
Realtime proxy support

sacOO7:
- Added test for realtime proxy support using env. variable


sacOO7:
Test failing because of https://github.com/golang/go/issues/63567

sacOO7:
Closing the PR, since issue with setting env. variables on CI cant be identified

PR #648, opened 2024-07-23
Test proxy

sacOO7:


sacOO7:
Closing the PR since we concluded that `HTTP_PROXY` and `HTTPS_PROXY` works for both REST and REALTIME. Same will be updated as a part of https://github.com/ably/ably-go/pull/647

PR #655, opened 2024-07-28
Fix presence newness check and flaky tests

sacOO7:
- As per https://sdk.ably.com/builds/ably/specification/main/features/#RTP2b2, newer message should have msgIndex bigger than other.
- Fixed flaky presence tests
- Avoids duplicate presence messages being emitted when server periodically sends `SYNC` message containing all presence messages.

### DIFF
--- a/ably/realtime_presence_integration_test.go
+++ b/ably/realtime_presence_integration_test.go
@@ -141,7 +141,7 @@ func TestRealtimePresence_EnsureChannelIsAttached(t *testing.T) {
 	assert.NoError(t, err)
 }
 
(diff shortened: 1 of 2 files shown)


ably/realtime_presence_integration_test.go
```diff
@@ -254,47 +254,6 @@ func SkipTestRealtimePresence_ServerSynthesized_Leave(t *testing.T) {
 	assert.Empty(t, members)
 }
 
-func TestRealtimePresence_ServerSynthesized_Leave(t *testing.T) {
```
sacOO7, line 257: This test is removed because unskipped test already covers all assertions with complete details

ttypic (approved):
LGTM

PR #662, opened 2024-08-19
Delete unnecessary recover error test

amnonbc:
The test removed in this commit expects that using an invalid recovery key leads to an ERROR protocol message being returned which thus fails the connection, but this isn't part of the spec, we make no guarantee that invalid recovery keys are rejected with an ERROR, and in fact returning a CONNECTED protocol message with the error field indicating the recovery key was invalid is a perfectly good response, which is what we will start to do in the future.

The RTN15c4 test was removed from ably-dotnet for this reason.
This PR removes it from ably-go.

### DIFF
--- a/ably/realtime_conn_spec_integration_test.go
+++ b/ably/realtime_conn_spec_integration_test.go
@@ -17,9 +17,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/ably/ably-go/ably"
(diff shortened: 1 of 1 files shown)


sacOO7 (commented):
Btw, this test is related to spec https://sdk.ably.com/builds/ably/specification/main/features/#RTN15c4. Are we planning to remove spec since we are deleting related integration tests? This also means we are changing server behaviour not to act like RTN15c4

sacOO7 (approved):
lgtm

PR #663, opened 2024-08-19
Add support for go1.23 and drop versions before 1.22

amnonbc:
Go 1.23 was released last week.
We follow the Go release policy of supporting the last two versions of Go.

See https://github.com/ably/ably-go#supported-versions-of-go

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ['1.19', '1.20', '1.21', '1.22']
+        go-version: ['1.19', '1.22', '1.23']
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ['1.19', '1.20', '1.21', '1.22']
+        go-version: ['1.19', '1.22', '1.23']
         protocol: ['json', 'msgpack']
 
     steps:


sacOO7:
Being active maintainer I would like to know reasons behind policy, considering there's lot of apps using older version of go and breaking changes for older versions will go undetected.

paddybyers:
We can't just drop support for older versions like that. We will use usage stats to determine who is using which versions and that will determine which versions are eligible to be dropped.

sacOO7:
> We can't just drop support for older versions like that. We will use usage stats to determine who is using which versions and that will determine which versions are eligible to be dropped.

Agreed, we don't drop older versions for other SDKs as well.

amnonbc:
> Being active maintainer I would like to know reasons behind policy, considering there's lot of apps using older version of go and breaking changes for older versions will go undetected.

Sure:
See https://go.dev/doc/devel/release#policy
and also https://go.dev/blog/compat

The Go team only maintain the last two major versions.
So versions 1.21 and earlier are obsolete and do not get bug fixes or fixes for security vulnerabilities.
Ably too does not support obsolete versions of the Go compiler.
So if a customer want to upgrade their ably-go version, they will need to upgrade to a non-obsolete version of 
the Go toolchain.

This policy is good for Ably as it means we do not need to test against a lot of obsolete toolchain versions.
And it is good for our customers as it will make them update to a supported version of the toolchain, 
so that they benefit from security and bug fixes.

We could force the toolchain upgrade on our customers by changing the Go version in https://go.dev/blog/compat
to 1.23.0. So if the customer has runs a Go 1.21 compiler,  it will download the Go 1.23.0 toolchain, and use that to build 
their app.

sacOO7 (approved):
Lgtm

amnonbc:
do we want to merge this?
(I don't have permission).
Also 1.24 will be released this month.

sacOO7:
> do we want to merge this? (I don't have permission). Also 1.24 will be released this month.

Let's wait for https://github.com/ably/ably-go/pull/679 to get merged, Tests on main are failing : (

PR #674, opened 2024-11-27
auth: Add option to allow basic auth for non-TLS requests

lmars:
We internally often want to use the Go SDK in private environments (e.g. local development, or within virtual private networks), where we do not use TLS, but find it convenient to be able to use API keys rather than tokens.

This commit adds an `InsecureAllowBasicAuthWithoutTLS` option which permits the use of an API key when the `NoTLS` option is also set.

We only require this in the Go SDK, so there is no intention of adding this option to the feature spec.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new option allowing basic authentication without TLS for enhanced flexibility.
	- Added a method to easily configure the new authentication option.
- **Bug Fixes**
	- Improved security checks for basic authentication over non-TLS connections.
- **Tests**
	- Added tests to ensure correct behavior of basic authentication options with and without TLS.

### DIFF
--- a/ably/auth.go
+++ b/ably/auth.go
@@ -545,7 +545,7 @@ func detectAuthMethod(opts *clientOptions) (int, error) {
 	if !isKeyValid {
 		return 0, newError(ErrInvalidCredential, errInvalidKey)
 	}
-	if opts.NoTLS {
+	if opts.NoTLS && !opts.InsecureAllowBasicAuthWithoutTLS {
 		return 0, newError(ErrInvalidUseOfBasicAuthOverNonTLSTransport, errInsecureBasicAuth)
 	}
 	return authBasic, nil
--- a/ably/options.go
+++ b/ably/options.go
@@ -408,6 +408,10 @@ type clientOptions struct {
 	// LogHandler controls the log output of the library. This is a function to handle each line of log output.
 	// platform specific (TO3c)
 	LogHandler Logger
+
+	// InsecureAllowBasicAuthWithoutTLS permits an API key to be used even if the connection
+	// will not use TLS, something which would otherwise not be permitted for security reasons.
+	InsecureAllowBasicAuthWithoutTLS bool
 }
 
 func (opts *clientOptions) validate() error {
@@ -1313,6 +1317,14 @@ func WithFallbackHostsUseDefault(fallbackHostsUseDefault bool) ClientOption {
 func WithDial(dial func(protocol string, u *url.URL, timeout time.Duration) (conn, error)) ClientOption {
 	return func(os *clientOptions) {
 		os.Dial = dial
+	}
+}
+
+// WithInsecureAllowBasicAuthWithoutTLS permits an API key to be used even if the connection
+// will not use TLS, something which would otherwise not be permitted for security reasons.
+func WithInsecureAllowBasicAuthWithoutTLS() ClientOption {
+	return func(opts *clientOptions) {
+		opts.InsecureAllowBasicAuthWithoutTLS = true
 	}
 }
 
--- a/ably/options_test.go
+++ b/ably/options_test.go
@@ -236,6 +236,28 @@ func TestScopeParams(t *testing.T) {
 	})
 }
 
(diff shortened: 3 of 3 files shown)


sacOO7:
Not sure why so many tests got flaky suddenly : ( . We always had green CI for every PR before.
One of the common test I can see failing is `TestRealtimeConn_RTN15i_OnErrorWhenConnected`
@lmars can you check what might be going wrong?

sacOO7 (commented):
Not sure why so many tests got flaky suddenly : ( . 
We always had green CI for every PR before.
One of the common test I can see failing is `TestRealtimeConn_RTN15i_OnErrorWhenConnected`
@lmars can you check what might be going wrong?

lmars:
@sacOO7 the failing test is `TestRealtimeConn_RTN16` which I have proposed a fix for in https://github.com/ably/ably-go/pull/675.

sacOO7:
@lmars we are planning to create one more PR to fix present enter 250 test right?

lmars:
@sacOO7 I've created a ticket for that here:

https://github.com/ably/ably-go/issues/676

It needs to be fixed, but I don't have an immediate plan to fix it.

Given this PR is unrelated to that failure, please can you review it so I can merge?

ably/options.go
```diff
@@ -408,6 +408,10 @@ type clientOptions struct {
 	// LogHandler controls the log output of the library. This is a function to handle each line of log output.
 	// platform specific (TO3c)
 	LogHandler Logger
+
+	// InsecureAllowBasicAuthWithoutTLS permits an API key to be used even if the connection
+	// will not use TLS, something which would otherwise not be permitted for security reasons.
+	InsecureAllowBasicAuthWithoutTLS bool
```
sacOO7, line 414: Can we make this assertive =>
`AllowBasicAuthWithoutTLS`

ably/auth.go
```diff
@@ -545,7 +545,7 @@ func detectAuthMethod(opts *clientOptions) (int, error) {
 	if !isKeyValid {
 		return 0, newError(ErrInvalidCredential, errInvalidKey)
 	}
-	if opts.NoTLS {
+	if opts.NoTLS && !opts.InsecureAllowBasicAuthWithoutTLS {
```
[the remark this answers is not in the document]
sacOO7: This is very easy to understand

ably/options.go
sacOO7:
```suggestion
	AllowBasicAuthWithoutTLS bool
```

sacOO7 (commented):
Suggested a small renaming change, so it's easier to understand while reading the code.
Also, would be good if you can add small unit/integration test for this 👍

ably/options.go
lmars: The point of the "insecure" prefix is to make it clear to users that this is an insecure option, it's a common approach (e.g. `insecureSkipVerify` to disable TLS verification, or `WithInsecureCredentials` to use plain text credentials with gRPC).

sacOO7: Okay, if it's only let users know that it's `insecure` option, default value for the same should be `false` right?

sacOO7: Seems default value of `InsecureSkipVerify` is set to `false`. It will remain the same if we interpret it as `SkipVerify`. I think prefixing with `Insurecure` shouldn't invert the underlying value

sacOO7: I think prefixing with `Insurecure` shouldn't invert the underlying value

sacOO7: Okay, seems by default it's set to `false`, I was confused with `WithInsecureAllowBasicAuthWithoutTLS` setting it to default value

sacOO7 (approved):
LGTM, can you add unit/integration test before merging the PR please?

sacOO7:
Simple test can be added to `options_test.go`, using `noTls=true` over `basic auth` returns error.

lmars:
@sacOO7 test added in 631f76c.

PR #675, opened 2024-11-27
tests: Don't assert connection key equality

lmars:
The format of the connection key is not public, and we reserve the right to change it server-side as we please, the only functionality the client expects is that providing the connection key of a previous connection results in the new connection having the same id, so that's all we need to test.

### DIFF
--- a/ably/realtime_conn_spec_integration_test.go
+++ b/ably/realtime_conn_spec_integration_test.go
@@ -1947,9 +1947,8 @@ func TestRealtimeConn_RTN16(t *testing.T) {
 	assert.NoError(t, err)
 
 	{ // RTN16f, RTN16j, RTN16d
-		assert.True(t, sameConnection(client.Connection.Key(), c.Connection.Key()),
+		assert.Equal(t, prevConnId, c.Connection.ID(),
 			"expected the same connection")
-		assert.Equal(t, prevConnId, c.Connection.ID())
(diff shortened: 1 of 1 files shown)


lmars:
@sacOO7 please can you review this?

sacOO7:
Yes, will take a look soon 👍

sacOO7:
It seems, there's one more test consistently failing
```
TestRealtimePresence_Sync250_RTP4
```
This test is causing job to run for ~2 hrs
Most of the time, error is 

`
[ErrorInfo :Rate limit exceeded; request rejected (nonfatal); metric = channel.maxRate; interval = 2024-11-27:17:02:8; permitted rate = 250; current rate = 266 code=42910 rate limit exceeded (nonfatal): request rejected (unspecified) statusCode=429] See https://help.ably.io/error/42910
`

sacOO7:
We never faced an issue like this before. `TestRealtimePresence_Sync250_RTP4` always used to pass.

lmars:
@sacOO7 ok, but I think that's unrelated to this change, I think we should tackle the issues one at a time, the change here definitely fixes `TestRealtimeConn_RTN16` which is failing consistently in main.

sacOO7:
Sure, I will approve this PR.
Just that I was worried `main` will be `red` once we merge this PR.

sacOO7 (approved):
LGTM. 
Considering we will fix other failing tests soon 👍

PR #677, opened 2025-01-06
[INF-5347] Update to endpoint client option

surminus:
## Summary

This implements ADR-119, which specifies the client connection options to update requests to the endpoints implemented as part of ADR-042.

The endpoint may be one of the following:

* a routing policy name (such as `main`)
* a nonprod routing policy name (such as `nonprod:sandbox`)
* a FQDN such as `foo.example.com`

The endpoint option is not valid with any of environment, restHost or realtimeHost, but we still intend to support the legacy options.

If the client has been configured to use any of these legacy options, then they should continue to work in the same way, using the same primary and fallback hostnames.

If the client has not been explicitly configured, then the hostnames will change to the new `ably.net` domain when the package is upgraded.

## Questions

Since this is my first stab at making changes to an SDK, I require a little guidance:

* What should I be doing with the feature spec references? Do I update all of them? Which are the correct ones to apply?
* I've tried to add tests where I think they are appropriate, but do I need to add any additional ones? Since we're using a test recorder which we're using a fqdn with the endpoint, do we need to test against sandbox? 
* Given this is likely a major version release, what is the process for this? 

## Summary by CodeRabbit

- **New Features**
	- Introduced advanced endpoint configuration options for the Ably client.
	- Added support for custom routing policies and fallback hosts.
	- Enhanced connection management with more flexible host selection.

- **Bug Fixes**
	- Improved host connection logic to handle multiple fallback scenarios.
	- Updated validation for endpoint configurations.

- **Tests**
	- Added comprehensive test coverage for new endpoint and host management features.
	- Expanded integration tests to validate connection behaviors.
	- Restructured tests to accommodate new endpoint logic and ensure clarity in testing.
	- Updated tests to reflect changes in endpoint configuration and expected outcomes.

### DIFF
--- a/ably/options.go
+++ b/ably/options.go
@@ -70,6 +70,10 @@ func defaultFallbackHosts() []string {
 }
 
 func getEndpointFallbackHosts(endpoint string) []string {
+	if endpoint == "sandbox" {
+		return endpointFallbacks("sandbox", "ably-realtime-nonprod.com")
+	}
+
 	if match := nonprodRegexp.FindStringSubmatch(endpoint); match != nil {
 		namespace := match[1]
 		return endpointFallbacks(namespace, "ably-realtime-nonprod.com")
@@ -84,16 +88,6 @@ func endpointFallbacks(namespace, root string) []string {
 		fallbacks = append(fallbacks, fmt.Sprintf("%s.%s.fallback.%s", namespace, id, root))
 	}
 	return fallbacks
-}
-
-func getEnvFallbackHosts(env string) []string {
-	return []string{
-		fmt.Sprintf("%s-%s", env, "a-fallback.ably-realtime.com"),
-		fmt.Sprintf("%s-%s", env, "b-fallback.ably-realtime.com"),
-		fmt.Sprintf("%s-%s", env, "c-fallback.ably-realtime.com"),
-		fmt.Sprintf("%s-%s", env, "d-fallback.ably-realtime.com"),
-		fmt.Sprintf("%s-%s", env, "e-fallback.ably-realtime.com"),
-	}
 }
 
 const (
@@ -262,17 +256,16 @@ type clientOptions struct {
 	// authOptions Embedded an [ably.authOptions] object (TO3j).
 	authOptions
 
-	// Endpoint specifies the domain used to connect to Ably. It has the following properties:
-	// If the endpoint option is a production routing policy name of the form [name] then the primary domain is [name].realtime.ably.net (REC1b4).
-	// If the endpoint option is a non-production routing policy name of the form nonprod:[name] then the primary domain is [name].realtime.ably-nonprod.net (REC1b3).
-	// If the endpoint option is a domain name, determined by it containing at least one period (.) then the primary domain is the value of the endpoint option (REC1b2).
+	// Endpoint specifies the domain used to connect to Ably (REC1a).
+	// If the endpoint option is specified then (REC1b):
 	// If any one of the deprecated options environment, restHost, realtimeHost are also specified then the options as a set are invalid (REC1b1).
+	// If the endpoint option is a domain name, determined by it containing at least one period (.) then the primary domain is the value of the endpoint option (REC1b2).
+	// Otherwise, if the endpoint option is a non-production routing policy name of the form nonprod:[name] then the primary domain is [name].realtime.ably-nonprod.net (REC1b3).
+	// Otherwise, if the endpoint option is a production routing policy name of the form [name] then the primary domain is [name].realtime.ably.net (REC1b4).
 	Endpoint string
 
 	// Deprecated: this property is deprecated and will be removed in a future version.
-	// RESTHost enables a non-default Ably host to be specified. For
-	// development environments only. The default value is rest.ably.io (RSC12,
-	// TO3k2).
+	// If the restHost option is specified the primary domain is the value of the restHost option REC1d1).
 	RESTHost string
 
 	// Deprecated: this property is deprecated and will be removed in a future version.
@@ -285,13 +278,13 @@ type clientOptions struct {
 	FallbackHosts []string
 
 	// Deprecated: this property is deprecated and will be removed in a future version.
-	// RealtimeHost enables a non-default Ably host to be specified for realtime connections.
-	// For development environments only. The default value is realtime.ably.io (RTC1d, TO3k3).
+	// If the realtimeHost option is specified the primary domain is the value of the realtimeHost option (REC1d2).
 	RealtimeHost string
 
 	// Deprecated: this property is deprecated and will be removed in a future version.
-	// Environment enables a custom environment to be used with the Ably service.
-	// Optional: prefixes both hostname with the environment string (RSC15b, TO3k1).
+	// If the deprecated environment option is specified then it defines a production routing policy name [name] (REC1c):
+	// If any one of the deprecated options restHost, realtimeHost are also specified then the options as a set are invalid (REC1c1).
+	// Otherwise, the primary domain is [name].realtime.ably.net (REC1c2).
 	Environment string
 
 	// Port is used for non-TLS connections and requests
@@ -460,11 +453,6 @@ func (opts *clientOptions) validate() error {
 	return nil
 }
 
-func (opts *clientOptions) isProductionEnvironment() bool {
-	env := opts.Environment
-	return empty(env) || strings.EqualFold(env, "production")
-}
-
 func (opts *clientOptions) activePort() (port int, isDefault bool) {
 	if opts.NoTLS {
 		port = opts.Port
@@ -486,65 +474,54 @@ func (opts *clientOptions) activePort() (port int, isDefault bool) {
 	return
 }
 
-func (opts *clientOptions) usingLegacyOpts() bool {
-	return empty(opts.Endpoint) && (!empty(opts.Environment) || !empty(opts.RESTHost) || !empty(opts.RealtimeHost))
-}
-
 // endpointFqdn handles an endpoint that uses a hostname, which may be an IPv4
 // address, IPv6 address or localhost
 func endpointFqdn(endpoint string) bool {
 	return strings.Contains(endpoint, ".") || strings.Contains(endpoint, "::") || endpoint == "localhost"
 }
 
-func (opts *clientOptions) getEndpoint() string {
-	endpoint := opts.Endpoint
-	if empty(endpoint) {
-		endpoint = defaultOptions.Endpoint
+func (opts *clientOptions) endpointValueWithLegacySupport() string {
+	if !empty(opts.Endpoint) {
+		return opts.Endpoint
 	}
 
-	if endpointFqdn(opts.Endpoint) {
-		return endpoint
-	}
+	if !empty(opts.Environment) {
+		if opts.Environment == "production" {
+			return defaultOptions.Endpoint
+		}
 
-	if match := nonprodRegexp.FindStringSubmatch(endpoint); match != nil {
-		namespace := match[1]
-		return fmt.Sprintf("%s.realtime.ably-nonprod.net", namespace)
-	}
-
-	return fmt.Sprintf("%s.realtime.ably.net", endpoint)
-}
-
-func (opts *clientOptions) getRestHost() string {
-	if !opts.usingLegacyOpts() {
-		return opts.getEndpoint()
+		return opts.Environment
 	}
 
 	if !empty(opts.RESTHost) {
 		return opts.RESTHost
-	}
-	if !opts.isProductionEnvironment() {
-		return opts.Environment + "-" + defaultOptions.RESTHost
-	}
-	return defaultOptions.RESTHost
-}
-
-func (opts *clientOptions) getRealtimeHost() string {
-	if !opts.usingLegacyOpts() {
-		return opts.getEndpoint()
 	}
 
 	if !empty(opts.RealtimeHost) {
 		return opts.RealtimeHost
 	}
-	if !empty(opts.RESTHost) {
-		logger := opts.LogHandler
-		logger.Printf(LogWarning, "restHost is set to %s but realtimeHost is not set so setting realtimeHost to %s too. If this is not what you want, please set realtimeHost explicitly.", opts.RESTHost, opts.RealtimeHost)
-		return opts.RESTHost
+
+	return defaultOptions.Endpoint
+}
+
+// REC2
+func (opts *clientOptions) getEndpoint() string {
+	ep := opts.endpointValueWithLegacySupport()
+
+	if endpointFqdn(ep) {
+		return ep
 	}
-	if !opts.isProductionEnvironment() {
-		return opts.Environment + "-" + defaultOptions.RealtimeHost
+
+	if ep == "sandbox" {
+		return "sandbox.realtime.ably-nonprod.net"
 	}
-	return defaultOptions.RealtimeHost
+
+	if match := nonprodRegexp.FindStringSubmatch(ep); match != nil {
+		namespace := match[1]
+		return fmt.Sprintf("%s.realtime.ably-nonprod.net", namespace)
+	}
+
+	return fmt.Sprintf("%s.realtime.ably.net", ep)
 }
 
 func empty(s string) bool {
@@ -552,7 +529,7 @@ func empty(s string) bool {
 }
 
 func (opts *clientOptions) restURL() (restUrl string) {
-	baseUrl := opts.getRestHost()
+	baseUrl := opts.getEndpoint()
 	_, _, err := net.SplitHostPort(baseUrl)
 	if err != nil { // set port if not set in baseUrl
 		port, _ := opts.activePort()
@@ -595,18 +572,12 @@ func (opts *clientOptions) getFallbackHosts() ([]string, error) {
 		return defaultOptions.FallbackHosts, nil
 	}
 
-	if opts.FallbackHosts == nil && !empty(opts.Endpoint) {
-		if endpointFqdn(opts.Endpoint) {
+	if opts.FallbackHosts == nil {
+		ep := opts.endpointValueWithLegacySupport()
+		if endpointFqdn(ep) {
 			return opts.FallbackHosts, nil
 		}
-		return getEndpointFallbackHosts(opts.Endpoint), nil
-	}
-
-	if opts.FallbackHosts == nil && empty(opts.RESTHost) && empty(opts.RealtimeHost) && isDefaultPort {
-		if opts.isProductionEnvironment() {
-			return defaultOptions.FallbackHosts, nil
-		}
-		return getEnvFallbackHosts(opts.Environment), nil
+		return getEndpointFallbackHosts(ep), nil
 	}
 
 	return opts.FallbackHosts, nil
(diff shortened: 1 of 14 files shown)


sacOO7:
Okay, looks like a big PR, will review by today EOD or tomorrow morning.

surminus:
> Okay, looks like a big PR, will review by today EOD or tomorrow morning.

Thanks, I don't think there's a huge rush, but given the implications I need to make sure it works as intended :)

sacOO7:
Okay, it seems we have standard spec for this as per https://sdk.ably.com/builds/ably/specification/main/features/#REC1.
We need to annotate relevant code accordingly. 
You can just annotate via simple comment next to the code snippet like
```
endpoint = "main" // REC1a
```

ably/options.go
```diff
@@ -59,13 +64,24 @@ var defaultOptions = clientOptions{
...
-		"d.ably-realtime.com",
-		"e.ably-realtime.com",
+	return endpointFallbacks("main", "ably-realtime.com")
+}
+
+func getEndpointFallbackHosts(endpoint string) []string {
+	if match := regexp.MustCompile(`^nonprod:(.*)$`).FindStringSubmatch(endpoint); match != nil {
+		namespace := match[1]
+		return endpointFallbacks(namespace, "ably-realtime-nonprod.com")
 	}
```
amnonbc, line 74: compiling a regexp in Go is very expensive. (But once a regexp is compiled, searching is fast and guaranteed to execute in a linear time).
So we usually compile the regexp at init time.

```suggestion

var nonProdRegexp = regexp.MustCompile(`^nonprod:(.*)$`)

func getEndpointFallbackHosts(endpoint string) []string {
	if match := nonProdRegexp.FindStringSubmatch(endpoint); match != nil {
		namespace := match[1]
		return endpointFallbacks(namespace, "ably-realtime-nonprod.com")
	}
```

This also has the benefit that if we have a typo in the regexp, so that it fails to compile, this will always panic when the program starts (which is what we want), rather than only when getEndpointFallbackHosts is called.

ably/options.go
```diff
@@ -450,7 +484,39 @@ func (opts *clientOptions) activePort() (port int, isDefault bool) {
...
+	endpoint := opts.Endpoint
+	if empty(endpoint) {
+		endpoint = defaultOptions.Endpoint
+	}
+
+	if endpointFqdn(opts.Endpoint) {
+		return endpoint
+	}
+
+	if match := regexp.MustCompile(`^nonprod:(.*)$`).FindStringSubmatch(endpoint); match != nil {
```
amnonbc, line 507: ditto - and we can even reuse `nonProdRegexp`

ably/options.go
surminus: Fixed up in ee7c1b1bd8016bb2ec1029aeebc39af977cd1f5a

ably/options.go
surminus: Fixed up in ee7c1b1bd8016bb2ec1029aeebc39af977cd1f5a

surminus:
I think I've actually implemented this slightly incorrectly as per [REC1c])(https://sdk.ably.com/builds/ably/specification/main/features/#REC1c): 

> If the deprecated environment option is specified then it defines a production routing policy name [name]

Whereas I have implemented it not to change behaviour if environment is specified, so I need to update it.

ably/options_test.go
```diff
@@ -13,52 +13,65 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestDefaultFallbacks_RSC15h(t *testing.T) {
+func TestDefaultFallbacks_REC2c(t *testing.T) {
```
sacOO7, line 16: I don't think we should override existing tests for deprecated clientOptions, instead they should be updated.
This is to make sure, deprecated features also work as expected.
 
Can you instead append new tests at the end of the test file and mark old tests as deprecated.
This is to make sure both old and new clientOptions work as expected. 

In the future, when we decide to remove deprecated clientOptions, we will remove both deprecated clientOptions and related tests.

sacOO7 (changes requested):

ably/options_test.go
sacOO7: Unfortunately, the official [ably doc for clientOptions](https://ably.com/docs/api/realtime-sdk#client-options) and current README still recommends use of old clientOptions.

sacOO7: Those docs won't be updated until we update all SDKs to use `endpoint` instead of `environment`

surminus: > Can you instead append new tests at the end of the test file and mark old tests as deprecated.
This is to make sure both old and new clientOptions work as expected.

The changes here will change the behavour of the legacy clientOptions, so keeping the old tests as they were will mean they will fail. How should we handle this? 

> Unfortunately, the official [ably doc for clientOptions](https://ably.com/docs/api/realtime-sdk#client-options) and current README still recommends use of old clientOptions.

I can update the current README. For the SDK docs, this is the first client we are making this change on, and I am not an SDK developer, so will probably need some guidance and support for the process in updating docs, releasing new versions etc.

sacOO7: Ideally, newly introduced `clientOptions` shouldn't break deprecated `clientOptions`.
Does new spec introduces breaking changes for deprecated clientOptions?
I would recommend to update the code in a way that, if `endpoint` is not specified, hosts will be calculated based on deprecated clientOptions

sacOO7: Also, we have ongoing work for `chat`, `livesync`, updating all SDKs with `endpoint` behavior will take at least a year or so.. 
Also, not sure if we have internal priorities for updating clientOptions across SDKs.

sacOO7: I am also adding @SimonWoolf to review the PR

surminus: See the [discussion in the ADR](https://ably.atlassian.net/wiki/spaces/ENG/pages/3428810778/ADR-119+ClientOptions+for+new+DNS+structure#Issues-and-options) about updating the current behaviour of the clientOptions.

> Also, not sure if we have internal priorities for updating clientOptions across SDKs.

Yes, I guess that's why an SRE is doing this work rather than an SDK developer ;)

sacOO7: Okay, ADR is still in progress 🤔, and seems decided to go with breaking option 2. 
Do we have the spec updated accordingly since ADR is still in progress?
If implemented with option2 ( also make sure we have spec updated accordingly ), we also need to update [UPDATING.md](https://github.com/ably/ably-go/blob/main/UPDATING.md ) and will mark a major release. i.e. 1.3.0

surminus: The spec you linked to before does appear to be updated, yes. It describes the breaking change behaviour (that `environment` will become the routing policy name). 

> If implemented with option2 ( also make sure we have spec updated accordingly ), we also need to update [UPDATING.md](https://github.com/ably/ably-go/blob/main/UPDATING.md) and will mark a major release. i.e. 1.3.0

:+1: Is `1.3.0` a major release? I would have thought that would be `2.0.0`. Or is there some weird Ably context I'm not aware of :sweat_smile:

surminus: Having discussed this with @paddybyers and @SimonWoolf, we don't necessarily think this is a breaking change. The tests _do_ need to change because the returned hostnames will change, but it's not actually causing any impact to the user.

The only thing that might cause a conflict is if both `WithEnvironment` and `WithRealtimeHost` or `WithRESTHost` are used together which previously was just ignored, but now will throw an error. I've updated the integration tests accordingly to use `WithEndpoint`. 

With this in mind, do we still mark this at 1.3.0? We don't actually specify clientOptions anywhere in docs as far as I can see, so not sure what I need to update, other than `CHANGELOG.md`.

ably/options.go
```diff
@@ -58,24 +63,31 @@ var defaultOptions = clientOptions{
...
+	return endpointFallbacks("main", "ably-realtime.com")
+}
+
+func getEndpointFallbackHosts(endpoint string) []string {
+	if endpoint == "sandbox" {
+		return endpointFallbacks("sandbox", "ably-realtime-nonprod.com")
+	}
+
+	if match := nonprodRegexp.FindStringSubmatch(endpoint); match != nil {
+		namespace := match[1]
```
lmars, line 78:
```suggestion
	if strings.HasPrefix(endpoint, "nonprod:") {
		namespace := strings.TrimPrefix(endpoint, "nonprod:")
```

ably/options.go
```diff
@@ -58,24 +63,31 @@ var defaultOptions = clientOptions{
...
-	return []string{
-		fmt.Sprintf("%s-%s", env, "a-fallback.ably-realtime.com"),
-		fmt.Sprintf("%s-%s", env, "b-fallback.ably-realtime.com"),
-		fmt.Sprintf("%s-%s", env, "c-fallback.ably-realtime.com"),
-		fmt.Sprintf("%s-%s", env, "d-fallback.ably-realtime.com"),
-		fmt.Sprintf("%s-%s", env, "e-fallback.ably-realtime.com"),
+func endpointFallbacks(namespace, root string) []string {
+	var fallbacks []string
+	for _, id := range []string{"a", "b", "c", "d", "e"} {
+		fallbacks = append(fallbacks, fmt.Sprintf("%s.%s.fallback.%s", namespace, id, root))
```
lmars, line 88:
```suggestion
	fallbacks := make([]string, 5)
	for i, id := range []string{"a", "b", "c", "d", "e"} {
		fallbacks[i] = fmt.Sprintf("%s.%s.fallback.%s", namespace, id, root)
```

ably/options.go
```diff
@@ -450,37 +474,62 @@ func (opts *clientOptions) activePort() (port int, isDefault bool) {
 	return
 }
 
-func (opts *clientOptions) getRestHost() string {
+// endpointFqdn handles an endpoint that uses a hostname, which may be an IPv4
+// address, IPv6 address or localhost
+func endpointFqdn(endpoint string) bool {
```
lmars, line 479:
```suggestion
func endpointFQDN(endpoint string) bool {
```

See https://go.dev/wiki/CodeReviewComments#initialisms

lmars: Also, given this returns a bool, I suggest:

```suggestion
func isEndpointFQDN(endpoint string) bool {
```

ably/options.go
```diff
@@ -450,37 +474,62 @@ func (opts *clientOptions) activePort() (port int, isDefault bool) {
...
-	if !empty(opts.RESTHost) {
-		logger := opts.LogHandler
-		logger.Printf(LogWarning, "restHost is set to %s but realtimeHost is not set so setting realtimeHost to %s too. If this is not what you want, please set realtimeHost explicitly.", opts.RESTHost, opts.RealtimeHost)
-		return opts.RESTHost
+
+	return defaultOptions.Endpoint
+}
+
+// REC2
+func (opts *clientOptions) getEndpoint() string {
```
lmars, line 508: Given this is taking the configured endpoint, like `main`, and turning that into a hostname, like `main.realtime.ably.net`, perhaps this should be `getHostname()`?

ably/options.go
```diff
@@ -1070,9 +1123,23 @@ func WithEchoMessages(echo bool) ClientOption {
 	}
 }
 
+// WithEndpoint is used for setting Endpoint using [ably.ClientOption].
+// Endpoint specifies the domain used to connect to Ably (REC1a).
+// If the endpoint option is specified then (REC1b):
+// If any one of the deprecated options environment, restHost, realtimeHost are also specified then the options as a set are invalid (REC1b1).
+// If the endpoint option is a domain name, determined by it containing at least one period (.) then the primary domain is the value of the endpoint option (REC1b2).
+// Otherwise, if the endpoint option is a non-production routing policy name of the form nonprod:[name] then the primary domain is [name].realtime.ably-nonprod.net (REC1b3).
+// Otherwise, if the endpoint option is a production routing policy name of the form [name] then the primary domain is [name].realtime.ably.net (REC1b4).
```
lmars, line 1132: Given this is a user facing docstring (i.e. it will appear [somewhere here](https://pkg.go.dev/github.com/ably/ably-go/ably#ClientOption)), I don't think it should have all this implementation detail and cryptic `(RECXX)` strings, that should be included as code comments within the implementation (e.g. on `clientOptions.getHostname()`), I think we want something simple and user friendly here, for example:

```suggestion
// WithEndpoint sets a custom endpoint for connecting to the Ably service
// (see [Platform Customization] for more information).
//
// [Platform Customization]: https://ably.com/docs/platform-customization
```

ably/options.go
```diff
@@ -1070,9 +1123,23 @@ func WithEchoMessages(echo bool) ClientOption {
...
+		os.Endpoint = env
+	}
+}
+
 // WithEnvironment is used for setting Environment using [ably.ClientOption].
-// Environment enables a custom environment to be used with the Ably service.
-// Optional: prefixes both hostname with the environment string (RSC15b, TO3k1).
+// If the deprecated environment option is specified then it defines a production routing policy name [name] (REC1c):
+// If any one of the deprecated options restHost, realtimeHost are also specified then the options as a set are invalid (REC1c1).
+// Otherwise, the primary domain is [name].realtime.ably.net (REC1c2).
```
lmars, line 1142: Ditto for this, it's not very user friendly.

ably/options_test.go
lmars: I agree, this is not a breaking change, at least not from an API perspective, no options have been removed, we're adding a new `endpoint` option and changing the hostnames that the `environment` option will now map to, which is not something we document nor guarantee to never change, so this can be a patch release.

> With this in mind, do we still mark this at 1.3.0? We don't actually specify clientOptions anywhere in docs as far as I can see, so not sure what I need to update, other than CHANGELOG.md.

I don't think we necessarily need to bump the minor version, I think the only thing we need to do is mark `WithEnvironment`, `WithRESTHost` and `WithRealtimeHost` as deprecated with a [deprecation comment](https://go.dev/wiki/Deprecated) saying "use WithEndpoint instead".

ably/rest_client_integration_test.go
```diff
@@ -250,7 +250,6 @@ func TestRest_RSC7_AblyAgent(t *testing.T) {
 		assert.NoError(t, err)
 
 		opts := []ably.ClientOption{
-			ably.WithEnvironment(ablytest.Environment),
 			ably.WithTLS(false),
 			ably.WithUseTokenAuth(true),
 			ably.WithRESTHost(serverURL.Host),
```
lmars, line 255: Should this use `ably.WithEndpoint(serverURL.Host)`?

ably/options_test.go
sacOO7: Okay, [ADR](https://ably.atlassian.net/wiki/spaces/ENG/pages/3428810778/ADR-119+ClientOptions+for+new+DNS+structure#Issues-and-options) clearly states we planning to go with option 2 and it is a breaking change. 
Not sure, but there might be confusion here. If it's not a breaking change, can we update ADR + spec accordingly

sacOO7: Also, if we planning to bump up the version to 1.3.*, we might as well add support for encryption as per latest issue reported by the customer -> https://ably-real-time.slack.com/archives/CURL4U2FP/p1736763296808579

ably/options.go
```diff
@@ -450,37 +474,62 @@ func (opts *clientOptions) activePort() (port int, isDefault bool) {
...
+	ep := opts.endpointValueWithLegacySupport()
+
+	if endpointFqdn(ep) {
+		return ep
 	}
-	if !opts.isProductionEnvironment() {
-		return opts.Environment + "-" + defaultOptions.RealtimeHost
+
+	if ep == "sandbox" {
+		return "sandbox.realtime.ably-nonprod.net"
```
SimonWoolf, line 516: do we think having backcompat behaviour for "sandbox" is worth specifying so all sdks do it?

SimonWoolf (approved):
looks reasonable to me

ably/options.go
surminus: 90c9a2bcf57260c53ea42f5d93cbc688c3c493d7

ably/options.go
surminus: 90c9a2bcf57260c53ea42f5d93cbc688c3c493d7

ably/options.go
surminus: 90c9a2bcf57260c53ea42f5d93cbc688c3c493d7

ably/options.go
surminus: I've updated this particular comment but we do seem to litter the comments with references to specs. Is it worth doing a more general tidy up of how we reference the specifications? I don't know we do it elsewhere, but when I first looked at the codebase I had no idea what they were referring to, and if a non-Ably person wanted to contribute, I guess they would have the same experience.

Rather than littering the code with references to specific spec codes, would it be better to just have a link to the overall spec for the version of the release?

ably/rest_client_integration_test.go
surminus: 90c9a2bcf57260c53ea42f5d93cbc688c3c493d7

ably/options.go
surminus: When I work on other SDKs I intend to add the same behaviour, is this something that we would update the spec with? I have never done this.

SimonWoolf: Yeah if we want all sdks to do it consistently it should be specified -- it's just a textile file that lives here https://github.com/ably/specification/blob/main/textile/features.textile?plain=1#L72-L103 , you can just do a pr to add a new spec item. I can if you like, but given you're making sdk prs probably good to do it so you know the flow

sacOO7 (commented):
Just wanted to [hightlight the previous comment](https://github.com/ably/ably-go/pull/677#discussion_r1913035003).
Also, I think we shouldn't override tests for existing deprecated options since we are not introducing a breaking change, and users upgrading to latest version will most likely avoid to transitioning to use of `withEndpoint` immediately. 
Instead, can we add new tests right next to existing ones? Or at minimum keep some tests covering all deprecated clientOptions 
PS. Those tests will be removed, when we decide to remove deprecated clientOptions in the future. This should be true across all SDKs.

ably/options_test.go
surminus: > If it's not a breaking change, can we update ADR + spec accordingly?

I'm going to follow up with @paddybyers to get the ADR updated to be clear that it is decided, and that this is not a breaking change. I will release this as the next minor version.

surminus:
> Also, I think we shouldn't override tests for existing deprecated options since we are not introducing a breaking change, and users upgrading to latest version will most likely avoid to transitioning to use of withEndpoint immediately.
Instead, can we add new tests right next to existing ones?

We're agreeing that this is not a breaking change because we're not guaranteeing the stability of the hostname, but the hostname _does_ change, which means test assertions *do* need to be updated.

I have pushed updates so that the default way we're setting hostnames in tests is by using `WithEndpoint` because tests should instruct the user about the correct usage of the library, but I have also kept tests using the deprecated options where relevant (ie where we're asserting on the returned hostname) within a separate test context. 

I'm not sure if I see the value in repeating all tests twice with separate client options because the hostname being used will be exactly the same, especially when we're not asserting on the hostname being used.

With regards to the ADR, I will chase for that to be updated and confirmed that we do not consider this a breaking change.

sacOO7:
Okay, I have updated label from `breaking` to `enhancement`

ably/options.go
lmars: > Is it worth doing a more general tidy up of how we reference the specifications?

I think it is (although obviously not in this PR), since comments on exported types and methods are end user facing, and should be written with the end user in mind, no end user needs to see these spec point references, only SDK developers, so they should probably only appear within method definitions, and if there are any which end users might find useful, should probably be something like "For more information, see [REC1a](https://sdk.ably.com/builds/ably/specification/main/features/#REC1a) in the Ably SDK specification"

ably/options.go
```diff
@@ -450,37 +466,62 @@ func (opts *clientOptions) activePort() (port int, isDefault bool) {
 	return
 }
 
-func (opts *clientOptions) getRestHost() string {
+// isEndpointFQDN handles an endpoint that uses a hostname, which may be an IPv4
+// address, IPv6 address or localhost
```
lmars, line 470:
```suggestion
// isEndpointFQDN returns true if the given endpoint is a hostname, which may be an IPv4
// address, IPv6 address or localhost
```

ably/options.go
```diff
@@ -1070,9 +1115,23 @@ func WithEchoMessages(echo bool) ClientOption {
...
+// [Platform Customization] for more information).
+//
+// [Platform Customization]: https://ably.com/docs/platform-customization
+func WithEndpoint(env string) ClientOption {
+	return func(os *clientOptions) {
+		os.Endpoint = env
+	}
+}
+
+// WithEndpoint sets a custom endpoint for connecting to the Ably service (see
```
lmars, line 1128:
```suggestion
// WithEnvironment sets a custom endpoint for connecting to the Ably service (see
```

lmars (approved):
Couple of comment suggestions, otherwise LGTM

surminus:
There is just one integration test failing: `TestRealtime_RTN17_Integration_HostFallback_Timeout` 

I can't figure out why it would be timing out. It seems to correctly set the hostname with the provider URL.

sacOO7:
@surminus what IDE/Editor you are using locally? It will be great if you can debug the failing test. You can use vscode and use [go extension](https://marketplace.visualstudio.com/items?itemName=golang.Go) to do the same.

sacOO7:
I am going to properly review the PR again, once everything is in place.

surminus:
@sacOO7 I've done some digging and the reason why that particular test is timing out is because the auth request goes to the httptest server rather than to sandbox. I _assume_ this is because previously we use `WithRealtimeHost` which set the realtime endpoint, but not the REST endpoint which is used for auth. Now we're using `WithEndpoint`, it is setting both, so the auth request times out. 

Is this the expected behaviour? I'm not sure what the correct thing to do here is, since it feels like something is fundamentally changing, particularly if the tests themselves depend on it.

sacOO7:
@surminus I will take a look once free. Just put here the names of newly failing tests

surminus:
No problem, whenever you have time. The only failing test right now is `TestRealtime_RTN17_Integration_HostFallback_Timeout`.

surminus:
Closed in lieu of https://github.com/ably/ably-go/pull/679

PR #685, opened 2025-07-09
feat: Close old connection when new ones set

kaschula:
Prevoiusly old connections were just overwritten as a new connections had been established. This was causing a go-routine leak as the old connections still had running eventloops.

This commit now asyncrously closes these connections so the `Connection` type is not blocked waiting for an old connection to close.

The websocket library has a 5 second timeout on closing old connections so no timeout logic has been a applied to the new go routine closing the old connection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection handling to ensure previous WebSocket connections are properly closed without blocking new connections.

* **Tests**
  * Added an integration test to verify that previous connections are closed correctly when a new connection is established.

### DIFF
--- a/ably/realtime_conn.go
+++ b/ably/realtime_conn.go
@@ -729,8 +729,18 @@ func (c *Connection) lockIsActive() bool {
 
 func (c *Connection) setConn(conn conn) {
 	c.connMtx.Lock()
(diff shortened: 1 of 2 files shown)


lmars (approved):
LGTM

PR #688, opened 2025-07-21
First attempt at adding vcdiff delta support

paddybyers:
This is a wholly AI-generated proposal to add vcdiff delta support to ably-go, posted here for discussion

### DIFF
--- a/ably/VCDIFF_PLUGIN.md
+++ b/ably/VCDIFF_PLUGIN.md
@@ -1,0 +1,169 @@
+# VCDiff Plugin for Ably Go SDK
+
+This document describes how to use VCDiff (RFC 3284) delta compression with the Ably Go SDK.
+
+## Overview
+
+The VCDiff plugin enables automatic decoding of delta-compressed messages sent by Ably. This reduces bandwidth usage and improves performance for channels that transmit similar messages over time.
+
+## Quick Start
+
+```go
+package main
+
+import (
+    "context"
+    "github.com/ably/ably-go/ably"
+)
+
+func main() {
+    // Create the VCDiff plugin
+    vcdiffPlugin := ably.NewVCDiffPlugin()
+
+    // Create Ably client with the plugin
+    client, err := ably.NewRealtime(
+        ably.WithKey("your-api-key"),
+        ably.WithVCDiffPlugin(vcdiffPlugin),
+    )
+    if err != nil {
+        panic(err)
+    }
+    defer client.Close()
+
+    // Get a channel with delta encoding enabled
+    channel := client.Channels.Get("my-channel",
+        ably.ChannelWithParams("delta", "vcdiff"))
+
+    // Subscribe to messages - they will be automatically decoded
+    ctx := context.Background()
+    unsubscribe, err := channel.Subscribe(ctx, "", func(msg *ably.Message) {
+        // Messages are automatically decoded by the VCDiff plugin
+        fmt.Printf("Received decoded message: %v\n", msg.Data)
+    })
+    if err != nil {
+        panic(err)
+    }
+    defer unsubscribe()
+
+    // Attach to the channel
+    err = channel.Attach(ctx)
+    if err != nil {
+        panic(err)
+    }
+
+    // Publish messages - they will be delta-encoded by the server
+    err = channel.Publish(ctx, "event", map[string]interface{}{
+        "temperature": 23.5,
+        "humidity": 45.2,
+        "timestamp": time.Now().Unix(),
+    })
+    if err != nil {
+        panic(err)
+    }
+}
+```
+
+## Implementation Details
+
+### Plugin Interface
+
+The VCDiff plugin implements the `VCDiffDecoder` interface:
+
+```go
+type VCDiffDecoder interface {
+    Decode(delta []byte, base []byte) ([]byte, error)
+}
+```
+
+### VCDiffPlugin
+
+- **File**: `vcdiff_plugin.go`
+- **Usage**: `ably.NewVCDiffPlugin()`
+- **Description**: Production-ready implementation using the [ably/vcdiff-go](https://github.com/ably/vcdiff-go) library
+
+### Client Configuration
+
+Enable VCDiff support by:
+
+1. **Adding the plugin to the client**:
+   ```go
+   client, err := ably.NewRealtime(
+       ably.WithKey("your-api-key"),
+       ably.WithVCDiffPlugin(ably.NewVCDiffPlugin()),
+   )
+   ```
+
+2. **Enabling delta on channels**:
+   ```go
+   channel := client.Channels.Get("my-channel",
+       ably.ChannelWithParams("delta", "vcdiff"))
+   ```
+
+## Error Handling
+
+The VCDiff plugin handles several error conditions according to the Ably specification:
+
+### Error Code 40018: Delta Decode Failure
+- **Cause**: VCDiff decode operation failed
+- **Recovery**: Channel automatically reattaches to recover (RTL18)
+- **Example**: Corrupted delta data, version mismatch
+
+### Error Code 40019: Missing VCDiff Plugin
+- **Cause**: Delta message received but no VCDiff plugin configured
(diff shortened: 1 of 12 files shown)


ably/realtime_channel.go
```diff
@@ -896,6 +918,86 @@ func (c *RealtimeChannel) notify(msg *protocolMessage) {
...
+
+	// Decode message with delta support
+	var cipher channelCipher
+	if c.options != nil {
+		cipher, _ = (*protoChannelOptions)(c.options).GetCipher()
+	}
+	decodedMsg, newBasePayload, err := msg.withDecodedDataAndContext(cipher, c.decodingContext)
+	if err != nil {
+		// Check for specific delta-related errors
+		if containsErrorCode(err, 40018) || containsErrorCode(err, 40019) {
```
ttypic, line 938: Not a go expert, but searching for error code inside string doesn't look right

ttypic (commented):
Overall implementation looks pretty good for me it covers all key points:

- store the last processed message data and id
- check if last id matches the one received with delta compression 
- decode with the last processed message data
- in case of problems initiate reattach 

Small notes: I would name client option property decoder instead of plugin and I don't think searching code inside error message is a right way to implement decoder error handler

ably/proto_message.go
```diff
@@ -193,6 +241,116 @@ func (m Message) withDecodedData(cipher channelCipher) (Message, error) {
...
+	encodings := strings.Split(m.Encoding, "/")
+
+	for len(encodings) > 0 {
+		encoding := encodings[len(encodings)-1]
+		encodings = encodings[:len(encodings)-1]
+		switch encoding {
+		case encVCDiff:
+			// Handle vcdiff delta decoding (PC3, RTL18, RTL19, RTL20)
+			if ctx == nil || ctx.VCDiffPlugin == nil {
+				return m, nil, fmt.Errorf("missing VCdiff decoder plugin (code 40019)")
```
lmars, line 261: Should this continue, perhaps with a log message, rather than returning an error?

ably/proto_message.go
```diff
@@ -193,6 +241,116 @@ func (m Message) withDecodedData(cipher channelCipher) (Message, error) {
 	return m, nil
 }
 
+// withDecodedDataAndContext decodes message data with support for delta decoding (RTL19).
+func (m Message) withDecodedDataAndContext(cipher channelCipher, ctx *DecodingContext) (Message, []byte, error) {
```
lmars, line 245: Perhaps just call this `decode`?

ably/proto_message.go
```diff
@@ -193,6 +241,116 @@ func (m Message) withDecodedData(cipher channelCipher) (Message, error) {
...
+
+		case encUTF8:
+			d, err := coerceString(m.Data)
+			if err != nil {
+				return m, nil, err
+			}
+			m.Data = d
+
+		case encJSON:
+			d, err := coerceBytes(m.Data)
```
lmars, line 311: Wouldn't a message with `encoding: "json"` come through as a string?

ably/proto_message.go
```diff
@@ -193,6 +241,116 @@ func (m Message) withDecodedData(cipher channelCipher) (Message, error) {
...
+				d, err = cipher.Decrypt(d)
+				if err != nil {
+					return m, nil, fmt.Errorf("decrypting message data: %w", err)
+				}
+				m.Data = d
+			} else {
+				return m, nil, fmt.Errorf("unknown encoding %s", encoding)
+			}
+		}
+		m.Encoding = strings.Join(encodings, "/")
```
lmars, line 339: Rather than splitting and re-joining the string (which allocates), perhaps use [`strings.LastIndex`](https://pkg.go.dev/strings#LastIndex) to iteratively slice `m.Encoding`?

ably/proto_message.go
```diff
@@ -193,6 +241,116 @@ func (m Message) withDecodedData(cipher channelCipher) (Message, error) {
...
+	}
+
+	// For non-delta messages, store the final decoded data as base payload (RTL19b)
+	if basePayloadForReturn == nil {
```
lmars, line 343: It's a bit unclear what this is for, doesn't the returned Message.Data field have the decoded data?

ably/realtime_channel.go
lmars: Yes I agree, we should either type assert the error to an ErrorInfo, or return specific error values and use [`errors.Is`](https://pkg.go.dev/errors#Is).

ably/realtime_channel.go
```diff
@@ -896,6 +918,86 @@ func (c *RealtimeChannel) notify(msg *protocolMessage) {
...
+		}
+		// For other errors, just log and continue (RSL6b)
+		c.client.log().Warnf("Message decode error on channel %q: %v", c.Name, err)
+		// Still update the message with partial decoding
+		*msg = decodedMsg
+		return nil
+	}
+
+	// Update message with decoded data
+	*msg = decodedMsg
```
lmars, line 949: It looks like `withDecodedDataAndContext` returns the same message, but just mutates that message's `Data` field, so this here seems redundant, `msg` and `decodedMsg` point to the same thing.

ably/realtime_channel.go
```diff
@@ -896,6 +918,86 @@ func (c *RealtimeChannel) notify(msg *protocolMessage) {
...
+
+	return nil
+}
+
+// startDecodeFailureRecovery implements the RTL18 recovery procedure.
+func (c *RealtimeChannel) startDecodeFailureRecovery(reason error) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	c.client.log().Errorf("Starting decode failure recovery for channel %q", c.Name)
```
lmars, line 965: This should probably include the error in the log message.

ably/vcdiff_plugin.go
```diff
@@ -0,0 +1,26 @@
...
+// NewVCDiffPlugin creates a new VCDiff plugin instance.
+func NewVCDiffPlugin() *VCDiffPlugin {
+	return &VCDiffPlugin{}
+}
+
+// Decode implements the VCDiffDecoder interface (VD2a, PC3a).
+// It uses the vcdiff-go library to decode VCDiff delta data.
+func (p *VCDiffPlugin) Decode(delta []byte, base []byte) ([]byte, error) {
+	// Use the vcdiff-go library to decode the delta
+	result, err := vcdiff.Decode(base, delta)
```
lmars, line 20: It's a bit odd that this has the args the other way around.

ably/vcdiff_plugin.go
```diff
@@ -0,0 +1,26 @@
...
+	if err != nil {
+		return nil, err
+	}
+	
+	return result, nil
```
lmars, line 25:
```suggestion
	return vcdiff.Decode(base, delta)
```

go.mod
```diff
@@ -1,8 +1,9 @@
 module github.com/ably/ably-go
 
-go 1.18
+go 1.22.3
```
lmars, line 3: Has this been bumped because we now require features only present in Go 1.22?

lmars (commented):
I've reviewed this the same way I'd review it if a human had written the code, so perhaps see what AI thinks about my comments 🙂

ably/proto_message.go
ttypic: I think it shouldn't, when decoder error happens we stop processing messages and initiate reattach instead

ably/VCDIFF_PLUGIN.md
```diff
@@ -0,0 +1,169 @@
...
+package main
+
+import (
+    "context"
+    "github.com/ably/ably-go/ably"
+)
+
+func main() {
+    // Create the VCDiff plugin
+    vcdiffPlugin := ably.NewVCDiffPlugin()
```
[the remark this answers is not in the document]
sacOO7: At the same time, will be good if we expose interface that others can implement in order to provide their own implementation, so it becomes pluggable

ttypic:
moved to https://github.com/ably/ably-go/pull/690

PR #693, opened 2025-10-04
fix: Prevent deadlock when releasing a realtime channel

lmars:
The RealtimeChannels mutex is acquired to retrieve the channel relating to an inbound message, so holding that mutex whilst waiting for Detach to complete causes a deadlock because the inbound DETACHED message the Detach call is waiting for also tries to acquire the mutex.

This commit fixes the deadlock by releasing the mutex before calling Detach, and then acquiring it again once it completes in order to release the channel.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when releasing realtime channels: channels now detach cleanly before removal, preventing rare deadlocks and ensuring resources are fully freed.
  * After release, channels are correctly removed and no longer accessible.

* **Tests**
  * Added an integration test validating the full channel release lifecycle, including attach/detach state transitions and post-release cleanup.


sacOO7 (approved):
LGTM

PR #700, opened 2026-03-24
Only run tests on the latest two Go versions

lmars:
The integration tests are quite heavyweight, and running them 10 times in parallel (i.e. JSON + msgpack across 5 Go versions) adds unnecessary load to the sandbox cluster.

Go has a backwards compatibility guarantee between versions, and also only officially supports the two latest Go versions, so it's sufficient for us to only test on the latest two Go versions in CI.

## Summary by CodeRabbit

* **Chores**
  * Updated continuous integration testing configuration to align with current Go version support standards.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -16,7 +16,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


owenpearson (approved):
sgtm 👍

PR #705, opened 2026-06-09
Provision a single shared sandbox app per test run

SimonWoolf:
## Summary

Integration tests previously provisioned a fresh sandbox app for **every test** (~150 per run) via the per-call-site `NewSandbox`/`NewRealtime`/`NewREST` helpers. Each provision creates a whole account and app server-side, which is expensive and puts needless load on the sandbox environment — and no other Ably SDK does this (ably-js provisions one app per suite from a shared appspec).

This PR provisions **one shared app per test run** instead, modelled on ably-js's approach.

## Commits

1. **`fix:` reset presence decoder destination between paginated pages** — a pre-existing SDK bug (not test-only). `fullPresenceDecoder` reused the same destination slice across pages in the `Items()` auto-paging iterator; under msgpack the codec reuses existing slice elements, so page 2+ decoded into `PresenceMessage`s whose `Data` still held the previous page's decoded values. With non-uniform `Data` types this corrupted decoding (`msgpack decode error: invalid byte descriptor for decoding bytes`). Fixed by resetting the destination to nil before decoding. Affects any user paginating presence via `Items()` over msgpack with mixed data.
2. **`chore:` bump ably-common submodule** — to a revision whose `test-app-setup.json` includes the `mutable` namespace (needed by the message-update tests) alongside the shared presence fixtures.
3. **`test:` provision a single shared sandbox app per run** — the refactor itself.

## What the refactor does (`ablytest/`)

- **`sandbox.go`**: drops the dynamic `Config`/`Key`/`Namespace`/`Presence`/`Channel` request-body structs (no test ever passed a custom config) and POSTs the static ably-common appspec verbatim, located via `runtime.Caller`. A minimal `Config` decodes only `appId` and `keys` from the response.
- Adds an **`Ably-Agent` header** (RSC7d) to the `/apps` POST and DELETE, reusing the client's agent string via the newly exported `ably.AgentIdentifier` / `ably.AblyAgentHeaderName`.
- **`GetSharedApp`** provisions the shared app once (`sync.Once`); `NewSandbox(nil)`, `MustSandbox`, `NewRealtime`, `NewREST` all return it. `Sandbox.Close` is a no-op unless the app is `owned`; teardown happens once via `CloseSharedApp` from a new `TestMain`.
- **`KeyParts`** selects the key carrying the `[*]*` wildcard capability rather than `keys[0]`, since the shared appspec's first key has only the default capability (which doesn't grant qualified/derived channels).
- Adds **`ablytest.ChannelName`** for process-unique channel names, used in the history/presence tests that assert on aggregate channel state and can no longer rely on per-app isolation.
- **`PresenceFixtures`** now matches the appspec's six members, including the AES-encrypted `client_encoded`, decoded in the presence-fixtures read tests via the new `PresenceFixturesCipher` helper.

## Testing

The full integration suite passes under **both** the JSON and msgpack protocols.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed paginated presence-history decoding to prevent stale `Data` values from carrying over across pages.

* **Improvements**
  * Exposed the SDK’s Ably-Agent header name and added a public helper for generating the agent identifier string.
  * Expanded built-in error code message mappings for clearer diagnostics.

* **Tests**
  * Refined integration test sandbox lifecycle/cleanup and added process-unique test channel naming to reduce interference.

### DIFF
--- a/ably/realtime_conn_integration_test.go
+++ b/ably/realtime_conn_integration_test.go
@@ -121,6 +121,10 @@ func TestRealtimeConn_ReceiveTimeout(t *testing.T) {
 		ably.WithRealtimeRequestTimeout(10*time.Millisecond),
 		ably.WithAutoConnect(false),
 	)
+	// Best-effort close to stop the connection's background reconnect loop. This
+	// uses a MessagePipe transport that can't complete a CLOSE handshake, so
+	// don't wait for the CLOSED state (FullRealtimeCloser) — that wait flakes.
+	defer client.Close()
 
 	states := make(ably.ConnStateChanges, 10)
 	{
--- a/ably/realtime_conn_integration_test.go
+++ b/ably/realtime_conn_integration_test.go
@@ -248,7 +248,7 @@ func TestRealtimeConn_SendErrorReconnects(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	go func() {
-		e := c.Channels.Get("test").Publish(ctx, "test", nil)
+		e := c.Channels.Get(ablytest.UniqueChannelName(t, "test")).Publish(ctx, "test", nil)
 		publishErr <- e
 	}()
 
--- a/ably/realtime_conn_integration_test.go
+++ b/ably/realtime_conn_integration_test.go
@@ -25,8 +25,8 @@ var connTransitions = []ably.ConnectionState{
 
 func TestRealtimeConn_AutoConnect_And_Close(t *testing.T) {
 	var rec ablytest.ConnStatesRecorder
-	app, client := ablytest.NewRealtime()
-	defer safeclose(t, ablytest.FullRealtimeCloser(client), app)
+	_, client := ablytest.NewRealtime()
+	defer safeclose(t, ablytest.FullRealtimeCloser(client))
 	off := rec.Listen(client)
 	defer off()
 
@@ -48,8 +48,8 @@ func TestRealtimeConn_No_AutoConnect(t *testing.T) {
 	opts := []ably.ClientOption{
 		ably.WithAutoConnect(false),
 	}
-	app, client := ablytest.NewRealtime(opts...)
-	defer safeclose(t, ablytest.FullRealtimeCloser(client), app)
+	_, client := ablytest.NewRealtime(opts...)
+	defer safeclose(t, ablytest.FullRealtimeCloser(client))
 	off := rec.Listen(client)
 	defer off()
 
@@ -68,8 +68,8 @@ func TestRealtimeConn_No_AutoConnect(t *testing.T) {
 }
 
 func TestRealtimeConn_AlreadyConnected(t *testing.T) {
-	app, client := ablytest.NewRealtime(ably.WithAutoConnect(false))
-	defer safeclose(t, ablytest.FullRealtimeCloser(client), app)
+	_, client := ablytest.NewRealtime(ably.WithAutoConnect(false))
+	defer safeclose(t, ablytest.FullRealtimeCloser(client))
 
 	err := ablytest.Wait(ablytest.ConnWaiter(client, client.Connect, ably.ConnectionEventConnected), nil)
 	assert.NoError(t, err,
@@ -116,12 +116,11 @@ func TestRealtimeConn_ReceiveTimeout(t *testing.T) {
 	}
 	in <- connected
 
-	app, client := ablytest.NewRealtime(
+	_, client := ablytest.NewRealtime(
 		ably.WithDial(MessagePipe(in, out, MessagePipeWithNowFunc(time.Now))),
 		ably.WithRealtimeRequestTimeout(10*time.Millisecond),
 		ably.WithAutoConnect(false),
 	)
-	defer safeclose(t, app)
 
 	states := make(ably.ConnStateChanges, 10)
 	{
@@ -167,10 +166,10 @@ func TestRealtimeConn_BreakConnLoopOnInactiveState(t *testing.T) {
 			in := make(chan *ably.ProtocolMessage)
 			out := make(chan *ably.ProtocolMessage, 16)
 
-			app, client := ablytest.NewRealtime(
+			_, client := ablytest.NewRealtime(
 				ably.WithDial(MessagePipe(in, out)),
 			)
-			defer safeclose(t, ablytest.FullRealtimeCloser(client), app)
+			defer safeclose(t, ablytest.FullRealtimeCloser(client))
 
 			connected := &ably.ProtocolMessage{
 				Action:            ably.ActionConnected,
@@ -228,8 +227,8 @@ func TestRealtimeConn_SendErrorReconnects(t *testing.T) {
 		}, nil
 	})
 
-	app, c := ablytest.NewRealtime(ably.WithDial(dial))
-	defer safeclose(t, ablytest.FullRealtimeCloser(c), app)
+	_, c := ablytest.NewRealtime(ably.WithDial(dial))
+	defer safeclose(t, ablytest.FullRealtimeCloser(c))
 
 	allowDial <- struct{}{}
 
@@ -294,12 +293,12 @@ func TestRealtimeConn_ReconnectFromSuspendedState(t *testing.T) {
 	dialErr <- nil
 	msgReceiveErr <- nil
 
-	app, c := ablytest.NewRealtime(ably.WithDial(dial),
+	_, c := ablytest.NewRealtime(ably.WithDial(dial),
 		ably.WithDisconnectedRetryTimeout(time.Second),
 		ably.WithSuspendedRetryTimeout(time.Second))
 	defer func() {
 		msgReceiveErr <- nil // receive safe close event
-		safeclose(t, ablytest.FullRealtimeCloser(c), app)
+		safeclose(t, ablytest.FullRealtimeCloser(c))
 	}()
 
 	err := ablytest.Wait(ablytest.ConnWaiter(c, c.Connect, ably.ConnectionEventConnected), nil)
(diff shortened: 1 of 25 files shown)


SimonWoolf:
Re the nitpick about using `http.NewRequestWithContext` for the DELETE at `ablytest/sandbox.go:269` — not taking this one.

It was flagged only because line 269 is the line this PR touched, but there are three identical `http.NewRequest` calls in the file (the POST at 221 and the JWT GET at 398 too). Fixing just one would make them inconsistent.

More importantly, `Sandbox.Close()` has no context in scope (signature `Close() error`, called as `defer app.Close()` from ~100 sites). The proposed fix passes `context.Background()`, which propagates no cancellation or timeout — it's purely cosmetic. Genuine propagation would mean threading a `ctx` through `Close()` and every caller. This is test infra and the HTTP client already has a fixed 1-minute timeout, so there's no behavioural gap to close here.

Leaving all three as-is for consistency.

— Claude

ablytest/sandbox.go
```diff
@@ -13,98 +14,118 @@ import (
...
 }
 
-type Connection struct {
-	Name string `json:"name"`
-	Key  string `json:"key"`
+// PresenceFixtures returns the presence members provisioned on the
+// persisted:presence_fixtures channel, matching the appspec JSON. client_encoded
+// is the cipher-encrypted form of client_decoded's data; tests that read it back
+// must configure the channel cipher (see PresenceFixturesCipher) to decode it.
+var PresenceFixtures = func() []Presence {
```
lmars: This says "matching the appspec JSON", could it just read from the fixture file rather than duplicating the data here in code?

ablytest/sandbox.go
```diff
@@ -13,98 +14,118 @@ import (
...
-			{ID: "persisted", Persisted: true},
-			{ID: "mutable", MutableMessages: true},
-		},
-		Channels: []Channel{
-			{
-				Name:     "persisted:presence_fixtures",
-				Presence: PresenceFixtures(),
-			},
-		},
+var loadAppSetup = func() func() appSetup {
```
lmars: This could use [`sync.OnceValue`](https://pkg.go.dev/sync#OnceValue) which effectively does the same thing.

ablytest/sandbox.go
```diff
@@ -13,98 +14,118 @@ import (
...
+	c := loadAppSetup().Cipher
+	key, err := base64.StdEncoding.DecodeString(c.Key)
+	if err != nil {
+		panic(err)
 	}
+	params := ably.Crypto.GetDefaultParams(ably.CipherParams{
+		Algorithm: ably.CipherAES,
+		Key:       key,
+	})
+	return params
```
lmars:
```suggestion
	return ably.Crypto.GetDefaultParams(ably.CipherParams{
		Algorithm: ably.CipherAES,
		Key:       key,
	})
```

ablytest/sandbox.go
```diff
@@ -125,31 +146,74 @@ func NewREST(opts ...ably.ClientOption) (*Sandbox, *ably.REST) {
 	return app, client
 }
 
+// MustSandbox returns the shared sandbox app, provisioning it on first use, and
+// panics if provisioning fails. The config argument is ignored: every test uses
+// the same app, provisioned from the shared appspec. It is retained only so the
+// many existing call sites compile unchanged.
```
lmars: I think we should just drop the argument (I acknowledge the diff might get larger, but it's largely a mechanical change I assume?)

ablytest/sandbox.go
```diff
@@ -125,31 +146,74 @@ func NewREST(opts ...ably.ClientOption) (*Sandbox, *ably.REST) {
...
 func MustSandbox(config *Config) *Sandbox {
-	app, err := NewSandbox(nil)
+	return GetSharedApp()
+}
+
+// NewSandbox returns the shared sandbox app (see MustSandbox). The config
+// argument is ignored.
+func NewSandbox(config *Config) (*Sandbox, error) {
```
lmars: As above, drop the arg?

ablytest/sandbox.go
```diff
@@ -125,31 +146,74 @@ func NewREST(opts ...ably.ClientOption) (*Sandbox, *ably.REST) {
...
+	return getSharedApp()
+}
+
+// NewSandboxWithEndpoint returns the shared sandbox app for the given endpoint.
+// The config argument is ignored. In practice the endpoint always matches the
+// package-level Endpoint, so this returns the same shared app as NewSandbox.
+func NewSandboxWithEndpoint(config *Config, endpoint string) (*Sandbox, error) {
```
lmars: Drop the unused `config` arg.

ablytest/sandbox.go
```diff
@@ -125,31 +146,74 @@ func NewREST(opts ...ably.ClientOption) (*Sandbox, *ably.REST) {
...
+	if endpoint == Endpoint {
+		return getSharedApp()
+	}
+	return provisionSandbox(endpoint, true)
+}
+
+var (
+	sharedAppOnce sync.Once
+	sharedApp     *Sandbox
+	sharedAppErr  error
```
lmars: Could use [`sync.OnceValues`](https://pkg.go.dev/sync#OnceValues) to cover these three args.

ablytest/sandbox.go
```diff
@@ -125,31 +146,74 @@ func NewREST(opts ...ably.ClientOption) (*Sandbox, *ably.REST) {
...
+)
+
+// GetSharedApp returns the process-wide shared sandbox app, provisioning it on
+// first call. All tests share this single app; isolation between tests is by
+// channel name rather than by app. Tests that call Close on it are no-ops — the
+// shared app is torn down once via CloseSharedApp (see TestMain).
+func GetSharedApp() *Sandbox {
```
lmars: We've now got `MustSandbox` and `GetSharedApp` which do exactly the same thing, should we just have one function?

ablytest/sandbox.go
```diff
@@ -125,31 +146,74 @@ func NewREST(opts ...ably.ClientOption) (*Sandbox, *ably.REST) {
...
 
-func NewSandboxWithEndpoint(config *Config, endpoint string) (*Sandbox, error) {
+// CloseSharedApp deletes the shared app if it was provisioned. It is intended to
+// be called once from TestMain after all tests have run.
+func CloseSharedApp() error {
+	if sharedApp == nil {
+		return nil
+	}
+	app := sharedApp
+	app.owned = true
```
lmars: What's the purpose of this field?

SimonWoolf: That field (`owned`) distinguished the shared app from a dedicated one, but on investigation the dedicated-app path was dead — no test ever provisioned a non-shared app — so I've removed the field entirely. `Close()` is now unconditionally a no-op and the shared app is deleted via an internal `delete()` from `CloseSharedApp`.

— Claude

ably/rest_channel_integration_test.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 func TestRESTChannel(t *testing.T) {
-	app, err := ablytest.NewSandbox(nil)
+	app, err := ablytest.NewSandbox()
 	assert.NoError(t, err)
 	defer app.Close()
```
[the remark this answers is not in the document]
SimonWoolf: Not making this change. Each of these channel names (`test_publish_channel`, `idempotent_test_1..6`, etc.) is used exactly once across the suite, so there's no cross-test residue within a run, and every run provisions a fresh app so there's no cross-run residue either. The full suite passes deterministically under both protocols. `ablytest.ChannelName(...)` was applied specifically to the channels that *are* genuinely shared across subtests/files (the `persisted:test` history tests); applying it to single-use names would add churn without changing behaviour.

— Claude

ablytest/sandbox.go
```diff
@@ -194,12 +245,21 @@ func NewSandboxWithEndpoint(config *Config, endpoint string) (*Sandbox, error) {
 	return nil, fmt.Errorf("Failed to request sandbox app after %d attempts.", RetryCount)
 }
 
+// Close is a no-op. The single shared app is owned by the test run, not by
+// individual tests, so the many per-test Close calls must not delete it;
+// teardown happens once via CloseSharedApp. Close is kept because tests call it
+// (often via defer) on the value returned by NewSandbox/NewREST/NewRealtime.
```
lmars: Sounds like we should just remove `Close()` altogether? Up to you.

lmars (approved):
One minor comment, otherwise LGTM

ablytest/sandbox.go
SimonWoolf: sure

ably/realtime_conn_integration_test.go
```diff
@@ -116,12 +116,11 @@ func TestRealtimeConn_ReceiveTimeout(t *testing.T) {
 	}
 	in <- connected
 
-	app, client := ablytest.NewRealtime(
+	_, client := ablytest.NewRealtime(
 		ably.WithDial(MessagePipe(in, out, MessagePipeWithNowFunc(time.Now))),
 		ably.WithRealtimeRequestTimeout(10*time.Millisecond),
 		ably.WithAutoConnect(false),
 	)
-	defer safeclose(t, app)
```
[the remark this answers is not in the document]
SimonWoolf: Fixed in 1c26017.

One clarification on the diagnosis: this was not a dropped client close. The original line was `defer safeclose(t, app)`, which only closed the sandbox app (now a no-op under the shared-app model) and never the realtime client — so the client was never closed here, before or after this PR. The added `defer safeclose(t, ablytest.FullRealtimeCloser(client))` closes it for the first time, bringing the test in line with its siblings (`BreakConnLoopOnInactiveState`, `SendErrorReconnects`). I also checked the other refactored sites and this was the only one missing a client close.

ablytest/sandbox.go
SimonWoolf: Done — `Close()` removed entirely in e71e4e0, along with all the per-test `defer app.Close()` / `safeclose(t, ..., app)` call sites.

SimonWoolf:
(holding off on merging because the integration tests seem to now sometimes flake with 42917, ie account-wide message rate limit. which is 700/s. even using only one account for the whole suite, with the relatively low level of parallelism, and lack of any tests that seem to be publishing particularly high volumes of messages, I just don't see how we could be getting anywhere near that, so not merging until that's figured out)

SimonWoolf:
@lmars requesting re-review for the last five commits, which fix a number of bugs causing the integration tests to fail or flake after the app change, including two that touch app code not just test code

lmars (approved):
LGTM

PR #706, opened 2026-07-07
test: gate ablytest sandbox provisioning behind an internal build tag

SimonWoolf:
The ablytest package is importable, so anything it exports is callable by downstream consumers. Those helpers are intended only for running ably-go's own integration tests, not as a supported API.

We don't want to actually put any security barries in place, because we want SDK development to be open to all, and there's no way to verify whether a given test app will be used for SDK development or not. But we can at least add a build tag that makes it at obvious that this is not the intended usecase.

This is not a security boundary (the code can be copied, or an external caller can just specify that tag), but rather a polite request, intended to avoid misunderstandings and unintentional reliance.

## Summary by CodeRabbit

* **Chores**
  * Tightened integration test execution to run only with the new internal test tag, helping separate these tests from standard builds.
  * Updated test workflow output to continue generating per-protocol reports.
* **Documentation**
  * Clarified integration test setup steps, including protocol-specific environment settings and the required build tags for new test files.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -46,7 +46,7 @@ jobs:
           ABLY_PROTOCOL: ${{ matrix.protocol == 'json' && 'application/json' || 'application/x-msgpack' }}
         run: |
           set -o pipefail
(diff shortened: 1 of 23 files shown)


SimonWoolf:
replacing with https://github.com/ably/ably-go/pull/707

PR #709, opened 2026-07-28
pubsub: add device and server packages

lmars:
Introduce per-side pubsub packages, so applications declare whether they are running on an end-user device or a server by their choice of package. Additive and shippable in a minor release.

## Packages

- **`pubsub/device`** — `NewClient`, returning a `Client` (= `ably.Realtime`)
- **`pubsub/server`** — `NewHTTPClient` and `NewRealtimeClient`, returning an `HTTPClient` (= `ably.REST`) and a `RealtimeClient` (= `ably.Realtime`)

The constructors return type aliases of today's clients rather than wrapper types: the side is the package and constructor choice plus the agent declaration, not a type distinction.

## Agent declaration

Each constructor appends a side-declaring agent (`ably-go-pubsub-device` / `ably-go-pubsub-server`) to the `Ably-Agent` header, which is what monthly-active-user classification keys on.

`WithAgents` now merges its entries into any agents already configured instead of replacing them, so the side agent appended by the constructors and a user-supplied `WithAgents` coexist regardless of option order.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added dedicated device-side client creation for Ably Pub/Sub connections.
  - Added server-side HTTP and realtime client constructors.
  - Client requests now identify whether they originate from a device or server.
  - Custom agent information is preserved and combined with these client identifiers.

- **Tests**
  - Added coverage verifying device and server client identification and custom agent merging.


pubsub/device/device.go
```diff
@@ -0,0 +1,21 @@
...
+// running on end-user devices, whose connections are identified by a
+// clientId and counted on accounts with monthly-active-user billing.
+package device
+
+import "github.com/ably/ably-go/ably"
+
+// agentName declares the side in the Ably-Agent header (RSC7d) so that
+// traffic from clients constructed by this package is classified as
+// device-side.
+const agentName = "ably-go-pubsub-device"
```
ttypic, line 11: probably should be `ably-pubsub-device-go`

paddybyers:
Do we expect there to be go devices?

lmars:
@paddybyers I think there was a recent discussion where it was suggested that the Go SDK could be used to build a desktop or CLI app, in which case it would be in a device context.

pubsub/device/device.go
lmars: Based on our internal decision record, the agent entry needs to end in `-device` or `-server`.

